PR #9, opened 2019-07-23
Added a contextual feature manager capable of evaluating a feature using an application defined context.

jimmyca15:
## IContextualFeatureManager

The contextual feature manager is a counterpart to the standard feature manager which is capable of utilizing a application defined context when evaluating the state of a feature. Applications that do not have a notion of an ambient context can pass their own context when evaluating feature state.

### Consumption

The `IContextualFeatureManager` can be obtained all the same ways that `IFeatureManager` can. This includes through dependency injection or by pulling directly from the service collection that feature management was added to.

```
IContextualFeatureManager cfm = services.GetRequiredService<IContextualFeatureManager>();

cfm.IsEnabled("featureName", new MyApplicationContext
{
  UserId = "someUser"
});
```

### Contextual Feature Filters
Contextual feature filters are feature filters that can utilize a context provided by the application when evaluating whether a feature is on or off. All contexts used by contextual feature filters must inherit from `IFeatureFilterContext` to declare intent to be used as a feature filter context. Implementations of contextual feature filters declare the interface that they require the feature filter context to implement to be able to properly handle them.

As an example, `IContextualFeatureFilter<IAccountContext>` requires a context that implements `IAccountContext` to be passed in. If a feature is checked for enabled and a context is provided that does not implement `IAccountContext` then the previously mentioned filter would not run.

### DIFF
--- a/Microsoft.FeatureManagement.sln
+++ b/Microsoft.FeatureManagement.sln
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio 15
-VisualStudioVersion = 15.0.28119.50
+# Visual Studio Version 16
+VisualStudioVersion = 16.0.29011.400
 MinimumVisualStudioVersion = 10.0.40219.1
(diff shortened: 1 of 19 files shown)


jimmyca15:
Will be abandoning this to implement async feature management, then will send a new updated version of this PR that takes advantage of async.

PR #14, opened 2019-09-09
Added the capability to pass in a context when checking if a feature is enabled.

jimmyca15:
# Passing In Application Context

This PR introduces the capability for applications to pass in a custom context to the feature management system for use in evaluating the state of features. The existing design for this PR is to gather this information from ambient contexts available within the application such as `HttpContext`. The problem with this existing design is that it is not available in console applications and many other frameworks.

### Consumption

The ability to pass a context when evaluating a feature has been added to `IFeatureManager`.

```
IFeatureManager fm = services.GetRequiredService<IFeatureManager>();

fm.IsEnabled("featureName", new MyApplicationContext
{
  UserId = "someUser"
});
```

### Contextual Feature Filters
Contextual feature filters are feature filters that can utilize a context provided by the application when evaluating whether a feature is on or off. Contextual feature filters are a generic type. Their generic type parameter describes the interface that the passed in context must implement for the filter to be able to evaluate it.

As an example, `IContextualFeatureFilter<IAccountContext>` requires a context that implements `IAccountContext` to be passed in. If a feature is checked for enabled and a context is provided that does not implement `IAccountContext` then the previously mentioned filter would not run.

An `IContextualFeatureFilter<object>` would be able to handle any passed in context.

### IFeatureFilterMetadata

With the introduction of `IContextualFeatureFilter` there are now two types of feature filters including `IFeatureFilter`. The two types of feature filters both inherit `IFeatureFilterMetadata`. `IFeatureFilterMetadata` is a marker interface and does not actually provide any feature filtering capabilities. It is used as the new parameter type for `IFeatureManagementBuilder.AddFeatureFilter`.

### DIFF
--- a/examples/ConsoleApp/FeatureFilters/AccountIdFilter.cs
+++ b/examples/ConsoleApp/FeatureFilters/AccountIdFilter.cs
@@ -1,0 +1,34 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+//
+using Consoto.Banking.AccountService.FeatureFilters;
+using Microsoft.Extensions.Configuration;
+using Microsoft.FeatureManagement;
+using System;
+using System.Collections.Generic;
+using System.Threading.Tasks;
+
+namespace Consoto.Banking.AccountService.FeatureManagement
+{
+    /// <summary>
+    /// A filter that uses the feature management context to ensure that the current task has the notion of an account id, and that the account id is allowed.
+    /// This filter will only be executed if an object implementing <see cref="IAccountContext"/> is passed in during feature evaluation.
+    /// </summary>
+    [FilterAlias("AccountId")]
+    class AccountIdFilter : IContextualFeatureFilter<IAccountContext>
+    {
+        public Task<bool> EvaluateAsync(FeatureFilterEvaluationContext featureEvaluationContext, IAccountContext accountId)
+        {
+            if (string.IsNullOrEmpty(accountId?.AccountId))
+            {
+                throw new ArgumentNullException(nameof(accountId));
+            }
+
+            var allowedAccounts = new List<string>();
+
+            featureEvaluationContext.Parameters.Bind("AllowedAccounts", allowedAccounts);
+
+            return Task.FromResult(allowedAccounts.Contains(accountId.AccountId));
+        }
+    }
+}
--- a/examples/ConsoleApp/FeatureFilters/AccountIdFilter.cs
+++ b/examples/ConsoleApp/FeatureFilters/AccountIdFilter.cs
@@ -17,18 +17,18 @@ namespace Consoto.Banking.AccountService.FeatureManagement
     [FilterAlias("AccountId")]
     class AccountIdFilter : IContextualFeatureFilter<IAccountContext>
     {
-        public Task<bool> EvaluateAsync(FeatureFilterEvaluationContext featureEvaluationContext, IAccountContext accountId)
+        public Task<bool> EvaluateAsync(FeatureFilterEvaluationContext featureEvaluationContext, IAccountContext accountContext)
         {
-            if (string.IsNullOrEmpty(accountId?.AccountId))
+            if (string.IsNullOrEmpty(accountContext?.AccountId))
             {
-                throw new ArgumentNullException(nameof(accountId));
+                throw new ArgumentNullException(nameof(accountContext));
             }
 
             var allowedAccounts = new List<string>();
 
             featureEvaluationContext.Parameters.Bind("AllowedAccounts", allowedAccounts);
 
-            return Task.FromResult(allowedAccounts.Contains(accountId.AccountId));
+            return Task.FromResult(allowedAccounts.Contains(accountContext.AccountId));
         }
     }
 }
--- a/src/Microsoft.FeatureManagement/ContextualFeatureFilterEvaluator.cs
+++ b/src/Microsoft.FeatureManagement/ContextualFeatureFilterEvaluator.cs
@@ -1,0 +1,99 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+//
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Reflection;
+using System.Threading.Tasks;
+
+namespace Microsoft.FeatureManagement
+{
+    /// <summary>
+    /// Provides a performance efficient method of evaluating IContextualFeatureFilter&lt;T&gt; without knowing what the generic type parameter is.
+    /// </summary>
+    class ContextualFeatureFilterEvaluator : IContextualFeatureFilter<object>
+    {
+        private IFeatureFilterMetadata _filter;
+        private Func<object, FeatureFilterEvaluationContext, object, Task<bool>> _evaluateFunc;
+
+        public ContextualFeatureFilterEvaluator(IFeatureFilterMetadata filter, Type appContextType)
+        {
+            if (filter == null)
+            {
+                throw new ArgumentNullException(nameof(filter));
+            }
+
+            if (appContextType == null)
+            {
+                throw new ArgumentNullException(nameof(appContextType));
+            }
+
+            Type targetInterface = GetContextualFilterInterface(filter, appContextType);
+
+            //
+            // Extract IContextualFeatureFilter<T>.EvaluateAsync method.
+            if (targetInterface != null)
+            {
+                MethodInfo evaluateMethod = targetInterface.GetMethod(nameof(IContextualFeatureFilter<object>.EvaluateAsync), BindingFlags.Public | BindingFlags.Instance);
+
+                _evaluateFunc = TypeAgnosticEvaluate(filter.GetType(), evaluateMethod);
+            }
+
+            _filter = filter;
+        }
+
+        public static Type GetContextualFilterInterface(IFeatureFilterMetadata filter, Type appContextType)
+        {
+            IEnumerable<Type> contextualFilterInterfaces = filter.GetType().GetInterfaces().Where(i => i.IsGenericType && i.GetGenericTypeDefinition().IsAssignableFrom(typeof(IContextualFeatureFilter<>)));
+
+            Type targetInterface = null;
+
+            if (contextualFilterInterfaces != null)
+            {
+                targetInterface = contextualFilterInterfaces.FirstOrDefault(i => i.GetGenericArguments()[0].IsAssignableFrom(appContextType));
+            }
+
+            return targetInterface;
+        }
+
+        public Task<bool> EvaluateAsync(FeatureFilterEvaluationContext evaluationContext, object context)
+        {
+            if (_evaluateFunc == null)
+            {
+                return Task.FromResult(false);
+            }
+
+            return _evaluateFunc(_filter, evaluationContext, context);
+        }
+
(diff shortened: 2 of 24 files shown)


examples/ConsoleApp/FeatureFilters/AccountIdFilter.cs
```diff
@@ -0,0 +1,34 @@
...
+namespace Consoto.Banking.AccountService.FeatureManagement
+{
+    /// <summary>
+    /// A filter that uses the feature management context to ensure that the current task has the notion of an account id, and that the account id is allowed.
+    /// This filter will only be executed if an object implementing <see cref="IAccountContext"/> is passed in during feature evaluation.
+    /// </summary>
+    [FilterAlias("AccountId")]
+    class AccountIdFilter : IContextualFeatureFilter<IAccountContext>
+    {
+        public Task<bool> EvaluateAsync(FeatureFilterEvaluationContext featureEvaluationContext, IAccountContext accountId)
```
MSGaryWang: >accountId [](start = 113, length = 9)

accountContext?

tests/Tests.FeatureManagement/TestFilter.cs
```diff
@@ -1,19 +1,35 @@
...
+    class TestFilter : IFeatureFilter, IContextualFeatureFilter<IAccountContext>
     {
         public Func<FeatureFilterEvaluationContext, bool> Callback { get; set; }
 
         public Task<bool> EvaluateAsync(FeatureFilterEvaluationContext context)
         {
             return Task.FromResult(Callback?.Invoke(context) ?? false);
         }
+
+        public Task<bool> EvaluateAsync(FeatureFilterEvaluationContext featureFilterEvaluationContext, IAccountContext accountContext)
```
MSGaryWang: >public Task<bool> EvaluateAsync(FeatureFilterEvaluationContext featureFilterEvaluationContext [](start = 8, length = 93)

maybe we can have another CallackWithContext to make it reusable like above

src/Microsoft.FeatureManagement/FeatureManager.cs
```diff
@@ -77,12 +89,28 @@ public async Task<bool> IsEnabledAsync(string feature)
                             Parameters = featureFilterSettings.Parameters 
                         };
 
-                        if (await filter.EvaluateAsync(context).ConfigureAwait(false))
+                        //
+                        // IFeatureFilter
+                        if (filter is IFeatureFilter featureFilter && await featureFilter.EvaluateAsync(context).ConfigureAwait(false))
```
MSGaryWang: >IFeatureFilter [](start = 38, length = 14)

we should check useAppContext first in case a filter instance implement both IFeatureFilter and IContextualFeatureFilter

src/Microsoft.FeatureManagement/FeatureManager.cs
```diff
@@ -145,5 +173,32 @@ private IFeatureFilter GetFeatureFilter(string filterName)
...
+                    IFeatureFilterMetadata metadata = GetFeatureFilterMetadata(filterName);
+
+                    ContextualFeatureFilterEvaluator f = null;
+
+                    if (ContextualFeatureFilterEvaluator.GetContextualFilterInterface(metadata, appContextType) != null)
+                    {
+                        f = new ContextualFeatureFilterEvaluator(metadata, appContextType);
+                    }
+
+                    return f;
```
MSGaryWang: Can we hide this logic inside ContextualFeatureFilterEvaluator class? It might be a little too much to know the details from outside. Make it like below with a static function

return ContextualFeatureFilterEvaluator.Create(metadata, appContextType);

src/Microsoft.FeatureManagement/ContextualFeatureFilterEvaluator.cs
```diff
@@ -0,0 +1,99 @@
...
+
+        public static Type GetContextualFilterInterface(IFeatureFilterMetadata filter, Type appContextType)
+        {
+            IEnumerable<Type> contextualFilterInterfaces = filter.GetType().GetInterfaces().Where(i => i.IsGenericType && i.GetGenericTypeDefinition().IsAssignableFrom(typeof(IContextualFeatureFilter<>)));
+
+            Type targetInterface = null;
+
+            if (contextualFilterInterfaces != null)
+            {
+                targetInterface = contextualFilterInterfaces.FirstOrDefault(i => i.GetGenericArguments()[0].IsAssignableFrom(appContextType));
```
MSGaryWang: >FirstOrDefault [](start = 61, length = 14)

if user's filter implement two IContextualFeatureFilter<A> and IContextualFeatureFilter<B> and class A : B {}. In this case, we might not just get the first but the best matched.

src/Microsoft.FeatureManagement/ContextualFeatureFilterEvaluator.cs
```diff
@@ -0,0 +1,99 @@
...
+
+            return (Func<object, FeatureFilterEvaluationContext, object, Task<bool>>)ret;
+        }
+
+        static Func<object, FeatureFilterEvaluationContext, object, Task<bool>> GenericTypeAgnosticEvaluate<TTarget, TParam1, TParam2, TReturn>(MethodInfo method)
+        {
+            Func<TTarget, FeatureFilterEvaluationContext, TParam2, Task<bool>> func = (Func<TTarget, FeatureFilterEvaluationContext, TParam2, Task<bool>>)Delegate.CreateDelegate
+                (typeof(Func<TTarget, FeatureFilterEvaluationContext, TParam2, Task<bool>>), method);
+
+            Func<object, FeatureFilterEvaluationContext, object, Task<bool>> ret = (object target, FeatureFilterEvaluationContext param1, object param2) => func((TTarget)target, param1, (TParam2)param2);
```
MSGaryWang: >ret [](start = 77, length = 3)

nit: this 'ret' confuse me, how about delegateFunc? Same as line 84.

src/Microsoft.FeatureManagement/FeatureManager.cs
jimmyca15: How about

```
  return ContextualFeatureFilterEvaluator.IsContextualFilter(metadata, appContextType) ?
    new ContextualFeatureFilterEvaluator(metadata, appContextType) :
    null;
```

---
In reply to: [328393965](https://github.com/microsoft/FeatureManagement-Dotnet/pull/14#discussion_r328393965) [](ancestors = 328393965)

examples/ConsoleApp/FeatureFilters/AccountIdFilter.cs
jimmyca15: Yes, thanks.

---
In reply to: [328363265](https://github.com/microsoft/FeatureManagement-Dotnet/pull/14#discussion_r328363265) [](ancestors = 328363265)

src/Microsoft.FeatureManagement/FeatureManager.cs
jimmyca15: This is actually intentional, even if appContext is usable we still want to run IFeatureFilter if they implement that interface.

---
In reply to: [328391708](https://github.com/microsoft/FeatureManagement-Dotnet/pull/14#discussion_r328391708) [](ancestors = 328391708)

src/Microsoft.FeatureManagement/ContextualFeatureFilterEvaluator.cs
jimmyca15: That scenario is quite confusing. I'm not sure it could be clear what to expect there. I think it would be better not to support that.

---
In reply to: [328395534](https://github.com/microsoft/FeatureManagement-Dotnet/pull/14#discussion_r328395534) [](ancestors = 328395534)

src/Microsoft.FeatureManagement/ContextualFeatureFilterEvaluator.cs
jimmyca15: If that is all that confused you then that is awesome :), this code is quite complicated. I'll update it.

---
In reply to: [328399200](https://github.com/microsoft/FeatureManagement-Dotnet/pull/14#discussion_r328399200) [](ancestors = 328399200)

jimmyca15:
@MSGaryWang I have responded to all of your feedback. In my latest commits you'll notice some changes to address the conversation we had about implementing multiple feature filter interfaces.

* To prevent complicated setups, `AddFeatureFilter` now throws ArgumentException if a feature filter is added that implements more than a single feature filter interface.
* The test feature filter was now broke into two separate filters.

MSGaryWang:
>                         }

This part might move to line 105 because filter wasn't used before that

---
Refers to: src/Microsoft.FeatureManagement/FeatureManager.cs:84 in 2688f1c. [](commit_id = 2688f1cdb328ba403009b114b3c4a8d7b05d80b1, deletion_comment = False)

jimmyca15:
>                         }

If there is no filter metadata for the given filter name then there's no reason to execute any of the code after this. I don't think we should be moving this. There's not going to be an `IFeatureFilter` or `IContextualFeatureFilter`. Note that `IFeatureFilterMetadata` is the base interface so the absence of this means the absence of any feature filter.

---
In reply to: [548133624](https://github.com/microsoft/FeatureManagement-Dotnet/pull/14#issuecomment-548133624) [](ancestors = 548133624)

---
Refers to: src/Microsoft.FeatureManagement/FeatureManager.cs:84 in 2688f1c. [](commit_id = 2688f1cdb328ba403009b114b3c4a8d7b05d80b1, deletion_comment = False)

MSGaryWang:
>                         }

For that case you might passing filter to GetContextualFeatureFilter() below to prevent possibly duplicated task to get it again? 

---
In reply to: [548453796](https://github.com/microsoft/FeatureManagement-Dotnet/pull/14#issuecomment-548453796) [](ancestors = 548453796,548133624)

---
Refers to: src/Microsoft.FeatureManagement/FeatureManager.cs:84 in 2688f1c. [](commit_id = 2688f1cdb328ba403009b114b3c4a8d7b05d80b1, deletion_comment = False)

jimmyca15:
@MSGaryWang the results of `GetFeatureFilterMetadata` are cached, I'd rather have the additional cache lookup (which only occurs the first time around) then passing in the filter. Actually passing in the filter would make things a bit more complicated considering how we are caching `ContextualFeatureFilterEvaluator`

MSGaryWang:
There seems still one left then let's get this in

---
In reply to: [537667781](https://github.com/microsoft/FeatureManagement-Dotnet/pull/14#issuecomment-537667781) [](ancestors = 537667781)

jimmyca15:
@MSGaryWang What do you mean? I have replied to all your feedback, see my latest reply

> @MSGaryWang the results of GetFeatureFilterMetadata are cached, I'd rather have the additional cache lookup (which only occurs the first time around) then passing in the filter. Actually passing in the filter would make things a bit more complicated considering how we are caching ContextualFeatureFilterEvaluator

examples/ConsoleApp/FeatureFilters/AccountIdFilter.cs
MSGaryWang: this one

---
In reply to: [328746974](https://github.com/microsoft/FeatureManagement-Dotnet/pull/14#discussion_r328746974) [](ancestors = 328746974,328363265)

MSGaryWang:
examples/ConsoleApp/FeatureFilters/AccountIdFilter.cs

---
In reply to: [549958553](https://github.com/microsoft/FeatureManagement-Dotnet/pull/14#issuecomment-549958553) [](ancestors = 549958553)

jimmyca15:
@MSGaryWang done.

PR #20, opened 2019-11-25
Added feature settings caching.

jimmyca15:
A perf analysis of the library pointed out a bottle neck in the feature management pipeline during obtaining feature settings. This is due to the fact that sections in the .NET Core configuration were being enumerated on every lookup.

The `ConfigurationFeatureSettingsProvider` has been improved to cache feature settings when they have been read from the .NET Core configuration system. The `IConfiguration`'s reload token is used to track when the configuration has become stale.

`ConfigurationFeatureSettingsProvider` now implements `IDisposable` since it registers a callback on a reload token. The disposable of this will happen when the service container is disposed.

### DIFF
--- a/src/Microsoft.FeatureManagement/ConfigurationFeatureSettingsProvider.cs
+++ b/src/Microsoft.FeatureManagement/ConfigurationFeatureSettingsProvider.cs
@@ -2,26 +2,52 @@
 // Licensed under the MIT license.
 //
 using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.Primitives;
 using System;
+using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Linq;
+using System.Threading;
 
 namespace Microsoft.FeatureManagement
 {
     /// <summary>
     /// A feature settings provider that pulls settings from the .NET Core <see cref="IConfiguration"/> system.
     /// </summary>
-    sealed class ConfigurationFeatureSettingsProvider : IFeatureSettingsProvider
+    sealed class ConfigurationFeatureSettingsProvider : IFeatureSettingsProvider, IDisposable
     {
         private const string FeatureFiltersSectionName = "EnabledFor";
         private readonly IConfiguration _configuration;
+        private readonly ConcurrentDictionary<string, IFeatureSettings> _settings;
+        private readonly IDisposable _changeSubscription;
+        private int _stale = 0;
 
         public ConfigurationFeatureSettingsProvider(IConfiguration configuration)
         {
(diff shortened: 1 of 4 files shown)


src/Microsoft.FeatureManagement/ConfigurationFeatureSettingsProvider.cs
```diff
@@ -2,26 +2,52 @@
...
         }
 
         public IFeatureSettings TryGetFeatureSettings(string featureName)
+        {
+            if (Interlocked.Exchange(ref _stale, 0) != 0)
+            {
+                _settings.Clear();
+            }
+
+            return _settings.GetOrAdd(featureName, (name) => ReadFeatureSettings(name));
```
MSGaryWang: >ReadFeatureSettings [](start = 61, length = 19)

I through we should just cache our settings not their result? Since we don't know how user implement their filters or how they want to cache it?

src/Microsoft.FeatureManagement/ConfigurationFeatureSettingsProvider.cs
```diff
@@ -2,26 +2,52 @@
...
+            _settings = new ConcurrentDictionary<string, IFeatureSettings>();
+
+            _changeSubscription = ChangeToken.OnChange(
+                () => _configuration.GetReloadToken(),
+                () => _stale = 1);
+        }
+
+        public void Dispose()
+        {
+            _changeSubscription.Dispose();
```
MSGaryWang: >. [](start = 31, length = 1)

nit: maybe better using ?. here and set  _changeSubscription to null

src/Microsoft.FeatureManagement/ConfigurationFeatureSettingsProvider.cs
jimmyca15: You're right. This is only caching the settings for the feature, not the result of their evaluations. The state of the feature is evaluated in `FeatureManager`. That class calls on this to get settings for a feature to know what filters to run, etc...

Whenever the configuration underneath us changes we update our cache. This is tracked by the `_stale` bit.

---
In reply to: [350426030](https://github.com/microsoft/FeatureManagement-Dotnet/pull/20#discussion_r350426030) [](ancestors = 350426030)

src/Microsoft.FeatureManagement/ConfigurationFeatureSettingsProvider.cs
jimmyca15: Sure, done.

---
In reply to: [350426050](https://github.com/microsoft/FeatureManagement-Dotnet/pull/20#discussion_r350426050) [](ancestors = 350426050)

PR #22, opened 2019-11-26
Fixed the build and package process.

jimmyca15:
Recent changes in build infrastructure have mandated some changes in the build. This PR fixes these issues.

### DIFF
--- a/.pipelines/pipeline.user.windows.yml
+++ b/.pipelines/pipeline.user.windows.yml
@@ -10,7 +10,6 @@ restore:
     - !!defaultcommand
       name: 'Restore'
       command: 'build.cmd'
-      arguments: 'RestoreOnly'
 
 build:
   commands:
--- a/.pipelines/pipeline.user.windows.yml
+++ b/.pipelines/pipeline.user.windows.yml
@@ -9,7 +9,7 @@ restore:
   commands:
     - !!defaultcommand
       name: 'Restore'
-      command: 'build.cmd'
+      command: 'dotnet restore'
 
 build:
   commands:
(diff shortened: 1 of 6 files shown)


jimmyca15:
This PR also updates the major version of the package from 1 to 2.

.pipelines/pipeline.user.windows.yml
```diff
@@ -10,7 +10,6 @@ restore:
     - !!defaultcommand
       name: 'Restore'
       command: 'build.cmd'
-      arguments: 'RestoreOnly'
```
zhenlan: Are we calling build.cmd twice, once for restore and once for build? Do we still need this restore section?

build.cmd
```diff
@@ -1 +1,3 @@
-powershell.exe -ExecutionPolicy Unrestricted -NoProfile -File "%~dp0build.ps1"
+call %~dp0build\InstallPowerShellCore.cmd
+
+%PowerShellCore% "%~dp0build.ps1" %1
```
zhenlan: For future proof, we can pass `%*` instead.

jimmyca15: Good advice, updated.

.pipelines/pipeline.user.windows.yml
jimmyca15: Yes, this was done because pulling nuget packages can only be done in the "Restore" phase of cdpx builds. Invoking build during the restore phase caused the nuget packages to get pulled and then cached for the rest of the build.

Although, building twice is unnecessary so I have just restored the "-RestoreOnly" argument and instead updated the build.ps1 script to respect it.

PR #36, opened 2019-12-17
Updated IFeatureManager to expose GetFeatureNames.

### LINKED ISSUE #29: Need ability to get all features
I need the ability to get a list of all features and potentially get only enabled features as well.

jimmyca15:
## Get All Features

This PR updates `IFeatureManager` to expose `Task<IEnumerable<string> GetFeatureNames()`. This is necessary to be able to enumerate features and check what is enabled when the feature set may expand or contract during the run time of the application.

Typical usage would be as follows.

```
List<string> enabledFeatures = new List<string>();

foreach (string feature in await _featureManager.GetFeatureNamesAsync())
{
    if (await _featureManager.IsEnabled(feature))
    {
        enabledFeatures.Add(feature);
    }
}

//
// do something with enabledFeatures.
```

This resolves #29.

### Internals

This PR also cleans up IFeatureSettingsProvider to prepare it for becoming a public API surface. Unnecessary interface clutter has been removed with the deletion of `IFeatureSettings` and `IFeatureFilterSettings`. Since these interfaces are POCOs, there is no need for them.


src/Microsoft.FeatureManagement/FeatureManager.cs
```diff
@@ -42,6 +42,28 @@ public Task<bool> IsEnabledAsync<TContext>(string feature, TContext appContext)
             return IsEnabledAsync(feature, appContext, true);
         }
 
+        public async Task<IEnumerable<string>> GetFeatureNamesAsync()
```
zhenlan: Since this is all operating in memory, why does it need to be an async function?

zhenlan: Alright, user can provide custom IFeatureSettingsProvider.

---
In reply to: [382884552](https://github.com/microsoft/FeatureManagement-Dotnet/pull/36#discussion_r382884552) [](ancestors = 382884552)

jimmyca15: Right the intent is to design around the capability for async feature settings provider. Although, currently `IFeatureSettingsProvider` is not public. That would be a different PR.

src/Microsoft.FeatureManagement/FeatureSettingsQueryOptions.cs
```diff
@@ -0,0 +1,18 @@
...
+    {
+        /// <summary>
+        /// The name of an individual feature to select.
+        /// </summary>
+        public string FeatureName { get; set; }
+
+        /// <summary>
+        /// A filter used to select all feature settings after the feature with the specified name.
+        /// </summary>
+        public string After { get; set; }
```
zhenlan: I am a bit hesitated with this one but don't have a good suggestion, so like to brainstorm a little :).

All we want is a way to query all feature names, but this requires that the query can be resumed from any previous query. This increases the difficulty of implementing custom `IFeatureSettingsProvider`. Even our own implementation in `ConfigurationFeatureSettingsProvider.cs` is not the easiest to understand if I am not aware of where we are from. It also makes it harder to debug if any issues are found in the future.

jimmyca15: The resume behavior is so that implemenations of `FeatureSettingsProvider` don't have to return all of the feature settings in one enumerable. That enumerable could be unbound.

Another way we can solve this is to have something like this

```
IFeatureSettingsProvider
{
  Task<FeatureSettings> GetFeatureSettingsAsync(string featureName);

  IAsyncEnumerable<FeatureSettings> GetAllFeatureSettingsAsync();
}
```

Async enumerable provides the chance for the implementator to page their results.

zhenlan: I like the new proposal. It separates the concerns.

jimmyca15: Sounds good to me. I'll make the update. We'll have to depend on IAsyncEnumerable, which based off of the Azure SDK ConfigurationClient means adding a reference to [https://www.nuget.org/packages/Azure.Data.AppConfiguration/](Microsoft.Bcl.AsyncInterfaces (>= 1.0.0))

zhenlan: Microsoft.Bcl.AsyncInterfaces does *not* depend on Azure SDK, right?

jimmyca15: Not at all. just System.Threading.Tasks.Extensions. Looks like that link in my last comment got messed up.

https://www.nuget.org/packages/Microsoft.Bcl.AsyncInterfaces/

jimmyca15: Looks like we should also be making this change for `GetFeatureNamesAsync`.

So it will go from

`Task<IEnumerable<string>> GetFeatureNamesAsync();`

to  

`IAsyncEnumerable<string> GetFeatureNamesAsync();`

zhenlan: Makes sense. Do we like to name it `GetAllFeatureNamesAsync()`?

jimmyca15: I would think the plural "names" is good enough to not have the `All` part. The only reason the `All` is in IFeatureSettingsProvider is because "settings" is the same plural/singular.

zhenlan: Sounds good. Agreed.

zhenlan: Old code to clean up.

src/Microsoft.FeatureManagement/FeatureManagerSnapshot.cs
```diff
@@ -14,12 +14,33 @@ class FeatureManagerSnapshot : IFeatureManagerSnapshot
...
+        private IEnumerable<string> _featureNames;
 
         public FeatureManagerSnapshot(IFeatureManager featureManager)
         {
             _featureManager = featureManager ?? throw new ArgumentNullException(nameof(featureManager));
         }
 
+        public async IAsyncEnumerable<string> GetFeatureNamesAsync()
+        {
+            if (_featureNames != null)
```
zhenlan: `if (_featureNames == null)`

src/Microsoft.FeatureManagement/ConfigurationFeatureSettingsProvider.cs
```diff
@@ -39,17 +40,53 @@ public void Dispose()
...
+            if (Interlocked.Exchange(ref _stale, 0) != 0)
+            {
+                _settings.Clear();
+            }
+
+            var featureSettings = new List<FeatureSettings>();
+
+            //
+            // Query all
+            foreach (string featureName in GetFeatureConfigurationSections().Select(s => s.Key))
```
zhenlan: Will it be possible for `GetFeatureConfigurationSections()` to return `null`?

jimmyca15: No, that's a private method where we don't return null.

src/Microsoft.FeatureManagement/FeatureManagerSnapshot.cs
jimmyca15: Thank you!

src/Microsoft.FeatureManagement/FeatureSettingsQueryOptions.cs
jimmyca15: Yes it is.

src/Microsoft.FeatureManagement/ConfigurationFeatureSettingsProvider.cs
```diff
@@ -39,17 +40,53 @@ public void Dispose()
...
+            {
+                FeatureSettings settings = _settings.GetOrAdd(featureName, (name) => ReadFeatureSettings(name));
```
zhenlan: It seems possible the `ReadFeatureSettings` is dealing with a different `IConfigurationSection` than what are returned from `GetFeatureConfigurationSections` in the `foreach` clause when the configuration is changed after the `Interlocked.Exchange` check.

We could add an overload of `ReadFeatureSettings` that takes an optional `IConfigurationSection`.

jimmyca15: We could, but wouldn't we want to return the latest settings of the feature?

The `GetFeatureConfigurationSections().Select(s => s.Key)` line is to enumerate over all the feature names that are in the system, not necessarily to acquire their current state. We pull the state/settings when the caller moves the enumerator. Maybe I could put a comment for that?

I think the real downside here is that feature names could be added into the IConfiguration system while the caller is enumerating over the returned IAsyncEnumerable and we would miss them. However, the fix for that would involve a lot of calls to retrieve configuration sections and I'm not sure it's worth it.

zhenlan: We do want the latest settings. That's why we called `_settings.Clear()` at the beginning of the function. But once we passed that line and made the first call to `GetFeatureConfigurationSections()`, we should stick with that snapshot of configuration. We shouldn't allow configuration changes while we enumerating the collection. This could cause all sorts of hard-to-debug issues when it happens.

I am thinking of an implementation something like this

```cs
foreach (IConfigurationSection featureSection in GetFeatureConfigurationSections())
{
  FeatureSettings settings = _settings.GetOrAdd(featureSection.Key, (featureSection) => ReadFeatureSettings(featureSection));

  ... ...
}

// ... and an overload of ReadFeatureSettings
private FeatureSettings ReadFeatureSettings(IConfigurationSection featureSection)
{
  // This overload will NOT call GetFeatureConfigurationSections(). 
  // It will simply use the IConfigurationSection that is passed in.
  // We can change the current ReadFeatureSettings(string featureName) to call this one.
}

jimmyca15: Discussed offline. IConfigurationSection does not provide snapshot behavior so there is no way to capture the point-in-time state of all features at the initial invocation of `GetAllFeatureSettingsAsync()`. Thus we will find all feature names that exist at this time, and as the async iterator is moved through we will return the latest state of that feature, or a "default" `FeatureSettings` object if the feature's configuration section was deleted during the async iteration.

PR #47, opened 2020-02-04
Removed `useAppContext` parameter from private `IsEnabledAsync`

antmdvs:
The `useAppContext` argument was only being passed from the contextual overload of `IsEnabledAsync`. It seems like the state represented by `useAppContext` could be determined based on whether `appContext` is null or non-null.


jimmyca15:
If the user passes a null value for `appContext` this breaks. The contextual feature filter is no longer run.

antmdvs:
I see. I guess it seems unintuitive to me that a `null` would constitute useful `appContext` for contextual filter logic so why even run the filter?

PR #50, opened 2020-02-21
Enhances FeatureManager to throw an exception for missing feature filters.

### LINKED ISSUE #13: An exception should be thrown if a configured feature filter has not been registered.

jimmyca15:
The current feature manager implementation only logs a warning when a feature filter is missing for a given feature. This makes missing feature filters tricky to diagnose and can lead to unintended behavior. This enhancement changes the default behavior of the feature manager so that it will throw a `FeatureManagementException` when it encounters a missing feature filter.

This new behavior can be disabled by configuring the feature management options. 

```
services.Configure<FeatureManagementOptions>(options =>
{
    options.IgnoreMissingFeatureFilters = true;
});
```

This resolves #13 

@codapus-co


src/Microsoft.FeatureManagement/FeatureManagementOptions.cs
```diff
@@ -0,0 +1,17 @@
...
+namespace Microsoft.FeatureManagement
+{
+    /// <summary>
+    /// Options that control the behavior of the feature management system.
+    /// </summary>
+    public class FeatureManagementOptions
+    {
+        /// <summary>
+        /// Controls the behavior of feature evaluation when dependent feature filters are missing.
+        /// If missing features filters are not ignored an exception will be thrown when attempting to evaluate a feature that depends on a missing feature filter.
```
zhenlan: NIT: ... feature~s~ filters ...

jimmyca15: Thanks ! I'll update before merge.

src/Microsoft.FeatureManagement/FeatureManager.cs
```diff
@@ -21,15 +22,22 @@ class FeatureManager : IFeatureManager
...
+            ILoggerFactory loggerFactory,
+            IOptions<FeatureManagementOptions> options)
         {
             _settingsProvider = settingsProvider;
             _featureFilters = featureFilters ?? throw new ArgumentNullException(nameof(featureFilters));
             _sessionManagers = sessionManagers ?? throw new ArgumentNullException(nameof(sessionManagers));
             _logger = loggerFactory.CreateLogger<FeatureManager>();
             _filterMetadataCache = new ConcurrentDictionary<string, IFeatureFilterMetadata>();
             _contextualFeatureFilterCache = new ConcurrentDictionary<string, ContextualFeatureFilterEvaluator>();
+            _options = options?.Value ?? throw new ArgumentNullException(nameof(options));
```
MSGaryWang: >options?.Value [](start = 23, length = 14)

so we don't expect a default value?

jimmyca15: The default value should be returned by Value.

PR #51, opened 2020-02-22
Fixed ISessionManager.TryGetAsync so that it can be implement as async.

### LINKED ISSUE #38: async implementation of TryGetAsync on ISessionManager not possible?
The `ISessionManager` interface switched to using an `async` approach, which is good in itself.

However the TryGetAsync function uses an `out` parameter which forbids declaring the function as `async` - so you cannot `await` `Task`s in the function body.

This feels a bit awkward - do you have any example on how to best tackle that?

jimmyca15:
`ISessionManager.TryGetAsync` currently contains an out parameter. This doesn't allow an async implementation. This PR combines the out parameter with the return value to enable async implementation.

resolves #38


src/Microsoft.FeatureManagement/ReadSessionResult.cs
```diff
@@ -0,0 +1,18 @@
+﻿namespace Microsoft.FeatureManagement
+{
+    /// <summary>
+    /// The result of reading the feature state for a session.
+    /// </summary>
+    public struct ReadSessionResult
```
zhenlan: Why don't we use nullable bool instead?

src/Microsoft.FeatureManagement/ISessionManager.cs
```diff
@@ -21,8 +21,7 @@ public interface ISessionManager
         /// Queries the session manager for the session's feature state, if any, for the given feature.
         /// </summary>
         /// <param name="featureName">The name of the feature.</param>
-        /// <param name="enabled">The state from the session, if any.</param>
-        /// <returns>True if the session manager was able to provide a state for the feature. False otherwise.</returns>
-        Task<bool> TryGetAsync(string featureName, out bool enabled);
+        /// <returns>A result indicating whether the session was able to provide a the state of the feature, and the state, if any.</returns>
+        Task<ReadSessionResult> TryGetAsync(string featureName);
```
zhenlan: Should the API name just be `GetAsync` (without Try)?

src/Microsoft.FeatureManagement/ReadSessionResult.cs
jimmyca15: Something threw me off with nullable bool. Perhaps because there once was a discussion to have feature manager return nullable bool as a meaning that the state of a feature couldn't be determined. 

But, nullable bool could work here. It's simpler. We can go with that.

src/Microsoft.FeatureManagement/ISessionManager.cs
jimmyca15: If we do nullable boolean I feel like the name should remain the same.

zhenlan: Normally the return value of Try* is to indicate whether the operation is successful and the operation result is returned via an out parameter. This is why the old API makes sense. Now that we return the operation result directly, which just happens to be boolean, the `Try` could be confusing.

jimmyca15: The `Try` semantic isn't as well defined in async methods, so I agree this could be confusing. After more thought I feel there is no need to have `Try` in this method name anyways. I'll change it to `GetAsync`. Thanks.

jimmyca15: @zhenlan updated. It appears I had pushed it to the wrong branch earlier so the update didn't show up.

PR #52, opened 2020-02-24
Updated feature manager to throw FeatureManagementException instead of InvalidOperationException.

jimmyca15:
Currently, the feature manager throws an `InvalidOperationException` in the event that multiple feature filters match a name provided in the feature's filter list. The InvalidOperationException constitutes an error that doesn't really make sense for this type of problem. 

Now that we have the `FeatureManagementException` we can propogate the correct error, `AmbigousFeatureFilter`.

### DIFF
--- a/src/Microsoft.FeatureManagement/FeatureManagementError.cs
+++ b/src/Microsoft.FeatureManagement/FeatureManagementError.cs
@@ -8,6 +8,11 @@
         /// <summary>
         /// A feature filter that was listed for feature evaluation was not found.
         /// </summary>
-        MissingFeatureFilter
+        MissingFeatureFilter,
+
+        /// <summary>
+        /// Multiple feature filters match the name listed for feature evaluation.
+        /// </summary>
+        ConflictingFeatureFilter
     }
 }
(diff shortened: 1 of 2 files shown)


src/Microsoft.FeatureManagement/FeatureManagementError.cs
```diff
@@ -8,6 +8,11 @@ public enum FeatureManagementError
         /// <summary>
         /// A feature filter that was listed for feature evaluation was not found.
         /// </summary>
-        MissingFeatureFilter
+        MissingFeatureFilter,
+
+        /// <summary>
+        /// Multiple feature filters match the name listed for feature evaluation.
+        /// </summary>
+        ConflictingFeatureFilter
```
zhenlan: I understand where the 'Conflicting' is from. However, from the user's perspective, I specified a feature filter, which cannot be matched. It's not because the feature filter I specified conflicts with anyone, but because the feature filter I specified is ambiguous, for example, I should have namespaced it to be more specific etc. Therefore, I suggest we name this something like `AmbiguousMatchFeatureFilter`, similar to what .net has in [AmbiguousMatchException](https://docs.microsoft.com/en-us/dotnet/api/system.reflection.ambiguousmatchexception?view=netcore-3.1). The intellisense can be updated accordingly.

jimmyca15: I was considering ambiguous as well since that's what .NET throws when a multiple controller actions match. Good idea. I think we can shorten it to `AmbiguousFeatureFilter`

PR #56, opened 2020-04-13
Added targeting support with the addition of the TargetingFilter.

jimmyca15:
This PR introduces a new built-in feature filter to support a targeting strategy for feature management. With this new feature filter users will be able to target specific users, groups, and a user wide percentage when rolling out new features.

The readme contains an in-depth overview of targeting and the steps for using the new feature filter.

@mrm9084 who couldn't be found in reviewers drop down.


examples/TargetingConsoleApp/Program.cs
```diff
@@ -0,0 +1,75 @@
...
+
+            //
+            // Setup application services + feature management
+            IServiceCollection services = new ServiceCollection();
+
+            services.AddSingleton(configuration)
+                    .AddFeatureManagement()
+                    .AddFeatureFilter<ContextualTargetingFilter>();
+
+            IUserRepository userRepository = new InMemoryUserRepository();
```
zhenlan, line 34: Are some files missed from the PR?

examples/TargetingConsoleApp/Program.cs
```diff
@@ -0,0 +1,75 @@
...
+                {
+                    const string FeatureName = "Beta";
+
+                    //
+                    // Get user
+                    User user = await userRepository.GetUser(userId);
+
+                    //
+                    // Check if feature enabled
+                    //
```
zhenlan, line 59: NIT: remove extra line

examples/TargetingConsoleApp/Program.cs
zhenlan: Never mind. Found them in a different folder :)

---
In reply to: [409118459](https://github.com/microsoft/FeatureManagement-Dotnet/pull/56#discussion_r409118459) [](ancestors = 409118459)

src/Microsoft.FeatureManagement/Targeting/Audience.cs
```diff
@@ -0,0 +1,28 @@
...
+        /// </summary>
+        public List<string> Users { get; set; }
+
+        /// <summary>
+        /// Includes users in the audience based off a group rollout.
+        /// </summary>
+        public List<GroupRollout> Groups { get; set; }
+
+        /// <summary>
+        /// Includes users in the audience based off a percentage of the total possible audience.
```
zhenlan, line 24: Can we be clear the expected value is 0..1 instead of 0..100?

src/Microsoft.FeatureManagement/Targeting/GroupRollout.cs
```diff
@@ -0,0 +1,21 @@
...
+    /// </summary>
+    public class GroupRollout
+    {
+        /// <summary>
+        /// The name of the group.
+        /// </summary>
+        public string Name { get; set; }
+
+        /// <summary>
+        /// The percentage of the group that should be considered part of the rollout.
```
zhenlan, line 17: Clarify the expected value is 0..1

src/Microsoft.FeatureManagement/Targeting/Audience.cs
jimmyca15: I'm confused here. The expected value is in the range [0,100]. It sounds like you're saying we should do [0,1].

src/Microsoft.FeatureManagement/Targeting/ContextualTargetingFilter.cs
```diff
@@ -0,0 +1,114 @@
...
+        }
+
+
+        /// <summary>
+        /// Determines if a given context id should be targeted based off the provided percentage
+        /// </summary>
+        /// <param name="contextId">A context identifier that determines what the percentage is applicable for</param>
+        /// <param name="percentage">The total percentage of possible context identifiers that should be targeted</param>
+        /// <returns>A boolean representing if the context identifier should be targeted</returns>
+        private bool IsTargeted(string contextId, double percentage)
```
zhenlan, line 95: Should we fail/log error if the percentage is outside of 0..1?

src/Microsoft.FeatureManagement/Targeting/Audience.cs
zhenlan: Since it's double now (instead of int), it can be confusing about the range. Yes, I think we should do [0,1]. The portal can do the necessary conversion.

jimmyca15: I would be confused if it was anything other than out of 100 since that's what percentage means. per one hundred.

zhenlan: Ok, I don't feel too strongly about this. You do have a point. Can we just clarify the range in the comment?

jimmyca15: Absolutely, I should clarify it.

examples/FeatureFlagDemo/AssignUserMiddleware.cs
```diff
@@ -0,0 +1,62 @@
...
+
+namespace FeatureFlagDemo
+{
+    /// <summary>
+    /// This is a contrived middleware used to assign user information to the request based off of parameters passed in via the request's query string parameters.
+    /// Typically an application will authenticate a user and populate user information during that process.
+    /// 
+    /// To assign a user use the following query string structure "?username=JohnDoe&groups=MyGroup1,MyGroup2"
+    /// </summary>
+    public class AssignUserMiddleware
```
zhenlan, line 19: Normally, this will be done as part of authentication in a real-world app. Can we mimic that in the example too?

examples/FeatureFlagDemo/HttpContextTargetingContextAccessor.cs
```diff
@@ -0,0 +1,66 @@
...
+using System.Collections.Generic;
+using System.Security.Claims;
+using System.Threading.Tasks;
+
+namespace FeatureFlagDemo
+{
+    /// <summary>
+    /// Provides an implementation of <see cref="ITargetingContextAccessor"/> that creates a targeting context using info from the current HTTP request.
+    /// </summary>
+    public class HttpContextTargetingContextAccessor : ITargetingContextAccessor
```
zhenlan, line 16: Neat. Love it.

src/Microsoft.FeatureManagement.AspNetCore/AspNetCoreFeatureManagementBuilderExtensions.cs
```diff
@@ -14,6 +15,21 @@ namespace Microsoft.FeatureManagement
     /// </summary>
     public static class AspNetCoreFeatureManagementBuilderExtensions
     {
+        /// <summary>
+        /// Adds the Microsoft.Targeting feature filter to the service collection along with the <see cref="ITargetingContextAccessor"/> implementation that the filter will use to access the targeting context.
+        /// </summary>
+        /// <typeparam name="T">The <see cref="ITargetingContextAccessor"/> implementation type.</typeparam>
+        /// <param name="builder">The feature management builder to extend.</param>
+        /// <returns>The feature management builder.</returns>
+        public static IFeatureManagementBuilder AddTargetingFilter<T>(this IFeatureManagementBuilder builder) where T : ITargetingContextAccessor
```
zhenlan, line 24: It's a convenience function but it also hides some details that could cause confusion. Since it's only one extra line, maybe we can just do it as part of the example?

examples/FeatureFlagDemo/ClaimTypes.cs
```diff
@@ -0,0 +1,10 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+//
+namespace FeatureFlagDemo
+{
+    static class ClaimTypes
+    {
+        public static string GroupName = "http://schemas.featureflagdemo.featuremanagement.microsoft.com/claims/groupname";
```
MSGaryWang, line 8: >microsoft [](start = 91, length = 9)

we might not want to use microsoft.com here since user would just copy it

examples/FeatureFlagDemo/AssignUserMiddleware.cs
```diff
@@ -0,0 +1,62 @@
...
+            }
+
+            if (httpContext.Request.Query.TryGetValue("groups", out StringValues groupsValue))
+            {
+                string[] groups = groupsValue.First().Split(',');
+
+                foreach (string group in groups)
+                {
+                    identity.AddClaim(new Claim(ClaimTypes.GroupName, group));
+                }
```
MSGaryWang, line 52: should we log group, too?

examples/TargetingConsoleApp/Program.cs
MSGaryWang, line 34: >InMemoryUserRepository [](start = 49, length = 22)

I am thinking the idea to pass the user/group from command line or Thread.CurrentPrincipal?

jimmyca15: The purpose of the example is to show how to use the feature manager without relying on an ambient context.

jimmyca15: You could imagine that user references are being read in through a file and then full information on user is being pulled from database which in this case is abstracted through IUserRepository.

examples/FeatureFlagDemo/AssignUserMiddleware.cs
jimmyca15: Sure.

examples/FeatureFlagDemo/ClaimTypes.cs
jimmyca15: That should be fine. A look at [`System.Security.Claims.ClaimTypes`](https://docs.microsoft.com/en-us/dotnet/api/system.security.claims.claimtypes?view=netframework-4.8) shows plenty of claims that use microsoft.com. The only problem is there is no group name claim.

src/Microsoft.FeatureManagement/Targeting/ContextualTargetingFilter.cs
```diff
@@ -0,0 +1,114 @@
...
+            _logger = loggerFactory?.CreateLogger<ContextualTargetingFilter>() ?? throw new ArgumentNullException(nameof(loggerFactory));
+        }
+
+        /// <summary>
+        /// Performs a targeting evaluation using the provided <see cref="TargetingContext"/> to determine if a feature should be enabled.
+        /// </summary>
+        /// <param name="context">The feature evaluation context.</param>
+        /// <param name="targetingContext">The targeting context to use during targeting evaluation.</param>
+        /// <returns>True if the feature is enabled, false otherwise.</returns>
+        public Task<bool> EvaluateAsync(FeatureFilterEvaluationContext context, ITargetingContext targetingContext)
```
MSGaryWang, line 38: >context [](start = 71, length = 7)

null checking?

src/Microsoft.FeatureManagement/Targeting/ContextualTargetingFilter.cs
```diff
@@ -0,0 +1,114 @@
...
+                        {
+                            return Task.FromResult(true);
+                        }
+                    }
+                }
+            }
+
+            //
+            // Check if the user is being targeted by a default rollout percentage
+            string defaultContextId = $"{targetingContext.UserId}\n{context.FeatureName}";
```
MSGaryWang, line 83: there might be the case with no userid but group info, we might use both or one of them which exist

src/Microsoft.FeatureManagement/Targeting/ContextualTargetingFilter.cs
```diff
@@ -0,0 +1,114 @@
...
+                {
+                    GroupRollout groupRollout = settings.Audience.Groups.FirstOrDefault(g => g.Name.Equals(group));
+
+                    if (groupRollout != null)
+                    {
+                        string audienceContextId = $"{targetingContext.UserId}\n{context.FeatureName}\n{group}";
+
+                        if (IsTargeted(audienceContextId, groupRollout.RolloutPercentage))
+                        {
+                            return Task.FromResult(true);
```
MSGaryWang, line 75: >true [](start = 51, length = 4)

nit: the coding style inconsistent with line 85

src/Microsoft.FeatureManagement/Targeting/ContextualTargetingFilter.cs
```diff
@@ -0,0 +1,114 @@
...
+            {
+                return Task.FromResult(true);
+            }
+
+            //
+            // Check if the user is in a group that is being targeted
+            if (targetingContext.Groups != null &&
+                settings.Audience.Groups != null)
+            {
+                foreach (string group in targetingContext.Groups)
```
MSGaryWang, line 65: >foreach [](start = 16, length = 7)

This might confuse people by the order we search. For example in audience:
Group A -> 80%
Group B -> 20 %
and in user context we have Groups -> [ B, A ]
We will match 20% here but user might expect 80% since audience order is A first. 
Which I can expect if you have group A and you should match group A first instead of group B in audience order.

jimmyca15: The order doesn't matter though. There would be no distinguishable difference. We test both groups and if either one should result in the feature being on then it will be on.

src/Microsoft.FeatureManagement/Targeting/ContextualTargetingFilter.cs
jimmyca15: I'm not sure if it's a style difference as much as it is a logical difference. On this line we want to return true. On 85 we want to return whatever the fallback percentage targeting result is.

src/Microsoft.FeatureManagement/Targeting/ContextualTargetingFilter.cs
jimmyca15: UserId being null is the same as userId of empty string in this case. Groups don't play a factor in the default rollout percentage evaluation. It is based purely on user.

src/Microsoft.FeatureManagement/Targeting/ContextualTargetingFilter.cs
jimmyca15: FeatureFilterEvaluationContext is guaranteed to not be null. The second parameter could be null, but I'm wondering if we should change it to disallow null.

jimmyca15: I'm going to go ahead and add null check for both of them even though the state machine won't pass null for the first parameter. The method is better off establishing it's contract explicitly. Good suggestion.

jimmyca15:
Updated the PR.

@zhenlan one thing worth noting is that we now throw an exception if the targeting settings are invalid. This is to prevent silent misconfiguration. My thought was that users will be deploying their apps with different targeting configurations as they progress through a rollout and if they ever misconfigure they should find out in their staging environment when the targeting filter throws exception.

examples/FeatureFlagDemo/AssignUserMiddleware.cs
jimmyca15: Updated this to use an authentication handler and a demo authentication scheme, `QueryString`

src/Microsoft.FeatureManagement.AspNetCore/AspNetCoreFeatureManagementBuilderExtensions.cs
jimmyca15: Removed and readme updated as well.

zhenlan:
> using Microsoft.AspNetCore.Authentication;

We need to add copyright header to all new files.

---
Refers to: examples/FeatureFlagDemo/Authentication/QueryStringAuthenticationHandler.cs:1 in ee2e384. [](commit_id = ee2e384b0258feeb0e94ae5686cfaf4fb9a346ef, deletion_comment = False)

README.md
```diff
@@ -421,10 +421,119 @@ This filter provides the capability to enable a feature based on a time window.
...
+## Targeting in a Console Application
+
+The targeting filter relies on a targeting context to evaluate whether a feature should be turned on. This targeting context contains information such as what user is currently being evaluated, and what groups the user in. In console applications there is typically no ambient context available to flow this information in to the targeting filter, thus it must be passed directly when `FeatureManager.IsEnabledAsync` is called. This is supported through the use of the `ContextualTargetingFilter`. Applications that need to float the targeting context into the feature manager should use this instead of the `TargetingFilter.`
+
+```
+services.AddFeatureManagement()
+        .AddFeatureFilter<ContextualTargetingFilter>();
+```
+
+Since `ContextualTargetingFilter` is an [`IContextualTargetingFilter<ITargetingContext>`](./README.md#Contextual-Feature-Filters), an implementation of ITargetingContext must be passed in to `IFeatureManager.IsEnabledAsync` for it to be able to evaluate and turn a feature on.
```
zhenlan, line 507: NIT: Add backtick to `ITargetingContext `.

src/Microsoft.FeatureManagement/Targeting/ContextualTargetingFilter.cs
```diff
@@ -0,0 +1,186 @@
...
+
+                foreach (GroupRollout groupRollout in settings.Audience.Groups)
+                {
+                    index++;
+
+                    if (groupRollout.RolloutPercentage < 0 || groupRollout.RolloutPercentage > 100)
+                    {
+                        //
+                        // Audience.Groups[1].RolloutPercentage
+                        paramName = $"{settings.Audience}.{settings.Audience.Groups}[{index}].{groupRollout.RolloutPercentage}";
```
zhenlan, line 174: NIT: it may be easier for users debugging if we include `groupRollout.Name` in the message.

jimmyca15: @zhenlan It's a been more than I would expect in an argument exception. I'd be fine without it, but if you want it I'm not opposed. Let me know.

zhenlan: It's okay.

lukelengl:
I found this filter in the documentation, but it appears that it's still in preview. When will this be released?

zhenlan:
@lukangel did you get a chance to try it out? Does it work for your scenario? Any feedback (either positive or negative) will help us to reach the stable version sooner.

lukangel:
@zhenlan, it looks like you @ the wrong Luke :)  @lukelengl

zhenlan:
Oops. @lukangel thanks 😄.

lukelengl:
@zhenlan No, I did not get a chance to try it. After reading through the documentation, using a custom filter was better suited for my needs.

PR #62, opened 2020-04-21
Added test for targeting.

jimmyca15:
Tests were omitted from the targeting PR. This PR adds tests to ensure targeting is functional is not broken in the future.


jimmyca15:
@zhenlan I'm going to add one more test. Will send a new PR.

PR #63, opened 2020-04-21
Added test for targeting.

jimmyca15:
Tests were omitted from the targeting PR. This PR adds tests to ensure targeting is functional and not broken in the future.

### DIFF
--- a/tests/Tests.FeatureManagement/FeatureManagement.cs
+++ b/tests/Tests.FeatureManagement/FeatureManagement.cs
@@ -243,6 +243,68 @@ namespace Tests.FeatureManagement
         }
 
         [Fact]
+        public async Task Targeting()
+        {
+            IConfiguration config = new ConfigurationBuilder().AddJsonFile("appsettings.json").Build();
+
+            var services = new ServiceCollection();
+
+            services
+                .Configure<FeatureManagementOptions>(options =>
+                {
+                    options.IgnoreMissingFeatureFilters = true;
+                });
+
+            services
+                .AddSingleton(config)
+                .AddFeatureManagement()
+                .AddFeatureFilter<ContextualTargetingFilter>();
+
+            ServiceProvider serviceProvider = services.BuildServiceProvider();
+
+            IFeatureManager featureManager = serviceProvider.GetRequiredService<IFeatureManager>();
+
+            string beta = Enum.GetName(typeof(Features), Features.Beta);
+
+            //
+            // Targeted by user id
+            Assert.True(await featureManager.IsEnabledAsync(beta, new TargetingContext
+            {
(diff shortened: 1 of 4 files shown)


tests/Tests.FeatureManagement/FeatureManagement.cs
```diff
@@ -242,6 +242,115 @@ public async Task Percentage()
...
+            {
+                UserId = "Jeff"
+            }));
+
+            //
+            // Not targeted by user id, but targeted by default rollout
+            Assert.True(await featureManager.IsEnabledAsync(beta, new TargetingContext
+            {
+                UserId = "Anne"
+            }));
```
zhenlan, line 281: Wow, if we change the hash algorithm in the future, this one will be broken :)

tests/Tests.FeatureManagement/Features.cs
```diff
@@ -5,6 +5,7 @@ namespace Tests.FeatureManagement
 {
     enum Features
     {
+        Beta,
```
zhenlan, line 8: Everything looks good. Just one minor suggestion - change the flag name to something like `TargetingTestFeature`. It will make it easier to read and maintain the test code.

tests/Tests.FeatureManagement/FeatureManagement.cs
```diff
@@ -242,6 +242,115 @@ public async Task Percentage()
...
+            var targetingContextAccessor = new OnDemandTargetingContextAccessor();
+
+            services.AddSingleton<ITargetingContextAccessor>(targetingContextAccessor);
+
+            services
+                .AddSingleton(config)
+                .AddFeatureManagement()
+                .AddFeatureFilter<TargetingFilter>();
+
+            ServiceProvider serviceProvider = services.BuildServiceProvider();
```
leonids2005, line 329: Just looking at this PR. Probably this is minor but there is a code duplication in tests initialization.

tests/Tests.FeatureManagement/FeatureManagement.cs
jimmyca15: Yep, but I think it should break. It would be a breaking change to end users after all, and that is something we want to catch with tests.

tests/Tests.FeatureManagement/Features.cs
jimmyca15: Good suggestion, updated.

tests/Tests.FeatureManagement/FeatureManagement.cs
jimmyca15: It has the benefit of making the tests self contained. I'll leave it for now.

PR #64, opened 2020-04-21
Fixed build and pack based to respect updated .NET Core tooling behavior

jimmyca15:


build.ps1
```diff
@@ -35,7 +35,7 @@ if ($RestoreOnly)
 else
 {
     # Build
-    dotnet build -o "$BuildRelativePath" -c $BuildConfig "$Solution" | Tee-Object -FilePath "$LogDirectory\build.log"
+    dotnet build -o "$BuildRelativePath" -c $BuildConfig "$Solution" /p:OutDir=$BuildRelativePath | Tee-Object -FilePath "$LogDirectory\build.log"
```
abhilasharora, line 38: The command `dotnet build` already sets `OutDir` as the value of `-o` parameter. Do we still need `/p:OutDir=$BuildRelativePath`?

jimmyca15: It turns out that we are fine without specifying the `-o`. So we can remove that. We use the `OutDir` property in the pack command to say where the binaries are.

jimmyca15: Updated to remove -o.

PR #69, opened 2020-05-21
Clean up test and migrate to 3.1

ganesanarun:


examples/ConsoleApp/ConsoleApp.csproj
```diff
@@ -1,24 +1,24 @@
...
-  <PropertyGroup>
-    <OutputType>Exe</OutputType>
-    <TargetFramework>netcoreapp2.1</TargetFramework>
-    <RootNamespace>Consoto.Banking.AccountService</RootNamespace>
-  </PropertyGroup>
+	<PropertyGroup>
+		<OutputType>Exe</OutputType>
+		<TargetFramework>netcoreapp3.1</TargetFramework>
+		<RootNamespace>Consoto.Banking.AccountService</RootNamespace>
+	</PropertyGroup>
```
zhenlan, line 7: It looks like a lot of files are changed only because of the indentation change. Could you please undo all those changes?

ganesanarun: Sure, what is the indentation you use?

tabs or space? 4 or 2.

ganesanarun: It's hard for me to find now.  default settings format like this, if you could provide the information, it will be easy for me to have that for this project in .editorconfig and use it.

zhenlan: From the PR, you can see many files have only formatting changes. Undo those changes in git and then you should be able to tell what's the indentation used in current files.

ganesanarun: Interesting, I did not think of that

ganesanarun: Now you can see that, it's not just indentation changes, in fact number of files are increased now.

ganesanarun: Should I have to do something from side?

jimmyca15:
This PR is a bit big and also some of the changes I don't necessarily agree with. This would be a good candidate to split into multiple PRs also, to discuss first to achieve an agreeable outcome.

jimmyca15:
I do think migrating examples to .NET Core 3.1 is a good goal and would be worth an independent PR.

jimmyca15:
I'm going to close this PR as it is too big for a single merge. There are many separate concerns being addressed here. Smaller PRs with a single concern would be preferred.

PR #71, opened 2020-05-26
Make missing feature filter test clearer

JosephWoodward:
Looking through this test and there's a bunch of implied knowledge that's required which isn't immediately obvious. This PR fixes that (though that's assuming you'd rather test the default behaviour as opposed to explicitly setting the `IgnoreMissingFeatureFilters` property to `false`).

### DIFF
--- a/tests/Tests.FeatureManagement/FeatureManagement.cs
+++ b/tests/Tests.FeatureManagement/FeatureManagement.cs
@@ -337,6 +337,12 @@ namespace Tests.FeatureManagement
             var services = new ServiceCollection();
 
             services
+                .Configure<FeatureManagementOptions>(options =>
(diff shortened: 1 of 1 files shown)


zhenlan:
@JosephWoodward yes, the intent is to test the default behavior, so we don't release a breaking change accidentally if someone changes the default behavior. With this PR, we will not be able to catch that breaking change any more.

JosephWoodward:
> @JosephWoodward yes, the intent is to test the default behavior, so we don't release a breaking change accidentally if someone changes the default behavior. With this PR, we will not be able to catch that breaking change any more.

That’s understandable and I thought that might be the case, I’ll close this PR then. 👍

Issue #68, opened 2020-05-13
Misconfigured Flags Should Error


jason-roberts:
I think that if a flag is used and that flag is not configured then the app should error and not just default the flag to false.

For example, if a configured flag is for some reason omitted in a production config (or mis-spelled) then the app will continue to run without error and the fault may not be noticed.

If a flag is being used programmability then this could cause data issues.

If a flag is being used to control UI elements such as some legal wording that needs to be shown, then by defaulting to false the company could be in breach of laws.

zhenlan:
A potential solution is to allow customization of the behavior when a flag is not found. By default, the flag is considered false, but a user can change the default to true or throw.

JosephWoodward:
> A potential solution is to allow customization of the behavior when a flag is not found. By default, the flag is considered false, but a user can change the default to true or throw.

I agree, this seems like a better solution.

jimmyca15:
I believe we should add a new error type to [`FeatureManagementError`](https://github.com/microsoft/FeatureManagement-Dotnet/blob/master/src/Microsoft.FeatureManagement/FeatureManagementError.cs). The type can be `MissingFeature`. Once we have the new error, we can throw a `FeatureManagementException` when a call for a missing feature occurs. We can then offer a switch in [`FeatureManagementOptions`](https://github.com/microsoft/FeatureManagement-Dotnet/blob/master/src/Microsoft.FeatureManagement/FeatureManagementOptions.cs) to control the new behavior.

If the behavior is on by default then it is a breaking change so it would need to be done in a major version bump.

maplemale:
Any chance this can be bumped / moved forward? Looks there was a PR and is now possibly stale?

PR #72, opened 2020-05-28
Add delegate to capture misconfigured feature flags

JosephWoodward:
This pull request aims to address #68. I'm interested in your thoughts so any comments then I can update accordingly.

### DIFF
--- a/src/Microsoft.FeatureManagement/FeatureManagementOptions.cs
+++ b/src/Microsoft.FeatureManagement/FeatureManagementOptions.cs
@@ -1,6 +1,10 @@
 ﻿// Copyright (c) Microsoft Corporation.
 // Licensed under the MIT license.
 //
+
+using System;
+using System.Threading.Tasks;
+
 namespace Microsoft.FeatureManagement
 {
     /// <summary>
@@ -13,5 +17,11 @@ namespace Microsoft.FeatureManagement
         /// If missing feature filters are not ignored an exception will be thrown when attempting to evaluate a feature that depends on a missing feature filter.
         /// </summary>
         public bool IgnoreMissingFeatureFilters { get; set; }
+
+        /// <summary>
+        /// Enables notification of feature flags that are missing a configuration.
+        /// If a feature flag exists with no configuration then this delegate will be invoked.
+        /// </summary>
+        public Func<string, Task> OnMissingFeatureConfiguration { get; set; }
     }
 }
--- a/src/Microsoft.FeatureManagement/FeatureManagementOptions.cs
+++ b/src/Microsoft.FeatureManagement/FeatureManagementOptions.cs
@@ -22,6 +22,6 @@ namespace Microsoft.FeatureManagement
         /// Enables notification of feature flags that are missing a configuration.
         /// If a feature flag exists with no configuration then this delegate will be invoked.
         /// </summary>
-        public Func<string, Task> OnMissingFeatureConfiguration { get; set; }
+        public Func<string, Task> OnMissingFeature { get; set; }
     }
 }
--- a/src/Microsoft.FeatureManagement/FeatureManagementOptions.cs
+++ b/src/Microsoft.FeatureManagement/FeatureManagementOptions.cs
@@ -4,6 +4,7 @@
 
 using System;
 using System.Threading.Tasks;
+using Microsoft.Extensions.Logging;
 
 namespace Microsoft.FeatureManagement
 {
@@ -22,6 +23,6 @@ namespace Microsoft.FeatureManagement
         /// Override default behaviour of a missing feature
         /// If a feature flag exists with no configuration then this delegate will be invoked enabling you to capture the missing feature name and/or override the default behaviour.
         /// </summary>
-        public Func<string, Task<bool>> OnMissingFeature { get; set; }
+        public Func<string, ILogger, Task<bool>> OnMissingFeature { get; set; }
     }
 }
--- a/src/Microsoft.FeatureManagement/FeatureManagementOptions.cs
+++ b/src/Microsoft.FeatureManagement/FeatureManagementOptions.cs
@@ -19,9 +19,9 @@ namespace Microsoft.FeatureManagement
         public bool IgnoreMissingFeatureFilters { get; set; }
 
         /// <summary>
-        /// Enables notification of feature flags that are missing a configuration.
-        /// If a feature flag exists with no configuration then this delegate will be invoked.
+        /// Override default behaviour of a missing feature
+        /// If a feature flag exists with no configuration then this delegate will be invoked enabling you to capture the missing feature name and/or override the default behaviour.
         /// </summary>
-        public Func<string, Task> OnMissingFeature { get; set; }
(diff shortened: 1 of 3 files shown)


src/Microsoft.FeatureManagement/FeatureManagementOptions.cs
```diff
@@ -13,5 +17,11 @@ public class FeatureManagementOptions
         /// If missing feature filters are not ignored an exception will be thrown when attempting to evaluate a feature that depends on a missing feature filter.
         /// </summary>
         public bool IgnoreMissingFeatureFilters { get; set; }
+
+        /// <summary>
+        /// Enables notification of feature flags that are missing a configuration.
+        /// If a feature flag exists with no configuration then this delegate will be invoked.
+        /// </summary>
+        public Func<string, Task> OnMissingFeatureConfiguration { get; set; }
```
zhenlan, line 25: How about just `OnMissingFeature`?

The Func should return `Task<bool>`, so someone can actually use this and change the default behavior (for example, return true. BTW, we should have a test for this scenario.).

src/Microsoft.FeatureManagement/FeatureManager.cs
```diff
@@ -141,6 +141,13 @@ private async Task<bool> IsEnabledAsync<TContext>(string feature, TContext appCo
                     }
                 }
             }
+            else
+            {
+                if (_options.OnMissingFeatureConfiguration != null)
+                {
+                    await _options.OnMissingFeatureConfiguration(feature);
```
zhenlan, line 148: We need to call `.ConfigureAwait(false)` here.

tests/Tests.FeatureManagement/FeatureManagement.cs
```diff
@@ -25,6 +25,7 @@ public class FeatureManagement
         private const string OffFeature = "OffFeature";
         private const string ConditionalFeature = "ConditionalFeature";
         private const string ContextualFeature = "ContextualFeature";
+        private const string MissingConfiguration = "MissingConfiguration";
```
zhenlan, line 28: How about `MissingFeature`?

src/Microsoft.FeatureManagement/FeatureManagementOptions.cs
JosephWoodward: The name `OnMissingFeature` sounds good to me.

What would the boolean indicate? Whether or not to error? My initial thoughts were you could use this delegate to choose how a missing feature is handled. For instance I could choose to throw an exception within the delegate if I so wished.

tests/Tests.FeatureManagement/FeatureManagement.cs
JosephWoodward: Sounds good 👍

src/Microsoft.FeatureManagement/FeatureManagementOptions.cs
zhenlan: Sorry for my brevity in the previous comment. Currently, the feature is considered false by default. Returning `Task<bool>` will enable full customization of the default behavior. For example, someone can change to always consider a feature is 'On' if it's missing. Another one can even have more complex logic and decide dynamically whether a feature should be considered on or off if it's missing.

JosephWoodward: Great, thanks for the clarification, see my latest commit.

src/Microsoft.FeatureManagement/FeatureManager.cs
```diff
@@ -141,6 +141,13 @@ private async Task<bool> IsEnabledAsync<TContext>(string feature, TContext appCo
                     }
                 }
             }
+            else
+            {
+                if (_options.OnMissingFeature != null)
+                {
+                    return await _options.OnMissingFeature(feature).ConfigureAwait(false);
```
zhenlan, line 148: I would assign the value to `enabled` instead of returning directly from here. Short-circuiting here also skipped the session manager code below.

JosephWoodward: Oh yeah, completely missed that. Thanks.

src/Microsoft.FeatureManagement/FeatureManagementOptions.cs
```diff
@@ -13,5 +17,11 @@ public class FeatureManagementOptions
         /// If missing feature filters are not ignored an exception will be thrown when attempting to evaluate a feature that depends on a missing feature filter.
         /// </summary>
         public bool IgnoreMissingFeatureFilters { get; set; }
+
+        /// <summary>
+        /// Override default behaviour of a missing feature
+        /// If a feature flag exists with no configuration then this delegate will be invoked enabling you to capture the missing feature name and/or override the default behaviour.
+        /// </summary>
+        public Func<string, Task<bool>> OnMissingFeature { get; set; }
```
zhenlan, line 25: We probably should let the Func take an `ILogger` too, so someone can log warning or error even if they don't want to change the default behavior. Thoughts?

JosephWoodward: Yeah, it sounds like a good idea. It'll save people from having to create/register an `IConfigureOptions<FeatureManagementOptions>` to inject a logger.

jimmyca15:
@JosephWoodward @zhenlan why do we need a callback instead of a switch like the already existing: `IgnoreMissingFeatureFilters`? The issue that this is resolving mentions that a missing feature can be an exceptional state. I agree with that. Adding additional functionality via a callback might be a stretch though.

jimmyca15:
I do see some comments regarding the capability to set all missing features to true via the callback. I'm not sure what the use case for that would be though? In that case I would expect that a feature really should be configured to true. Depending on any behavior regarding missing features can be dangerous which is exactly what the original issue is mentioning.

JosephWoodward:
@jimmyca15

> why do we need a callback instead of a switch like the already existing: `IgnoreMissingFeatureFilters`?

A switch forces your application to either throw an exception or lose any visibility of a missing feature configuration. For instance, if someone were to inadvertently clean up the wrong feature configuration within my Cloud provider of choice then my application would start erroring. This PR proposes adding a callback to enable the consumer to choose how a missing feature configuration is interpreted (log it as an error/warning for instance) 

I have no strong opinion on the returned boolean though, personally I'd rather it always returned false.

jimmyca15:
@JosephWoodward I don't think the callback is the way to go here. Is it true that all calls to the feature manager for a missing feature should behave the same? Perhaps, perhaps not. It's not really standard usage for the library, and I would argue that it is an exceptional state. In that case I would think our best bet would be to stick with standard error propagation via an exception. With that approach, if any logging is needed, then the proper exception could be caught, logged, and rethrown. Or if the application wishes, it could be swallowed.

JosephWoodward:
@jimmyca15 I see. At the moment this library does not throw an exception if the configuration is missing so it's not such an issue. With this in mind what are your thoughts on the way forward with the initial issue posed in #68?

PR #73, opened 2020-05-30
Editor config

ganesanarun:
Editor config will help everyone to have same kind of code style.  
Avoid indentation issues in future.

Added editor config of roslyn https://raw.githubusercontent.com/dotnet/roslyn/master/.editorconfig


jimmyca15:
There is no ratified coding style in place for this library, also we are not planning on using the tool at the moment. Perhaps you could open an issue to facilitate a discussion if a coding style gate is desired and what tool could achieve it.

PR #77, opened 2020-06-10
Add netcoreapp3.1 as additional target for NuGet packages

AndreyTretyak:
With the release of .NET Core 3.0, many ASP.NET Core assemblies are no longer published to NuGet as packages. ([info](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/target-aspnetcore?view=aspnetcore-3.1&tabs=visual-studio))

So using libraries that target only `netstandard2.0` will lead to conflicts in `netcoreapp3.1` when you are getting old versions of packages as transient dependencies and you are unable to override them with a new version, because it's not published as NuGet package.
 
In order to solve this issue libraries add multiple targets, one for `netstandard2.0` and one `netcoreapp3.0`or higher (currently `3.1` since `3.0` no longer supported).

For example [Microsoft.Extensions.Logging](https://www.nuget.org/packages/Microsoft.Extensions.Logging/3.1.5) and other extension packages following this pattern.

This pull request adds `netcoreapp3.1` as a second target for feature management packages and for unit tests (ideally unit tests should also target `net472` or other version of full framework, but it will require more work).

### DIFF
--- a/src/Microsoft.FeatureManagement.AspNetCore/Microsoft.FeatureManagement.AspNetCore.csproj
+++ b/src/Microsoft.FeatureManagement.AspNetCore/Microsoft.FeatureManagement.AspNetCore.csproj
@@ -2,7 +2,7 @@
   <Import Project="..\..\build\NugetProperties.props" />
 
   <PropertyGroup>
-    <TargetFramework>netstandard2.0</TargetFramework>
+    <TargetFrameworks>netstandard2.0;netcoreapp3.1</TargetFrameworks>
     <SignAssembly>true</SignAssembly>
     <DelaySign>false</DelaySign>
     <AssemblyOriginatorKeyFile>..\..\build\Microsoft.FeatureManagement.snk</AssemblyOriginatorKeyFile>
@@ -17,15 +17,19 @@
     <PackageIconUrl>https://aka.ms/AzureAppConfigurationPackageIcon</PackageIconUrl>
     <Copyright>© Microsoft Corporation. All rights reserved.</Copyright>
   </PropertyGroup>
-  
+
   <ItemGroup>
     <ProjectReference Include="..\Microsoft.FeatureManagement\Microsoft.FeatureManagement.csproj" />
   </ItemGroup>
-  
-  <ItemGroup>
+
+  <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.0'" >
     <PackageReference Include="Microsoft.AspNetCore.Mvc.TagHelpers" Version="2.1.3" />
   </ItemGroup>
-  
+
+  <ItemGroup Condition="'$(TargetFramework)' != 'netstandard2.0'" >
+    <FrameworkReference Include="Microsoft.AspNetCore.App" />
+  </ItemGroup>
+
   <PropertyGroup>
     <DocumentationFile>bin\$(Configuration)\$(TargetFramework)\$(RuntimeIdentifier)\XMLComments\$(MSBuildProjectName).xml</DocumentationFile>
   </PropertyGroup>
--- a/src/Microsoft.FeatureManagement.AspNetCore/Microsoft.FeatureManagement.AspNetCore.csproj
+++ b/src/Microsoft.FeatureManagement.AspNetCore/Microsoft.FeatureManagement.AspNetCore.csproj
@@ -23,7 +23,7 @@
   </ItemGroup>
 
   <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.0'" >
-    <PackageReference Include="Microsoft.AspNetCore.Mvc.TagHelpers" Version="2.2.0" />
+    <PackageReference Include="Microsoft.AspNetCore.Mvc.TagHelpers" Version="2.1.3" />
   </ItemGroup>
 
   <ItemGroup Condition="'$(TargetFramework)' != 'netstandard2.0'" >
--- a/src/Microsoft.FeatureManagement.AspNetCore/Microsoft.FeatureManagement.AspNetCore.csproj
+++ b/src/Microsoft.FeatureManagement.AspNetCore/Microsoft.FeatureManagement.AspNetCore.csproj
@@ -23,7 +23,7 @@
   </ItemGroup>
 
   <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.0'" >
-    <PackageReference Include="Microsoft.AspNetCore.Mvc.TagHelpers" Version="2.1.3" />
+    <PackageReference Include="Microsoft.AspNetCore.Mvc.TagHelpers" Version="2.2.0" />
   </ItemGroup>
 
   <ItemGroup Condition="'$(TargetFramework)' != 'netstandard2.0'" >
--- a/src/Microsoft.FeatureManagement/Microsoft.FeatureManagement.csproj
+++ b/src/Microsoft.FeatureManagement/Microsoft.FeatureManagement.csproj
@@ -2,7 +2,7 @@
   <Import Project="..\..\build\NugetProperties.props" />
 
   <PropertyGroup>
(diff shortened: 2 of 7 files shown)


AndreyTretyak:
@jimmyca15 Could you please have a look?

src/Microsoft.FeatureManagement/Microsoft.FeatureManagement.csproj
```diff
@@ -2,7 +2,7 @@
   <Import Project="..\..\build\NugetProperties.props" />
 
   <PropertyGroup>
-    <TargetFramework>netstandard2.0</TargetFramework>
+    <TargetFrameworks>netstandard2.0;netstandard2.1</TargetFrameworks>
```
jimmyca15, line 5: Why target netstandard2.1 here, but netcoreapp3.1 in Microsoft.FeatureManagement.AspNetCore ?

AndreyTretyak: asp .net core starting from version 3 supported only in .net core (not in netstandard). Since Microsoft.FeatureManagement does not have asp .net core dependencies it could target netstandard (for Mono or Xamarin I guess), but when we need asp .net core libs it could be only net core.

jimmyca15: I see. I think it may be better to follow the precedent set by the packages that we are depending on. For example [Microsoft.Extensions.Logging](https://www.nuget.org/packages/Microsoft.Extensions.Logging/3.1.5) targets netstandard2.0 and netcoreapp3.1.

src/Microsoft.FeatureManagement/Microsoft.FeatureManagement.csproj
```diff
@@ -18,19 +18,22 @@
...
+  <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.0'">
+    <PackageReference Include="Microsoft.Bcl.AsyncInterfaces" Version="1.1.1" />
+  </ItemGroup>
+
   <ItemGroup>
-    <PackageReference Include="Microsoft.Bcl.AsyncInterfaces" Version="1.0.0" />
-    <PackageReference Include="Microsoft.Extensions.Configuration.Binder" Version="2.1.10" />
-    <PackageReference Include="Microsoft.Extensions.Logging" Version="2.1.1" />
+    <PackageReference Include="Microsoft.Extensions.Configuration.Binder" Version="3.1.5" />
+    <PackageReference Include="Microsoft.Extensions.Logging" Version="3.1.5" />
```
jimmyca15, line 28: Bumping the major version of dependencies can lead to breaking changes. We would need to bump the major version of our package as well. I'm also not sure how requiring 3.x.x packages versions is going to behave for those who are running ASP.NET Core 2.1. As far as I know, the 2.1.x packages are still supported.

jimmyca15: Let's just skip the major version bump for these. 

2.1.x is still supported, and those who need any new APIs exposed in 3.1.x packages will most likely already have references to them resulting in auto upgrade.

src/Microsoft.FeatureManagement/Microsoft.FeatureManagement.csproj
AndreyTretyak: Sure, no problem. I also think it's better, but decided not to limit availability without a valid reason.

src/Microsoft.FeatureManagement/Microsoft.FeatureManagement.csproj
AndreyTretyak: It should not affect this PR so it makes sense to don't update them here. And you are right that it could be overridden by users. I've reverted to the version that was used before. Thank you

AndreyTretyak: Also about NuGet package support, if I'm not mistaken, Logger and other Extension packages support are tied to .net core support, and since net core 3.0 already out of support those packed also not supported so I would suggest updating them (in a separate PR).

jimmyca15: .NET Core 2.1 is still in support so Microsoft.Extensions... 2.1.x packages are still supported.

src/Microsoft.FeatureManagement.AspNetCore/Microsoft.FeatureManagement.AspNetCore.csproj
```diff
@@ -17,15 +17,19 @@
...
+
   <ItemGroup>
     <ProjectReference Include="..\Microsoft.FeatureManagement\Microsoft.FeatureManagement.csproj" />
   </ItemGroup>
-  
-  <ItemGroup>
-    <PackageReference Include="Microsoft.AspNetCore.Mvc.TagHelpers" Version="2.1.3" />
+
+  <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.0'" >
+    <PackageReference Include="Microsoft.AspNetCore.Mvc.TagHelpers" Version="2.2.0" />
```
jimmyca15, line 26: We need to stick with 2.1.x versions even here for full supported integration with ASP.NET Core 2.1 apps. 2.1.3 should be fine since it is the latest 2.1.x patch version.

.NET Core 2.1 [is still supported](https://dotnet.microsoft.com/platform/support/policy/dotnet-core).

src/Microsoft.FeatureManagement/Microsoft.FeatureManagement.csproj
AndreyTretyak: Ah, sorry I've skipped the dates for net core 2.1

src/Microsoft.FeatureManagement.AspNetCore/Microsoft.FeatureManagement.AspNetCore.csproj
AndreyTretyak: Sorry skipped this one, version reverted to the previous value

AndreyTretyak: Are you okay with keeping package updates for test projects?

PawOrmstrupMadsen:
Hi, Any chance for a nuget package update with this included ? i very much need it - we are porting a large app to 3.1 from 2.2 and this is the last problematic nuget package we are using.

AndreyTretyak:
Hi @PawOrmstrupMadsen, I think after this change you should be able to use package in 3.1, although you might need to explicitly reference packages like `Microsoft.Extensions.Logging` to bump their versions. What packages are causing you problems?

PawOrmstrupMadsen:
i have exceptions in the analysers because there are refs to the old mvc 2.2 (https://www.nuget.org/packages/Microsoft.AspNetCore.Mvc.TagHelpers/)

AndreyTretyak:
I could see two possible reasons for this problem:
1. You are using FeatureManagement package versions without this change, I don't think there've been a release with it.
2. Project that using this packages targeting netstandard or full framework, since for netcore3.1 Microsoft.AspNetCore.Mvc.TagHelpers won't be a dependency.

So as soon as release with this change would be available and if your project target netcoreapp3.1 or higher problems should be resolved.

PawOrmstrupMadsen:
Yes - That's what i wrote :) i'm asking you if you can push an update to nuget with this merge included.

AndreyTretyak:
Ah sorry for the confusion, I've misunderstood you and thought you would like to bump versions that package using.
I think @jimmyca15 could be able to provide some release dates

jimmyca15:
@PawOrmstrupMadsen  I just sent out a PR, and once that gets in I plan on releasing a 2.2.0 preview. Trying to get the PR in quickly.

Issue #57, opened 2020-04-14
Provide for Distributed Systems


udlose:
**Feature Request:**
Based on documentation and blog articles I've read and from poking around some of the source code, the FeatureManagement system appears to be a wrapper on top of the Configuration System.  The Configuration System doesn't easily provide for setting up a configuration source that is distributed - e.g. Database, Distributed Cache, etc.

**Question:**
Is there a recommendation for accomplishing this? I would expect this to be something that is easily configurable thru the Feature Management API but I must be missing something.

JosephWoodward:
It sounds like this is somewhat related to [this question](https://github.com/microsoft/FeatureManagement-Dotnet/issues/46).

NicolasReyDotNet:
I have the exact same point, so I started to store my FeatureManagement json in a distributed system (Hashicorp Vault) instead of in a local Appsettings files, but I don't know how to refresh any changes.
I've just asked the [question](https://github.com/microsoft/FeatureManagement-Dotnet/issues/65).

udlose:
IMO, this is a configuration that should be available thru FeatureManagement without having to write custom code for this type of usage scenario.

JosephWoodward:
@udlose Agreed, hopefully the team will consider #46 and make it possible. 🤞

udlose:
> @udlose Agreed, hopefully the team will consider #46 and make it possible. 🤞

@JosephWoodward  The diffierence between #46 and this issue is that I'm asking for this functionality to be "built-in" to FeatureManagement so that I don't have to use `IFeatureSettingsProvider` to do a custom implementation myself.  There should be a "plug-n-play" configuration where we tell FeatureManagement where to look for the settings based on some defined interface, and then it handles the "monitoring, polling, etc." to look for changes and update itself accordingly.

Cloud-based solutions are so ubiquitous now that it's shocking that there isn't a solution here for distributed settings storage.

wapplegate:
So it looks like there is a PR for #46 it would be great if it could be merged and released soon.

jimmyca15:
This has been added in [the latest release](https://github.com/microsoft/FeatureManagement-Dotnet/releases/tag/2.2.0-preview).

NicolasREY69330:
@jimmyca15 
Great news! Will you update the doc providing some samples?

jimmyca15:
@NicolasREY69330 

[...]

Issue #46, opened 2020-01-31
Consider making IFeatureSettingsProvider public


pardahlman:
There are many system in the wild that uses other feature toggling frameworks that does not load its data from `IConfiguration`, nor has a data source that can be used as a ConfigurationProvider. The idea behind making `IFeatureSettingsProvider` public is that these systems could use a custom implementation as part of their migration story over to this toggle framework.

slomangino123:
Looking into a way to feature flag api endpoints easily I stumbled across the same issue as @pardahlman. As such we are looking for a way to toggle feature flags without using IConfiguration but rather a third party service. Being able to provide a decorated IFeatureSettingsProvider would allow sourcing feature flags from anywhere.

JosephWoodward:
Came here to say the same thing. It'd be great to be able to use this library for more than `IConfiguration` based datasources. Making this interface public would enable that.

jimmyca15:
This has been added in [the latest release](https://github.com/microsoft/FeatureManagement-Dotnet/releases/tag/2.2.0-preview).

PR #79, opened 2020-06-25
Added support for custom feature providers.

jimmyca15:
This PR introduces customization points to enable pulling feature definitions from anywhere. As mentioned in attached issues, this functionality will enable developers to pull feature definitions from existing legacy sources or external vendors that already support feature management.

New public API

`public IFeatureDefinitionProvider;`

`public FeatureDefinition;`

`public FeatureFilterConfiguration;`

## Usage

The readme has been updated to explain the steps required to wire up this new extensibility point. A test named `CustomSettingsProvider` has also been added that provides a small example of a custom feature definition provider. 

Relevant issues:
#57 
#46 
#8


tests/Tests.FeatureManagement/InMemoryFeatureSettingsProvider.cs
```diff
@@ -0,0 +1,33 @@
+﻿using Microsoft.FeatureManagement;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+
+namespace Tests.FeatureManagement
+{
+    class InMemoryFeatureSettingsProvider : IFeatureDefinitionProvider
```
zhenlan, line 9: NIT: following the new naming `InMemoryFeatureDefinitionsProvider`

tests/Tests.FeatureManagement/FeatureManagement.cs
```diff
@@ -483,5 +483,64 @@ public async Task SwallowsExceptionForMissingFeatureFilter()
 
             Assert.False(isEnabled);
         }
+
+        [Fact]
+        public async Task CustomSettingsProvider()
```
zhenlan, line 488: NIT: following the new naming `CustomFeatureDefinitionProvider`.

tests/Tests.FeatureManagement/FeatureManagement.cs
```diff
@@ -483,5 +483,64 @@ public async Task SwallowsExceptionForMissingFeatureFilter()
...
+                    }
+                }
+            };
+
+            var services = new ServiceCollection();
+
+            services
+                .Configure<FeatureManagementOptions>(options =>
+                {
+                    options.IgnoreMissingFeatureFilters = true;
```
zhenlan, line 511: This does not seem needed.

README.md
```diff
@@ -541,6 +541,28 @@ Feature state is provided by the IConfiguration system. Any caching and dynamic
...
+```
+public interface IFeatureDefinitionProvider
+{
+    Task<FeatureDefinition> GetFeatureDefinitionAsync(string featureName);
+
+    IAsyncEnumerable<FeatureDefinition> GetAllFeatureDefinitionsAsync();
+}
+```
+
+To use an implementation of `IFeatureDefinitionProvider` it must be added into the service collection before adding feature management. The following example adds an implementation of `IFeatureDefinitionProvider` named `InMemoryFeatureSettingsProvider`.
```
zhenlan, line 559: NIT: for consistent naming `InMemoryFeatureDefinitionProvider`.

src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
```diff
@@ -81,7 +81,7 @@ public async IAsyncEnumerable<FeatureSettings> GetAllFeatureSettingsAsync()
             }
         }
 
-        private FeatureSettings ReadFeatureSettings(string featureName)
+        private FeatureDefinition ReadFeatureSettings(string featureName)
```
zhenlan, line 84: Do you like to be consistent `ReadFeatureDefinition`?

jimmyca15: Yes, thank you. I'll update it.

README.md
jimmyca15: Good point. Let me change it.

tests/Tests.FeatureManagement/FeatureManagement.cs
jimmyca15: Good point, it doesn't. I have removed it.

jimmyca15:
@zhenlan thanks for the review. I have updated it according to your feedback.

src/Microsoft.FeatureManagement/FeatureManager.cs
```diff
@@ -16,7 +16,7 @@ namespace Microsoft.FeatureManagement
     /// </summary>
     class FeatureManager : IFeatureManager
     {
-        private readonly IFeatureSettingsProvider _settingsProvider;
+        private readonly IFeatureDefinitionProvider _settingsProvider;
```
zhenlan, line 19: NIT: `_featureDefinitionProvider`?

src/Microsoft.FeatureManagement/FeatureManager.cs
```diff
@@ -25,7 +25,7 @@ class FeatureManager : IFeatureManager
         private readonly FeatureManagementOptions _options;
 
         public FeatureManager(
-            IFeatureSettingsProvider settingsProvider,
+            IFeatureDefinitionProvider settingsProvider,
```
zhenlan, line 28: NIT: `featureDefinitionProvider`?

src/Microsoft.FeatureManagement/FeatureManager.cs
```diff
@@ -90,7 +90,7 @@ private async Task<bool> IsEnabledAsync<TContext>(string feature, TContext appCo
                     // For all enabling filters listed in the feature's state calculate if they return true
                     // If any executed filters return true, return true
 
-                    foreach (FeatureFilterSettings featureFilterSettings in settings.EnabledFor)
+                    foreach (FeatureFilterConfiguration featureFilterSettings in settings.EnabledFor)
```
zhenlan, line 93: NIT: `featureFilterConfiguration`?

src/Microsoft.FeatureManagement/FeatureManager.cs
```diff
@@ -72,7 +72,7 @@ private async Task<bool> IsEnabledAsync<TContext>(string feature, TContext appCo
 
             bool enabled = false;
 
-            FeatureSettings settings = await _settingsProvider.GetFeatureSettingsAsync(feature).ConfigureAwait(false);
+            FeatureDefinition settings = await _settingsProvider.GetFeatureDefinitionAsync(feature).ConfigureAwait(false);
```
zhenlan, line 75: NIT: `settings` -> `featureDefinition`?

src/Microsoft.FeatureManagement/FeatureFilterConfiguration.cs
```diff
@@ -7,12 +7,12 @@
 namespace Microsoft.FeatureManagement
 {
     /// <summary>
-    /// The settings that define a feature filter.
+    /// The configuration of a feature filter.
     /// </summary>
-    class FeatureFilterSettings
+    public class FeatureFilterConfiguration
```
zhenlan, line 12: I am wondering if it will be better if we call this `FeatureFilterDefinition`.

README.md
```diff
@@ -541,6 +541,28 @@ Feature state is provided by the IConfiguration system. Any caching and dynamic
...
+}
+```
+
+To use an implementation of `IFeatureDefinitionProvider` it must be added into the service collection before adding feature management. The following example adds an implementation of `IFeatureDefinitionProvider` named `InMemoryFeatureDefinitionProvider`.
+
+```
+services.AddSingleton<IFeatureDefinitionProvider, InMemoryFeatureDefinitionProvider>()
+        .AddFeatureManagement()
+```
+
```
zhenlan, line 565: Do you like the idea that we add these two APIs so we won't always add `ConfigurationFeatureDefinitionProvider` to DI and make it easier for users? Or is this too much and we can always add in the future?
```
AddFeatureManagement<IFeatureDefinitionProvider>()
AddFeatureManagement(IFeatureDefinitionProvider fd)
```

src/Microsoft.FeatureManagement/FeatureFilterConfiguration.cs
jimmyca15: The reason I left it as Configuration rather than definition is that there is only one definition for a feature. However the same feature filter might appear in multiple features. They would just be configured differently. So to me, this was the configuration of a feature filter rather than a definition of one.

README.md
jimmyca15: I'd go with add in the future for that one.

jimmyca15: Especially if in the future we get a request to provide the capability to pull feature settings from an external source. Then it will point out that maybe the API that you proposed could help with the discoverability that we already offer that option.

jimmyca15:
@zhenlan

I updated the settings variables to use the new definition name. Thanks for that.

I still couldn't come up with a better name for `FeatureFilterConfiguration`, let me know if you have any idea on that or if it sounds right.

tests/Tests.FeatureManagement/InMemoryFeatureDefinitionProvider.cs
```diff
@@ -0,0 +1,33 @@
...
+    {
+        private IEnumerable<FeatureDefinition> _definitions;
+
+        public InMemoryFeatureDefinitionProvider(IEnumerable<FeatureDefinition> featureDefinitions)
+        {
+            _definitions = featureDefinitions ?? throw new ArgumentNullException(nameof(featureDefinitions));
+        }
+
+#pragma warning disable CS1998 // Async method lacks 'await' operators and will run synchronously
+        public async IAsyncEnumerable<FeatureDefinition> GetAllFeatureDefinitionsAsync()
```
zhenlan, line 19: We should add a test for this API too.

README.md
zhenlan: Agreed. Makes sense.

zhenlan:
> I still couldn't come up with a better name for `FeatureFilterConfiguration`, let me know if you have any idea on that or if it sounds right.

Let's go with it then!

tests/Tests.FeatureManagement/InMemoryFeatureDefinitionProvider.cs
jimmyca15: I agree. Having a test for our `ConfigurationFeatureDefinitionProvider.GetAllFeatureDefinitionsAsync` would be handy. I'll add one, but in a different PR to prevent making this one bigger.

jimmyca15: I went back and looked and we already have a test for that code path: https://github.com/microsoft/FeatureManagement-Dotnet/blob/fddc98b1ef7fd2efbd5d362bc5194c80bab95a8c/tests/Tests.FeatureManagement/FeatureManagement.cs#L430 

It uses our built-in implementation of `IFeatureDefinitionProvider`, but there's no difference between our implementation of `IFeatureDefinitionProvider`, and an external one. The `CustomFeatureDefinitionProvider` test ensures that an external `IFeatureDefinitionProvider` can be wired up. So it seems to me like the two tests together provide the coverage we need.

* `IFeatureDefinitionProvider` methods tested
* custom `IFeatureDefinitionProvider` is registered and consumed successfully

PR #83, opened 2020-07-10
Updated build pipeline to use .NET Core 3.1 sdk.

jimmyca15:
The change to target multiple frameworks resulted in a couple required changes to get build and pack working as expected.

* Use .NET Core 3.1 sdk container image
* Remove output path specification from build/pack command
  * This causes the builds for each target framework to overwrite one another resulting in build errors and mispackaged binaries
* Updated build and pack to determine what PowerShell should be used
  * .NET Core 3.1 container image comes with `pwsh` built-in

### DIFF
--- a/build/ChoosePowerShell.cmd
+++ b/build/ChoosePowerShell.cmd
@@ -1,0 +1,23 @@
+:: where.exe does not exist in windows container, application specific test must be used to check for existence
+
+PowerShell -Command Write-Host "a"
+
+IF %ERRORLEVEL% == 0 (
+
+    set PowerShell=PowerShell
+
+    exit /B 0
+)
+
+pwsh -Command Write-Host "a"
+
+IF %ERRORLEVEL% == 0 (
+
+    set PowerShell=pwsh
+
+    exit /B 0
+)
+
+echo Could not find a suitable PowerShell executable.
+
+EXIT /B 1
--- a/build/ChoosePowerShell.cmd
+++ b/build/ChoosePowerShell.cmd
@@ -1,19 +1,19 @@
 :: where.exe does not exist in windows container, application specific test must be used to check for existence
 
-PowerShell -Command Write-Host "a"
-
-IF %ERRORLEVEL% == 0 (
-
-    set PowerShell=PowerShell
-
-    exit /B 0
-)
-
 pwsh -Command Write-Host "a"
 
 IF %ERRORLEVEL% == 0 (
 
     set PowerShell=pwsh
+
+    exit /B 0
+)
+
+PowerShell -Command Write-Host "a"
+
+IF %ERRORLEVEL% == 0 (
+
+    set PowerShell=PowerShell
 
     exit /B 0
 )
--- a/build/InstallPowerShellCore.cmd
+++ b/build/InstallPowerShellCore.cmd
@@ -1,5 +1,0 @@
-if not exist "%USERPROFILE%\.dotnet\tools\pwsh.exe" (
-  dotnet tool install --global PowerShell --version 6.2.3
-)
-
-set PowerShellCore="%USERPROFILE%\.dotnet\tools\pwsh.exe"
--- a/pack.ps1
+++ b/pack.ps1
@@ -8,11 +8,13 @@ Note: build.cmd should be run before running this script.
 
 [CmdletBinding()]
 param(
+    [Parameter()]
+    [ValidateSet('Debug','Release')]
+    [string]$BuildConfig = "Release"
 )
 
 $ErrorActionPreference = "Stop"
 
-$PrebuiltBinariesDir = "bin\BuildOutput"
 $PublishRelativePath = "bin\PackageOutput"
 $LogDirectory = "$PSScriptRoot\buildlogs"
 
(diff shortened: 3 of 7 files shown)


build/InstallPowerShellCore.cmd
```diff
@@ -1,5 +0,0 @@
-if not exist "%USERPROFILE%\.dotnet\tools\pwsh.exe" (
-  dotnet tool install --global PowerShell --version 6.2.3
-)
-
-set PowerShellCore="%USERPROFILE%\.dotnet\tools\pwsh.exe"
```
abhilasharora, line 5: Is the pre-installed `pwsh.exe` at a different path in the new container image?

build/ChoosePowerShell.cmd
```diff
@@ -0,0 +1,23 @@
+:: where.exe does not exist in windows container, application specific test must be used to check for existence
+
+PowerShell -Command Write-Host "a"
```
abhilasharora, line 3: Is there a scenario where we would use the `PowerShell` instead of `pwsh`?

build/InstallPowerShellCore.cmd
jimmyca15: It's on the PATH in the new container image. We can treat the file path of `pwsh.exe` as opaque now.

build/ChoosePowerShell.cmd
jimmyca15: Building on dev machine. Our dev machines typically have powershell by default but not pwsh. Wouldn't want to block someone from cloning and building.

build/InstallPowerShellCore.cmd
jimmyca15: I just checked to see if it happened to be at the same or different path. Looks like the pre-installed pwsh.exe is not at `%USERPROFILE%\.dotnet\tools\pwsh.exe`

build/ChoosePowerShell.cmd
```diff
@@ -0,0 +1,23 @@
...
+
+IF %ERRORLEVEL% == 0 (
+
+    set PowerShell=PowerShell
+
+    exit /B 0
+)
+
+pwsh -Command Write-Host "a"
```
abhilasharora, line 12: I believe we are printing `a` only to validate if `pwsh` is a valid command in `PATH`. How about using `WHERE pwsh` to make the intent more clear?

build/ChoosePowerShell.cmd
abhilasharora: Okay, makes sense. Since `pwsh` is a newer version, should we attempt to use it first, followed by `PowerShell` in case the former isn't available?

pack.ps1
```diff
@@ -32,9 +34,7 @@ foreach ($project in $targetProjects)
     $projectPath = "$PSScriptRoot\src\$project\$project.csproj"
     $outputPath = "$PSScriptRoot\src\$project\$PublishRelativePath"
 
-    #
-    # The build system expects pre-built binaries to be in the folder pointed to by 'OutDir'.
-    dotnet pack -o "$outputPath" /p:OutDir=$PrebuiltBinariesDir "$projectPath" --no-build | Tee-Object -FilePath "$LogDirectory\build.log"
+    dotnet pack -c $BuildConfig -o "$outputPath" "$projectPath" --no-build | Tee-Object -FilePath "$LogDirectory\build.log"
```
zhenlan, line 37: I'm a bit confused. Since we say `--no-build` here, why do we need a `-c $BuildConfig`?

zhenlan: Oh, is that just to tell `dotnet pack` where to pick up binaries? Okay then.

build/ChoosePowerShell.cmd
```diff
@@ -0,0 +1,23 @@
+:: where.exe does not exist in windows container, application specific test must be used to check for existence
+
+PowerShell -Command Write-Host "a"
+
+IF %ERRORLEVEL% == 0 (
+
+    set PowerShell=PowerShell
```
zhenlan, line 7: What's the case `PowerShell` is available?

pack.ps1
abhilasharora: Yes, that's correct. The default configuration is `Debug` when using `--no-build`, so we explicitly override it. https://github.com/dotnet/sdk/issues/9570

build/ChoosePowerShell.cmd
zhenlan: Never mind. Just saw you answered the same question from Abhilash.

build/ChoosePowerShell.cmd
jimmyca15: We could, but I'm not sure why we would. If the order was important then the design to use whichever one is available would be flawed.

build/ChoosePowerShell.cmd
jimmyca15: The comment at the top of the file mentions that where does not exist in the container.

abhilasharora: Okay, I would have expected it in the `System32` directory. Thanks for checking!

build/ChoosePowerShell.cmd
abhilasharora: My understanding is if the user has both `PowerShell 4.0` and `pwsh 6.0` installed, then this script would use the former, even though the latter is a newer version. In this case, the order does influence the version used for execution.

zhenlan: More people will have 3.1 SDK installed on their dev box over time, so people are more likely to use the same "tool" on dev and build machines. It's good to be consistent. Maybe it's a good idea to print out what is picked for easier debugging in case they do present a behavior difference one day.

jimmyca15: That's right. But I'm trying to see if there is a concern with using PowerShell 4.0 even if pwsh 6.0 is installed. If there was, then the approach that I went with that either one could be used for building is invalid.

jimmyca15: I'll reverse the order. We use pwsh in the official build. If the official build fails for any reason and we suspect pwsh as the cause then it will be easier to debug on dev machine if pwsh is first.

Issue #90, opened 2020-08-05
Feature names with colon in name are not parsed correctly


peterwurzinger:
Given the following configuration:
```json
"FeatureManagement": {
    "Feature1": true,
    "Prefix:Feature2": true
}
```
Querying the features was a bit surprising at first:
```cs
featureManager.IsEnabledAsync("Feature1") //returns true
featureManager.IsEnabledAsync("Prefix:Feature2") //returns false
```
Listing the feature names is also quite misleading:
```cs
featureManager.GetFeatureNamesAsync(); //returns "Feature1", "Prefix"
```

By taking a closer look, the colon is of course not only a gimmick, but more like a common separator for config sections. So when using the JSON file provider, the abovementioned configuration could also be semantically identical rephrased as
```json
"FeatureManagement": {
    "Feature1": true,
    "Prefix": {
        "Feature2": true
    }
}
```
And that explains the weird behavior.

To solve that, would it be a possible to tokenize given feature names by `:` , treating them as config-sections and only the last segment as the name? Imho that would fix the `"Prefix":Feature" : true` case, as well as enable feature grouping (see above).

jimmyca15:
In general, the ':' character should be avoided in setting names in json settings files in .NET Core. The ':' character is treated specially as you have called out. I don't think it would be a good idea to go against that concept. It may be a good idea to call this out in the readme though.

peterwurzinger:
Well, since `IsEnabledAsync("Prefix:Feature")` will **never** behave as expected, would it then be an option to throw an exception, if the input string contains a colon?

jimmyca15:
That's a great suggestion. Technically if someone is using a custom `IFeatureDefinitionProvider` then colon might be fine for them, but if the default `ConfigurationFeatureDefinitionProvider` is being used then we can throw an `ArgumentException` to prevent confusion in a case like this.

zhiyuanliang-ms:
https://github.com/microsoft/FeatureManagement-Dotnet/pull/288
Resolved.

PR #95, opened 2020-09-14
Prevent usage of ':' in feature names.

jimmyca15:
Utilizing ':' in feature names causes problems due to how the .NET Core configuration system treats this character. This can lead to confusing situations when a developer tries to check whether a feature name that contains a ':' character is enabled.

This PR updates the `ConfigurationFeatureDefinitionProvider` to throw an `ArgumentException` if ':' is passed in with a feature name to prevent developers from hitting these issues. Custom implementations of `IFeatureDefinitionProvider` are free to support features with ':'.

#90


zhiyuanliang-ms:
Move to https://github.com/microsoft/FeatureManagement-Dotnet/pull/288

src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
```diff
@@ -48,6 +48,11 @@ public Task<FeatureDefinition> GetFeatureDefinitionAsync(string featureName)
                 throw new ArgumentNullException(nameof(featureName));
             }
 
+            if (featureName.Contains(ConfigurationPath.KeyDelimiter))
+            {
+                throw new ArgumentException($"The value '{ConfigurationPath.KeyDelimiter}' is not allowed in the feature name.", nameof(featureName));
```
jimmyca15, line 53: If our intent is truly to block ':', which I believe it is, then I think we should define the ':' ourselves. Right now the disallowed char is dependent upon the value of ConfigurationPath.KeyDelimiter. Even if that were to change, we wouldn't necessarily be ready to change our implementation here.

PR #101, opened 2020-10-24
Fixed race conditions in FeatureManagerSnapshot

gfoidl:
There are some race conditions in `IsEnabledAsync`. This PR fixes them.

### DIFF
--- a/src/Microsoft.FeatureManagement/FeatureManagerSnapshot.cs
+++ b/src/Microsoft.FeatureManagement/FeatureManagerSnapshot.cs
@@ -13,8 +13,8 @@ namespace Microsoft.FeatureManagement
     class FeatureManagerSnapshot : IFeatureManagerSnapshot
     {
         private readonly IFeatureManager _featureManager;
-        private readonly IDictionary<string, bool> _flagCache = new Dictionary<string, bool>();
-        private IEnumerable<string> _featureNames;
+        private readonly Dictionary<string, Task<bool>> _flagCache = new Dictionary<string, Task<bool>>();
+        private List<string> _featureNames;
 
         public FeatureManagerSnapshot(IFeatureManager featureManager)
         {
@@ -25,52 +25,77 @@ namespace Microsoft.FeatureManagement
         {
             if (_featureNames == null)
             {
-                var featureNames = new List<string>();
-
-                await foreach (string featureName in _featureManager.GetFeatureNamesAsync().ConfigureAwait(false))
-                {
-                    featureNames.Add(featureName);
-                }
-
-                _featureNames = featureNames;
+                _featureNames = await CreateListAsync(_featureManager).ConfigureAwait(false);
             }
 
             foreach (string featureName in _featureNames)
             {
                 yield return featureName;
             }
+
+            static async ValueTask<List<string>> CreateListAsync(IFeatureManager featureManager)
+            {
+                var featureNames = new List<string>();
+
+                await foreach (string featureName in featureManager.GetFeatureNamesAsync().ConfigureAwait(false))
+                {
+                    featureNames.Add(featureName);
+                }
+
+                return featureNames;
+            }
         }
 
-        public async Task<bool> IsEnabledAsync(string feature)
+        public Task<bool> IsEnabledAsync(string feature)
         {
-            //
-            // First, check local cache
-            if (_flagCache.ContainsKey(feature))
+            if (_flagCache.TryGetValue(feature, out Task<bool> task))
             {
-                return _flagCache[feature];
+                return task;
             }
 
-            bool enabled = await _featureManager.IsEnabledAsync(feature).ConfigureAwait(false);
+            return Core();
 
-            _flagCache[feature] = enabled;
+            Task<bool> Core()
+            {
+                lock (_flagCache)
+                {
+                    if (_flagCache.TryGetValue(feature, out task))
+                    {
+                        return task;
+                    }
 
-            return enabled;
+                    task = _featureManager.IsEnabledAsync(feature);
+                    _flagCache.Add(feature, task);
+
+                    return task;
+                }
+            }
         }
 
-        public async Task<bool> IsEnabledAsync<TContext>(string feature, TContext context)
+        public Task<bool> IsEnabledAsync<TContext>(string feature, TContext context)
         {
-            //
-            // First, check local cache
-            if (_flagCache.ContainsKey(feature))
+            if (_flagCache.TryGetValue(feature, out Task<bool> task))
             {
-                return _flagCache[feature];
+                return task;
             }
 
-            bool enabled = await _featureManager.IsEnabledAsync(feature, context).ConfigureAwait(false);
+            return Core();
 
-            _flagCache[feature] = enabled;
+            Task<bool> Core()
+            {
+                lock (_flagCache)
+                {
+                    if (_flagCache.TryGetValue(feature, out task))
+                    {
+                        return task;
+                    }
 
-            return enabled;
+                    task = _featureManager.IsEnabledAsync(feature, context);
+                    _flagCache.Add(feature, task);
+
+                    return task;
+                }
+            }
         }
     }
 }
--- a/src/Microsoft.FeatureManagement/FeatureManagerSnapshot.cs
+++ b/src/Microsoft.FeatureManagement/FeatureManagerSnapshot.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT license.
 //
 using System;
+using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Threading.Tasks;
 
@@ -13,7 +14,7 @@ namespace Microsoft.FeatureManagement
     class FeatureManagerSnapshot : IFeatureManagerSnapshot
     {
         private readonly IFeatureManager _featureManager;
-        private readonly Dictionary<string, Task<bool>> _flagCache = new Dictionary<string, Task<bool>>();
+        private readonly ConcurrentDictionary<string, Task<bool>> _flagCache = new ConcurrentDictionary<string, Task<bool>>();
         private List<string> _featureNames;
 
         public FeatureManagerSnapshot(IFeatureManager featureManager)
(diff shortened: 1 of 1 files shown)


src/Microsoft.FeatureManagement/FeatureManagerSnapshot.cs
```diff
@@ -24,53 +24,78 @@ public FeatureManagerSnapshot(IFeatureManager featureManager)
         public async IAsyncEnumerable<string> GetFeatureNamesAsync()
         {
             if (_featureNames == null)
+            {
+                _featureNames = await CreateListAsync(_featureManager).ConfigureAwait(false);
```
gfoidl, line 28: Didn't fix this race, as
* I assume this method isn't hot
* in the case of a race the list is created multiple time and `_featureNames` will get overwritten (in atomic way due being a reference), so unnecessary lists are just thrown away and are eligible for GC

src/Microsoft.FeatureManagement/FeatureManagerSnapshot.cs
```diff
@@ -13,8 +13,8 @@ namespace Microsoft.FeatureManagement
     class FeatureManagerSnapshot : IFeatureManagerSnapshot
     {
         private readonly IFeatureManager _featureManager;
-        private readonly IDictionary<string, bool> _flagCache = new Dictionary<string, bool>();
-        private IEnumerable<string> _featureNames;
+        private readonly Dictionary<string, Task<bool>> _flagCache = new Dictionary<string, Task<bool>>();
```
gfoidl, line 16: `ConcurrentDictionary<K,V>` would be possible, but the factory in `GetOrAdd` becomes the overload with state for .NET 5, which isn't used here. So in order to avoid a closure (= allocation) a plain `Dictionary<K,V>` is used, that is locked only in the case when no entry is found (so on the cold-path).

gfoidl (commented):
Some notes for review.

src/Microsoft.FeatureManagement/FeatureManagerSnapshot.cs
jimmyca15: Why did this method get touched?

src/Microsoft.FeatureManagement/FeatureManagerSnapshot.cs
jimmyca15: I would still use `ConcurrentDictionary<K,V>` here to resolve any race conditions. They are much better at fine-grain locking than what we could do here in a simple manner. Also, introducing any lock management will automatically make this code more complex to maintain. `ConcurrentDictionary` should provide us the simplest fix, although it may not have the optimization that you mentioned.

src/Microsoft.FeatureManagement/FeatureManagerSnapshot.cs
```diff
@@ -24,53 +24,78 @@ public FeatureManagerSnapshot(IFeatureManager featureManager)
...
             {
-                return _flagCache[feature];
+                return task;
             }
 
-            bool enabled = await _featureManager.IsEnabledAsync(feature, context).ConfigureAwait(false);
+            return Core();
 
-            _flagCache[feature] = enabled;
+            Task<bool> Core()
```
jimmyca15, line 84: This looks like an abuse of inline functions. The implementation could be taken out and replace the `return Core();` line with no resulting code duplication.

src/Microsoft.FeatureManagement/FeatureManagerSnapshot.cs
gfoidl: Will update to the CDN.
Just re-checked and the optimization I meant got added to .NET Core 2.0 and .NET Standard 2.1, so it's usable here 😃

src/Microsoft.FeatureManagement/FeatureManagerSnapshot.cs
gfoidl: I can't remember the exact motivation for the change, but in the case `_featureNames != null` the method will be way shorter, as when the lazy-init logic is part of the default-codepath. Aka hot-/cold-path optimization.

Back then when I made the change, I just did some optimization, so maybe I was a bit too aggressive in optimizing. Now I don't know if this change is worth it.

Revert this one?

src/Microsoft.FeatureManagement/FeatureManagerSnapshot.cs
gfoidl: This is outdated with 24faebd55c9ccd91a5f764dd90d9824dbbe09cc0.

src/Microsoft.FeatureManagement/FeatureManagerSnapshot.cs
jimmyca15: Yes, I would revert this one.

src/Microsoft.FeatureManagement/FeatureManagerSnapshot.cs
```diff
@@ -25,52 +26,49 @@ public async IAsyncEnumerable<string> GetFeatureNamesAsync()
...
-            // First, check local cache
-            if (_flagCache.ContainsKey(feature))
-            {
-                return _flagCache[feature];
-            }
-
-            bool enabled = await _featureManager.IsEnabledAsync(feature, context).ConfigureAwait(false);
-
-            _flagCache[feature] = enabled;
+#if NETSTANDARD2_0
```
jimmyca15, line 52: Adding ifdefs increases the code complexity. I'm not sure it's worth it here. I'd opt for just sticking with the NETSTANDARD2_0 implementation.

src/Microsoft.FeatureManagement/FeatureManagerSnapshot.cs
gfoidl: Reverted with b4855a9ea2f9982f839ec9c06655e1e89c061b2f

src/Microsoft.FeatureManagement/FeatureManagerSnapshot.cs
gfoidl: With NETSTANDARD2_0 we have repeated allocations for the delegate, that the overload with state (starting from NS 2.1) can avoid. See this [sharplab](https://sharplab.io/#v2:EYLgtghgzgLgpgJwD4AEBMBGAsAKBQBgAIUMA6AYQHsAbauAYxgEtKA7KCt+gVwQTlYwA3LgLEMAFhF4iJAKzTcrCGDhQADhHpxCAWSb0ElKJQBmMUgDE4EGLzi6IygOZxVg3AG9chX8TSE1rb2ji6IAMrK6lAAFpQwAPoAcuFoRCCEAJJBdvyhEK4IkRDRcTA+ft44fjWE6ghMAG62OvwQACZs1ACeWTkhTgWIhAmmNrkOg4XStX71TS2EbZ2sPYRUrDx8AjAAIgbMbBAI3QA8JPgANMQAbKfAlDQAfE8jptQF5FoxOgC8hKw4AB3dZcXj8QT7RgsZQnc4YK63e6PagvAAUAEoZrNCBVZigAMxIh7PLJQACiymAdHaAEEoN1NmiLoQxsF+Bi8bUqjjaigAOxvD7OL70H6kADicBgAHkELT2u00WyJtdjs5CL9XqNxgMwghSJkKVSafTGfQ0eqMViuTUAL5ch3VPxc9CBXV5KYRKKxeLJcgEjCEDLZD2TfXFUrxLk8nHzZrwJY2FZrUPs8NDBBvMP5aa2uYNBOtZNdXobLYQvYHGHHM4Xa4oO4k1Ha96fb5/AHA0GbcE7KGHWF1hENpso9E252zfO+QnE8dkykQalwOkMpkslX2TlT7kzvmC0bC0XiqWy+WKtH73lb/iXa84y0IZzXUxgDGa15vw3G5em9cWla967ryNQ6umuaIJOvJOvaXKukSTCCIgphaDoaYTJBRQ+mUwZ9DmXoIDGjrwbuc5IfACCodo+EQYRxEgX4jbIqSRpLiua7msyCKsmG0F+E6TpAA=).

I believe it's worth it. 
Say it's used in a controller and checked for each request. So we can save #requests delegate allocations.

jimmyca15: Can't we avoid having the ifdefs intermingled with our business logic here by having an extension method with the 3.1 signature and having the ifdef there. I don't expect any unnecessary closures as a result. If we can do that I'm fine with it.

jimmyca15: Then again, optimizing for perf without numbers doesn't sound like the best strategy. For this it's hard to say how much we'd be saving in any case without numbers.

gfoidl: > optimizing for perf without numbers 

The whole rationale of this API on ConcurrentDictionary is about saving the allocation of the delegate and potentially the closure. I don't need numbers for this, and won't do any benchmarks. So much discussion for this change...and the code reads quite nice with these two #ifdefs (IMHO). An extension method will just complicate things, as a) where to place it and b) it just creates more complicate closures.

I'll leave this PR as is. If it doesn't meet the bar, so please close it. Sorry that I have to say it that way.

jimmyca15: I stay away from ifdef without a strong reason to do so since I favor code maintenance and readability even though it may provide an optimization here for netcoreapp3.1. I would opt to remove it. It may just be me. @MSGaryWang @zhenlan for their insight.

gfoidl: What do you think about https://github.com/gfoidl/FeatureManagement-Dotnet/commit/aaa47c3ab3f374ffedfd46706d5d02e939fad770
It's just one #ifdef, by passing `null` in as context (the FeatureManager does the same, so this has no side-effects).

jimmyca15: I see, here's my thoughts.
* If we are going to use ifdef I certainly do like the idea of limiting it to one place instead of multiple.
* This does have side-effects. The code no longer behaves the same way. [FeatureManager](https://github.com/microsoft/FeatureManagement-Dotnet/blob/2240941480f833fea27601368eb910260ce71966/src/Microsoft.FeatureManagement/FeatureManager.cs#L45) has different code paths when `IsEnabledAsync` is called with or without a context. Even if `null` is passed.
* I still would opt out of adding an ifdef in general, but would like to defer to one of the mentioned collaborators.

src/Microsoft.FeatureManagement/FeatureManagerSnapshot.cs
```diff
@@ -13,8 +14,8 @@ namespace Microsoft.FeatureManagement
     class FeatureManagerSnapshot : IFeatureManagerSnapshot
     {
         private readonly IFeatureManager _featureManager;
-        private readonly IDictionary<string, bool> _flagCache = new Dictionary<string, bool>();
-        private IEnumerable<string> _featureNames;
+        private readonly ConcurrentDictionary<string, Task<bool>> _flagCache = new ConcurrentDictionary<string, Task<bool>>();
```
rjgotten, line 17: FWIW this isn't actually race proof. `ConcurrentDictionary.GetOrAdd` has atomic publish semantics; not atomic execution semantics. If you want to prevent duplicate executions for costly feature flag evaluations, you have to cache a `Lazy<>` with its mode set to `ExecutionAndPublication` instead.

Basically:
```cs
public Task<bool> IsEnabledAsync(string feature)
{
  return _flagCache
    .GetOrAdd(feature, (key, arg) => new Lazy<Task<bool>>(
      async () => await arg.IsEnabledAsync(key).ConfigureAwait(false),
      LazyThreadSafetyMode.ExecutionAndPublication
    ), _featureManager )
    .Value;
}
```

jimmyca15: Thanks @rjgotten for the suggested improvement.

jimmyca15:
This PR has not had any recent activity. In an effort to address #136, a new PR, #141, has been created. Any future discussion can be moved there.

PR #102, opened 2020-10-28
Fixed issue with the captive dependency

oholubovskyi:
I stumbled with the issue during the implementation of my custom `IFeatureDefinitionProvider`.

For `CustomFeatureDefinitionProvider : IFeatureDefinitionProvider` I need to retrieve the list of available features from the feature database. `IFeatureManager` registered as a singleton, it cause an issue with the `DbContext` injection. `CustomFeatureDefinitionProvider` registered as scoped, but `DbContext` in `CustomFeatureDefinitionProvider` injected from Application Root Scope instead of Request Scope, which is not correct (`DbContext` acts like a singleton in this case, even if its lifetime defined as Scoped in the `Startup`).

Registering `IFeatureManager` with the Scoped lifetime allow us to avoid the captive dependency issue, when IoC-container use wrong (base(application root) instead of current(request)) scope for the instantiation of services.

### DIFF
--- a/src/Microsoft.FeatureManagement/ServiceCollectionExtensions.cs
+++ b/src/Microsoft.FeatureManagement/ServiceCollectionExtensions.cs
@@ -26,7 +26,7 @@ namespace Microsoft.FeatureManagement
             // Add required services
             services.TryAddSingleton<IFeatureDefinitionProvider, ConfigurationFeatureDefinitionProvider>();
 
-            services.AddSingleton<IFeatureManager, FeatureManager>();
+            services.AddScoped<IFeatureManager, FeatureManager>();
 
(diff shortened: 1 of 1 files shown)


jimmyca15:
That's a fundamental design change and I'm not convinced it's necessary. I would imagine there would be a way for you to access the scoped services inside of your `CustomFeatureDefinitionProvider`. For example, if I am writing a singleton and want access to a request-scoped service I can depend on IHttpContextAccessor and use `IHttpContextAccessor.HttpContext.RequestServices`

L022714 (approved):

zhiyuanliang-ms:
@oholubovskyi  We provided `AddScopedFeatureManagement` API in 3.1.0. Also, the `FeatureManager` class is public now.

Issue #108, opened 2020-12-04
Any plans to support .net 5?


Sergey-Terekhin:
Hello!
Do you have any plans  to support released .net 5? And if yes, then?
Thank you

jimmyca15:
To clarify the ask here, you just mean to add net5.0 as a target framework of the package right? Or are you experiencing some issues currently?

If that is the ask, I can answer that yes we plan to add net5.0 as a target just as netcoreapp3.1 has been added. The timeline would be early in the new year.

jimmyca15:
This was added in 2.3.0 release.
https://github.com/microsoft/FeatureManagement-Dotnet/releases/tag/2.3.0

PR #112, opened 2021-01-07
Add net5.0 target framework.

jimmyca15:
Add net5.0 target framework.

* Add target framework
* Update build image to have 5.0 sdk
* Updated tests to run all supported target frameworks.

### DIFF
--- a/tests/Tests.FeatureManagement/FeatureManagement.cs
+++ b/tests/Tests.FeatureManagement/FeatureManagement.cs
@@ -4,6 +4,7 @@
 using Microsoft.AspNetCore;
 using Microsoft.AspNetCore.Builder;
 using Microsoft.AspNetCore.Hosting;
+using Microsoft.AspNetCore.Mvc;
 using Microsoft.AspNetCore.TestHost;
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.DependencyInjection;
@@ -84,7 +85,7 @@ namespace Tests.FeatureManagement
 
                     services.AddMvcCore(o =>
                     {
-                        o.EnableEndpointRouting = false;
+                        DisableEndpointRouting(o);
                         o.Filters.AddForFeature<MvcFilter>(ConditionalFeature);
                     });
                 })
@@ -132,7 +133,7 @@ namespace Tests.FeatureManagement
                         .AddFeatureManagement()
                         .AddFeatureFilter<TestFilter>();
 
-                    services.AddMvcCore(o => o.EnableEndpointRouting = false);
+                    services.AddMvcCore(o => DisableEndpointRouting(o));
                 })
             .Configure(app => app.UseMvc()));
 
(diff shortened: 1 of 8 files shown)


tests/Tests.FeatureManagement/Tests.FeatureManagement.csproj
```diff
@@ -18,13 +18,28 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.AspNetCore.TestHost" Version="3.1.5" />
-    <PackageReference Include="Microsoft.Extensions.Configuration.Json" Version="3.1.5" />
     <PackageReference Include="Microsoft.NET.Test.Sdk" Version="16.6.1" />
```
avanigupta, line 21: If Microsoft.NET.Test.Sdk Version="16.6.1" is applicable only for netcoreapp3.1 and net5.0, should we move it to their respective PackageReference sections?

tests/Tests.FeatureManagement/Tests.FeatureManagement.csproj
```diff
@@ -18,13 +18,28 @@
...
     <PackageReference Include="xunit" Version="2.4.1" />
     <PackageReference Include="xunit.runner.visualstudio" Version="2.4.1" />
   </ItemGroup>
 
+  <ItemGroup Condition="'$(TargetFramework)' == 'netcoreapp2.1'">
+    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="15.9.0" />
```
abhilasharora, line 27: Are there any build issues when using version `16.6.1` or newer for `netcoreapp2.1`?

tests/Tests.FeatureManagement/FeatureManagement.cs
```diff
@@ -84,7 +85,7 @@ public async Task Integrates()
 
                     services.AddMvcCore(o =>
                     {
-                        o.EnableEndpointRouting = false;
```
abhilasharora, line 87: How does enabling or disabling this option affect the behavior of our test project?

tests/Tests.FeatureManagement/Tests.FeatureManagement.csproj
jimmyca15: Yes I think it would probably be better that way.

tests/Tests.FeatureManagement/Tests.FeatureManagement.csproj
jimmyca15: Not build issues, but there were runtime issues. The test wouldn't even start. I looked back at what we were using as a reference when we were using netcoreapp2.1 as the test runtime and it was indeed 15.9.0. When I reverted the reference to that for netcoreapp2.1 it worked correctly.

tests/Tests.FeatureManagement/FeatureManagement.cs
jimmyca15: The tests were originally written to use the ASP.NET Core 2.1 test host, so that use the original routing strategy instead of endpoint routing which was introduced in ASP.NET Core 3.1. 3.1 had a breaking change to where you can't call `UseMvc` anymore without explicitly setting `EnableEndpointRouting` to false. So any test that calls `UseMvc` needs to have this.

tests/Tests.FeatureManagement/Tests.FeatureManagement.csproj
jimmyca15: Updated.

abhilasharora (approved):

avanigupta (approved):
:shipit:

jimmyca15:
In response to #108

PR #113, opened 2021-01-25
Updated Microsoft.FeatureManagement to 2.3.0 for release.

jimmyca15:
* Updated Microsoft.FeatureManagement to 2.3.0 for release.
  * Release adds .net 5 as a target.

* Updated build to use separate versions for different projects.

### DIFF
--- a/build/Versioning.props
+++ b/build/Versioning.props
@@ -1,0 +1,28 @@
+<Project>
+
+  <!-- Local Development -->
+  <PropertyGroup Condition="'$(CDP_BUILD_TYPE)'==''">
+    <MinorVersion>9999</MinorVersion>
+    <PatchVersion>9999</PatchVersion>
+  </PropertyGroup>
+
+  <!-- Buddy builds -->
+  <PropertyGroup Condition="'$(CDP_BUILD_TYPE)'!='' AND '$(CDP_BUILD_TYPE)'!='Official'">
+    <!-- Note that CDP_PACKAGE_VERSION_NUMERIC uses format Major.Minor.MMddyyrrrr, which causes compile error CS7034 because .NET Core doesn't allow version numbers higher than 65534. -->
+    <!-- CDP_PATCH_NUMBER is updated daily by the build system, the addition of a revision number makes the build number unique by the minute -->
+    <!-- The revision number is the number of minutes since midnight [0, 1440) -->
+    <!-- Example build number, 1.0.573.1227 -->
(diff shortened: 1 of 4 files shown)


build/Versioning.props
```diff
@@ -0,0 +1,28 @@
...
+    <TodayTicks>$([System.DateTime]::Today.Ticks)</TodayTicks>
+    <TicksSinceMidnight>$([MSBuild]::Subtract($(NowTicks), $(TodayTicks)))</TicksSinceMidnight>
+    <MinutesSinceMidnight>$([MSBuild]::Divide($(TicksSinceMidnight), 600000000))</MinutesSinceMidnight>
+    <Floored>$([System.Math]::Floor($(MinutesSinceMidnight)))</Floored>
+    <Revision>$(Floored)</Revision>
+    <PreviewVersion>-preview-$(CDP_PATCH_NUMBER)-$(Revision)</PreviewVersion>
+  </PropertyGroup>
+
+  <PropertyGroup>
+    <Version>$(MajorVersion).$(MinorVersion).$(PatchVersion)$(PreviewVersion)</Version>
```
abhilasharora, line 25: `PreviewVersion` seems to be undefined for official builds.

jimmyca15: That's because the current version is not a preview. If it was then we could put it under `<PatchVersion` and the official build would have a preview suffix.

abhilasharora: This line runs for all builds, but doesn't seem to have any value for official builds. Have you tested it for an official build?

jimmyca15: Yes, when a variable is undefined it's value is empty string. So for official build since the current version doesn't have a preview we end up with 2.3.0

avanigupta:
Just to confirm - adding .NET 5 as target will not be a breaking change, correct?

jimmyca15:
> Just to confirm - adding .NET 5 as target will not be a breaking change, correct?

The only behavior that could change as a result of adding a target for .NET 5 would be in apps running on  .NET 5. But, the fact that we aren't targeting it right now means we haven't really established support for it yet. 

Another thing, we added 3.1 targeting in the [update from 2.1 to 2.2](https://github.com/microsoft/FeatureManagement-Dotnet/releases/tag/2.2.0-preview)

avanigupta (approved):

build/Versioning.props
abhilasharora: Okay, if it is empty string, then it should work as expected. Thanks!

abhilasharora (approved):

PR #131, opened 2021-07-22
Add cancellation token parameter to async feature management interfaces.

jimmyca15:
This PR adds a cancellation token parameter to async feature management interfaces. Adding this parameter is a breaking change and requires a major version bump. The project versions have been updated to 3.0.0-preview.

### DIFF
--- a/src/Microsoft.FeatureManagement.AspNetCore/Microsoft.FeatureManagement.AspNetCore.csproj
+++ b/src/Microsoft.FeatureManagement.AspNetCore/Microsoft.FeatureManagement.AspNetCore.csproj
@@ -3,10 +3,9 @@
 
   <!-- Official Version -->
   <PropertyGroup>
-    <MajorVersion>3</MajorVersion>
-    <MinorVersion>0</MinorVersion>
+    <MajorVersion>2</MajorVersion>
+    <MinorVersion>3</MinorVersion>
     <PatchVersion>0</PatchVersion>
-    <PreviewVersion>-preview</PreviewVersion>
   </PropertyGroup>
 
   <Import Project="..\..\build\Versioning.props" />
--- a/src/Microsoft.FeatureManagement.AspNetCore/Microsoft.FeatureManagement.AspNetCore.csproj
+++ b/src/Microsoft.FeatureManagement.AspNetCore/Microsoft.FeatureManagement.AspNetCore.csproj
@@ -3,9 +3,10 @@
 
   <!-- Official Version -->
   <PropertyGroup>
-    <MajorVersion>2</MajorVersion>
-    <MinorVersion>3</MinorVersion>
+    <MajorVersion>3</MajorVersion>
+    <MinorVersion>0</MinorVersion>
     <PatchVersion>0</PatchVersion>
+    <PreviewVersion>-preview</PreviewVersion>
   </PropertyGroup>
 
   <Import Project="..\..\build\Versioning.props" />
--- a/src/Microsoft.FeatureManagement.AspNetCore/TagHelpers/FeatureTagHelper.cs
+++ b/src/Microsoft.FeatureManagement.AspNetCore/TagHelpers/FeatureTagHelper.cs
@@ -4,6 +4,7 @@
 using Microsoft.AspNetCore.Razor.TagHelpers;
 using System.Collections.Generic;
 using System.Linq;
+using System.Threading;
 using System.Threading.Tasks;
 
 namespace Microsoft.FeatureManagement.Mvc.TagHelpers
@@ -55,8 +56,8 @@ namespace Microsoft.FeatureManagement.Mvc.TagHelpers
                 IEnumerable<string> names = Name.Split(',').Select(n => n.Trim());
 
                 enabled = Requirement == RequirementType.All ?
-                    await names.All(async n => await _featureManager.IsEnabledAsync(n).ConfigureAwait(false)) :
-                    await names.Any(async n => await _featureManager.IsEnabledAsync(n).ConfigureAwait(false));
+                    await names.All(async n => await _featureManager.IsEnabledAsync(n, CancellationToken.None).ConfigureAwait(false)) :
+                    await names.Any(async n => await _featureManager.IsEnabledAsync(n, CancellationToken.None).ConfigureAwait(false));
             }
 
             if (Negate)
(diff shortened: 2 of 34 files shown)


src/Microsoft.FeatureManagement.AspNetCore/TagHelpers/FeatureTagHelper.cs
```diff
@@ -55,8 +56,8 @@ public override async Task ProcessAsync(TagHelperContext context, TagHelperOutpu
                 IEnumerable<string> names = Name.Split(',').Select(n => n.Trim());
 
                 enabled = Requirement == RequirementType.All ?
-                    await names.All(async n => await _featureManager.IsEnabledAsync(n).ConfigureAwait(false)) :
-                    await names.Any(async n => await _featureManager.IsEnabledAsync(n).ConfigureAwait(false));
+                    await names.All(async n => await _featureManager.IsEnabledAsync(n, CancellationToken.None).ConfigureAwait(false)) :
+                    await names.Any(async n => await _featureManager.IsEnabledAsync(n, CancellationToken.None).ConfigureAwait(false));
             }
```
avanigupta, line 61: Are we missing another `ConfigureAwait(false)` here?

build/NugetProperties.props
```diff
@@ -6,10 +6,10 @@
 
   <!-- Official Version -->
   <PropertyGroup>
-    <MajorVersion>2</MajorVersion>
-    <MinorVersion>2</MinorVersion>
+    <MajorVersion>3</MajorVersion>
```
avanigupta, line 9: I think you have some old code in this branch. The new versions now exist in .csproj files of both projects.

avanigupta:
In the AspNetCore project, we have another public interface: `IDisabledFeaturesHandler`. Do we need to add CancellationToken for its `HandleDisabledFeatures` method?

avanigupta:
>             TargetingContext targetingContext = await _contextAccessor.GetContextAsync().ConfigureAwait(false);

Does this `GetContextAsync` method need CancellationToken?

---
Refers to: src/Microsoft.FeatureManagement/Targeting/TargetingFilter.cs:52 in 8067cd6. [](commit_id = 8067cd623db037a37d545e7b434d2f7793a27594, deletion_comment = False)

src/Microsoft.FeatureManagement.AspNetCore/TagHelpers/FeatureTagHelper.cs
jimmyca15: Yes, thanks. I have added it.

build/NugetProperties.props
jimmyca15: You're right. I rebased and am now using the new versioning placement.

src/Microsoft.FeatureManagement.AspNetCore/Microsoft.FeatureManagement.AspNetCore.csproj
```diff
@@ -3,9 +3,10 @@
 
   <!-- Official Version -->
   <PropertyGroup>
-    <MajorVersion>2</MajorVersion>
-    <MinorVersion>3</MinorVersion>
+    <MajorVersion>3</MajorVersion>
+    <MinorVersion>0</MinorVersion>
     <PatchVersion>0</PatchVersion>
+    <PreviewVersion>-preview</PreviewVersion>
```
avanigupta, line 9: We should only change the version of `Microsoft.FeatureManagement` package first. Once that has been released to NuGet, we can update the version in `Microsoft.FeatureManagement.AspNetCore`.

jimmyca15:
> In the AspNetCore project, we have another public interface: `IDisabledFeaturesHandler`. Do we need to add CancellationToken for its `HandleDisabledFeatures` method?

It has ActionExecutingContext already which gives cancellation via HttpContext.RequestAborted. It is meant to be similar to [OnActionExecutionAsync ](https://docs.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.mvc.filters.actionfilterattribute.onactionexecutionasync?view=aspnetcore-5.0)which behaves similarly.

src/Microsoft.FeatureManagement.AspNetCore/Microsoft.FeatureManagement.AspNetCore.csproj
jimmyca15: Thanks, updated.

avanigupta (approved):

src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
```diff
@@ -40,7 +41,7 @@ public void Dispose()
             _changeSubscription = null;
         }
 
-        public Task<FeatureDefinition> GetFeatureDefinitionAsync(string featureName)
+        public Task<FeatureDefinition> GetFeatureDefinitionAsync(string featureName, CancellationToken _)
```
WeihanLi, line 44: Maybe better to have a proper parameter name like `cancellationToken`

src/Microsoft.FeatureManagement/IContextualFeatureFilter.cs
```diff
@@ -18,7 +19,8 @@ public interface IContextualFeatureFilter<TContext> : IFeatureFilterMetadata
         /// </summary>
         /// <param name="featureFilterContext">A feature filter evaluation context that contains information that may be needed to evaluate the filter. This context includes configuration, if any, for this filter for the feature being evaluated.</param>
         /// <param name="appContext">A context defined by the application that is passed in to the feature management system to provide contextual information for evaluating a feature's state.</param>
+        /// <param name="cancellationToken">The cancellation token to cancel the operation.</param>
         /// <returns>True if the filter's criteria has been met, false otherwise.</returns>
-        Task<bool> EvaluateAsync(FeatureFilterEvaluationContext featureFilterContext, TContext appContext);
+        Task<bool> EvaluateAsync(FeatureFilterEvaluationContext featureFilterContext, TContext appContext, CancellationToken cancellationToken);
```
WeihanLi, line 24: Maybe keep a default value to simplify the operation when no need for cancelling?

src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
jimmyca15: I agree. I think the intent of using a discard here to indicate synchronous completion doesn't really fit will. I'll send a PR to update.

src/Microsoft.FeatureManagement/IContextualFeatureFilter.cs
jimmyca15: I think that's a good suggestion. Will add it to the PR to address the cancellation token parameter names.

PR #138, opened 2021-08-31
Revert "Added default value for cancellation token in interfaces."

jimmyca15:
Reverts microsoft/FeatureManagement-Dotnet#133


avanigupta (approved):

jimmyca15:
v3 work will be done in feature/v3 branch.

PR #139, opened 2021-08-31
Revert "Add cancellation token parameter to async feature management interfaces."

jimmyca15:
Reverts microsoft/FeatureManagement-Dotnet#131


avanigupta (approved):

jimmyca15:
v3 work will be done in feature/v3 branch.

PR #140, opened 2021-09-01
Added option to throw when attempting to evaluate a missing feature.

jimmyca15:
Addresses #68

### DIFF
--- a/src/Microsoft.FeatureManagement/FeatureManagementOptions.cs
+++ b/src/Microsoft.FeatureManagement/FeatureManagementOptions.cs
@@ -11,12 +11,14 @@ namespace Microsoft.FeatureManagement
         /// <summary>
         /// Controls the behavior of feature evaluation when dependent feature filters are missing.
         /// If missing feature filters are not ignored an exception will be thrown when attempting to evaluate a feature that depends on a missing feature filter.
+        /// The default value is false.
         /// </summary>
         public bool IgnoreMissingFeatureFilters { get; set; }
 
         /// <summary>
         /// Controls the behavior of feature evaluation when the target feature is missing.
         /// If missing features are not ignored an exception will be thrown when attempting to evaluate them.
+        /// The default value is true.
         /// </summary>
         public bool IgnoreMissingFeatures { get; set; } = true;
     }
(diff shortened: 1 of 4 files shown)


src/Microsoft.FeatureManagement/FeatureManagementOptions.cs
```diff
@@ -13,5 +13,11 @@ public class FeatureManagementOptions
         /// If missing feature filters are not ignored an exception will be thrown when attempting to evaluate a feature that depends on a missing feature filter.
         /// </summary>
         public bool IgnoreMissingFeatureFilters { get; set; }
+
+        /// <summary>
+        /// Controls the behavior of feature evaluation when the target feature is missing.
+        /// If missing features are not ignored an exception will be thrown when attempting to evaluate them.
+        /// </summary>
```
zhenlan, line 20: This comment will be published in the documentation. We should mention the default is true.

Since you are on this, can you add the default value in comments for IgnoreMissingFeatureFilters above too?

jimmyca15: Good idea.

jimmyca15: Added.

zhenlan (approved):

src/Microsoft.FeatureManagement/FeatureManager.cs
```diff
@@ -141,6 +141,19 @@ private async Task<bool> IsEnabledAsync<TContext>(string feature, TContext appCo
                     }
                 }
             }
+            else
+            {
+                string errorMessage = $"The feature declaration for the feature '{feature}' was not found.";
+
+                if (!_options.IgnoreMissingFeatures)
+                {
+                    throw new FeatureManagementException(FeatureManagementError.MissingFeatureFilter, errorMessage);
```
avanigupta, line 150: Should this error type be `FeatureManagementError.MissingFeature`?

jimmyca15: Yes, thanks. Here's a PR to fix: https://github.com/microsoft/FeatureManagement-Dotnet/pull/142

Issue #136, opened 2021-08-24
FeatureManagementSnapshot is not thread safe


asuciu1:
Hello,

I have a problem using this in Azure Functions as I am occasionally getting the following. 

``System.InvalidOperationException: Operations that change non-concurrent collections must have exclusive access. A concurrent update was performed on this collection and corrupted its state. The collection's state is no longer correct.    
System.Collections.Generic.Dictionary`2.TryInsert(TKey key, TValue value, InsertionBehavior behavior)    
System.Collections.Generic.Dictionary`2.set_Item(TKey key, TValue value)    
Microsoft.FeatureManagement.FeatureManagerSnapshot.IsEnabledAsync[TContext](String feature, TContext context)``

The way this is implemented is the following:
- When the function is called the Feature Manager is injected as a scoped service with an expiration date of x seconds
- At the expiration date the values are refreshed
- As the features are set in a non thread safe dictionary the aforementioned exception is occasionally thrown when `IsEnabledAsync()` is called

Thanks for your help!

PR #141, opened 2021-09-01
Support concurrency in IFeatureManagerSnapshot

jimmyca15:
The current implementation of IFeatureManagerSnapshot is not thread safe. This PR enhances the implementation to support concurrency.

Addresses #136

### DIFF
--- a/src/Microsoft.FeatureManagement/FeatureManagerSnapshot.cs
+++ b/src/Microsoft.FeatureManagement/FeatureManagerSnapshot.cs
@@ -2,7 +2,9 @@
 // Licensed under the MIT license.
 //
 using System;
+using System.Collections.Concurrent;
 using System.Collections.Generic;
+using System.Threading;
 using System.Threading.Tasks;
 
 namespace Microsoft.FeatureManagement
@@ -13,7 +15,7 @@ namespace Microsoft.FeatureManagement
     class FeatureManagerSnapshot : IFeatureManagerSnapshot
     {
         private readonly IFeatureManager _featureManager;
-        private readonly IDictionary<string, bool> _flagCache = new Dictionary<string, bool>();
+        private readonly ConcurrentDictionary<string, Lazy<Task<bool>>> _flagCache = new ConcurrentDictionary<string, Lazy<Task<bool>>>();
         private IEnumerable<string> _featureNames;
 
         public FeatureManagerSnapshot(IFeatureManager featureManager)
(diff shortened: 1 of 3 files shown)


src/Microsoft.FeatureManagement/FeatureManagerSnapshot.cs
```diff
@@ -43,34 +45,26 @@ await foreach (string featureName in _featureManager.GetFeatureNamesAsync().Conf
...
-            _flagCache[feature] = enabled;
+            Lazy<Task<bool>> evaluator = _flagCache.GetOrAdd(
+                feature,
+                (key) => 
+                    new Lazy<Task<bool>>(
+                        () => _featureManager.IsEnabledAsync(key),
+                        LazyThreadSafetyMode.ExecutionAndPublication));
 
-            return enabled;
+            return await evaluator.Value.ConfigureAwait(false);
```
gfoidl, line 55: You don't need to await the task here, just return it thus saving the async state machinery (allocations).

I'm not sure if the `Lazy` is needed. It's safer for race conditions in the sense of flag evaluations, but caching is a best effort solution which should be fast, so I would trade perf over possible "double evaluations".

jimmyca15: @gfoidl thanks that's a good improvement with the await removal.

I'm going to remove the double evaluation protection in this PR and think about that problem a bit more.

avanigupta (approved):

src/Microsoft.FeatureManagement/FeatureManagerSnapshot.cs
```diff
@@ -41,36 +42,18 @@ await foreach (string featureName in _featureManager.GetFeatureNamesAsync().Conf
...
-            }
-
-            bool enabled = await _featureManager.IsEnabledAsync(feature).ConfigureAwait(false);
-
-            _flagCache[feature] = enabled;
-
-            return enabled;
+            return _flagCache.GetOrAdd(
+                feature,
+                (key) => _featureManager.IsEnabledAsync(key));
```
zhenlan, line 49: Should we have `.ConfigureAwait(false)` here?

jimmyca15: That's only needed where the task is awaited. So the consumer of this should `ConfigureAwait` if they are awaiting the result.

zhenlan: Makes sense. We can't save the `ConfigureAwait` from the user code in this case.

zhenlan (approved):

jimmyca15:
Had to rebase to resolve a merge conflict.

Issue #145, opened 2021-09-19
FilterAliasAttribute throws ArgumentNullException with the value of alias param as the paramName rather than the nameof the alias param


brental:
The FilterAliasAttribute constructor whether the alias parameter IsNullOrEmpty and if it is, throws an ArgumentNullException. However, it is throwing the ArgumentNullException with the value of the alias parameter (which would always be null or empty) rather than the nameof the alias parameter as the ArgumentNullException expects. So, it seems like the constructor should be this:

``` C#
public FilterAliasAttribute(string alias)
{
    if (string.IsNullOrEmpty(alias))
    {
        throw new ArgumentNullException(nameof(alias));
    }

    Alias = alias;
}
```

Rather than this:

``` C#
public FilterAliasAttribute(string alias)
{
    if (string.IsNullOrEmpty(alias))
    {
        throw new ArgumentNullException(alias);
    }

    Alias = alias;
}
```

brental:
Closing as a fix has been merged.

PR #146, opened 2021-09-19
Updated FilterAliasAttribute to use nameof(alias) when throwing an ArgumentNullException when the alias param is null or empty.

brental:
Updated FilterAliasAttribute to use nameof(alias) when throwing an ArgumentNullException when the alias param is null or empty. Fix for issue #145.


jimmyca15:
@brental thanks!

jimmyca15 (approved):

PR #148, opened 2021-09-22
Updated version to 2.4.0.

jimmyca15:
Updated version to 2.4.0.

### DIFF
--- a/src/Microsoft.FeatureManagement.AspNetCore/Microsoft.FeatureManagement.AspNetCore.csproj
+++ b/src/Microsoft.FeatureManagement.AspNetCore/Microsoft.FeatureManagement.AspNetCore.csproj
@@ -4,7 +4,7 @@
   <!-- Official Version -->
   <PropertyGroup>
     <MajorVersion>2</MajorVersion>
-    <MinorVersion>3</MinorVersion>
+    <MinorVersion>4</MinorVersion>
     <PatchVersion>0</PatchVersion>
   </PropertyGroup>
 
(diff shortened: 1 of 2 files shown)


src/Microsoft.FeatureManagement.AspNetCore/Microsoft.FeatureManagement.AspNetCore.csproj
```diff
@@ -4,7 +4,7 @@
   <!-- Official Version -->
   <PropertyGroup>
     <MajorVersion>2</MajorVersion>
-    <MinorVersion>3</MinorVersion>
+    <MinorVersion>4</MinorVersion>
```
avanigupta, line 7: We should only update the base package version and release that first. Once Microsoft.FeatureManagement v2.4.0 is on NuGet, we can bump up the version of AspNetCore project and also update the base package reference at the same time.

jimmyca15: Why is that?

jimmyca15: The AspNetCore project does not reference the base project by package version.
https://github.com/microsoft/FeatureManagement-Dotnet/blob/423501a59d8d086bbe60baa89dedcded1133213f/src/Microsoft.FeatureManagement.AspNetCore/Microsoft.FeatureManagement.AspNetCore.csproj#L33

avanigupta: Oh I see, in that case we don't need separate version upgrades because AspNetCore project has a reference to the base project directly. 
For DotNetProvider, we need two different version upgrades because AspNetCore project references the latest base package from NuGet.

avanigupta (approved):

PR #155, opened 2021-11-25
Add support for the ':' character in feature flags when using the default feature flag definition provider.

jimmyca15:
This PR adds support for feature flags containing the ':' character.

Currently a feature with the name "Feature:With:Colons" cannot be found by the feature manager. With this change, the feature will now be found and evaluated as expected.

### DIFF
--- a/src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
+++ b/src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
@@ -18,6 +18,7 @@ namespace Microsoft.FeatureManagement
     sealed class ConfigurationFeatureDefinitionProvider : IFeatureDefinitionProvider, IDisposable
     {
         private const string FeatureFiltersSectionName = "EnabledFor";
+        private static readonly string[] SectionSeparator = new string[] { ConfigurationPath.KeyDelimiter };
         private readonly IConfiguration _configuration;
         private readonly ConcurrentDictionary<string, FeatureDefinition> _definitions;
         private IDisposable _changeSubscription;
@@ -83,8 +84,39 @@ namespace Microsoft.FeatureManagement
 
         private FeatureDefinition ReadFeatureDefinition(string featureName)
         {
-            IConfigurationSection configuration = GetFeatureDefinitionSections()
-                                                    .FirstOrDefault(section => section.Key.Equals(featureName, StringComparison.OrdinalIgnoreCase));
+            IConfigurationSection configuration;
+
+            if (!featureName.Contains(ConfigurationPath.KeyDelimiter))
+            {
+                configuration = GetFeatureDefinitionSections()
+                    .FirstOrDefault(section => section.Key.Equals(featureName, StringComparison.OrdinalIgnoreCase));
+            }
+            else
+            {
+                //
+                // Feature names with configuration path delimiters require traversing children sections for resolution
+                IEnumerable<string> sectionNames = featureName.Split(SectionSeparator, StringSplitOptions.None);
+
+                IEnumerable<IConfigurationSection> sections;
+
+                IConfigurationSection section = null;
+
+                foreach (string sectionName in sectionNames)
+                {
+                    sections = section == null ?
+                        GetFeatureDefinitionSections() :
+                        section.GetChildren();
+
+                    section = sections.FirstOrDefault(section => section.Key.Equals(sectionName, StringComparison.OrdinalIgnoreCase));
+
+                    if (section == null)
+                    {
+                        break;
+                    }
+                }
+
+                configuration = section;
+            }
 
             if (configuration == null)
             {
--- a/tests/Tests.FeatureManagement/FeatureManagement.cs
+++ b/tests/Tests.FeatureManagement/FeatureManagement.cs
@@ -26,6 +26,7 @@ namespace Tests.FeatureManagement
         private const string OffFeature = "OffFeature";
         private const string ConditionalFeature = "ConditionalFeature";
         private const string ContextualFeature = "ContextualFeature";
+        private const string FeatureWithColons = "Feature:With:Colons";
 
         [Fact]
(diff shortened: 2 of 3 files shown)


jimmyca15:
cc @MaryanneNjeri I wasn't able to add you as a reviewer.

src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
```diff
@@ -83,8 +84,39 @@ public async IAsyncEnumerable<FeatureDefinition> GetAllFeatureDefinitionsAsync()
...
+
+                    section = sections.FirstOrDefault(section => section.Key.Equals(sectionName, StringComparison.OrdinalIgnoreCase));
+
+                    if (section == null)
+                    {
+                        break;
+                    }
+                }
+
+                configuration = section;
```
avanigupta, line 118: If `featureName` is `Feature:With:Colons`, after the foreach loop ends, we'll set `configuration` to the section named `Colons` . Then we call `ReadFeatureDefinition(configuration)`, and that method will return the `FeatureDefinition` object where `FeatureDefinition.Name` will be set to `Colons` instead of `Feature:With:Colons` ([code](https://github.com/microsoft/FeatureManagement-Dotnet/blob/8d1774ff0521dbbf0f1f35551cc2d7115d1f65e9/src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs#L172)).

src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
```diff
@@ -83,8 +84,39 @@ public async IAsyncEnumerable<FeatureDefinition> GetAllFeatureDefinitionsAsync()
...
+
+                IEnumerable<IConfigurationSection> sections;
+
+                IConfigurationSection section = null;
+
+                foreach (string sectionName in sectionNames)
+                {
+                    sections = section == null ?
+                        GetFeatureDefinitionSections() :
+                        section.GetChildren();
```
avanigupta, line 108: `section == null` will only be true for the first iteration in the foreach loop. Instead of doing this check inside the loop every time, can we initialize `sections` to `GetFeatureDefinitionSections()` before the loop?

src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
```diff
@@ -83,8 +84,39 @@ public async IAsyncEnumerable<FeatureDefinition> GetAllFeatureDefinitionsAsync()
...
+            if (!featureName.Contains(ConfigurationPath.KeyDelimiter))
+            {
+                configuration = GetFeatureDefinitionSections()
+                    .FirstOrDefault(section => section.Key.Equals(featureName, StringComparison.OrdinalIgnoreCase));
+            }
+            else
+            {
+                //
+                // Feature names with configuration path delimiters require traversing children sections for resolution
+                IEnumerable<string> sectionNames = featureName.Split(SectionSeparator, StringSplitOptions.None);
```
avanigupta, line 98: Could we define `SectionSeparator` as a `string` instead of `string[]`? Another option is to directly use `ConfigurationPath.KeyDelimiter` because this is a special delimiter that's unlikely to change in .NET.

jimmyca15: I just stored it that way since `Split` accepts string[] and `ConfigurationPath.KeyDelimiter` is a string instead of a char.

src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
jimmyca15: Thanks. This made me realize a bigger problem which is how this works with `GetFeatureDefinitionAsync`. We may not be able to handle that

jimmyca15: GetFeatureDefinitionAsync can't detect features that have `:`. That's a glaring hole in supporting `:`. We may just want to block `:` because of that.

src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
avanigupta: FWIW, [Split ](https://docs.microsoft.com/en-us/dotnet/api/system.string.split?view=net-6.0#System_String_Split_System_String_System_StringSplitOptions_)also accepts string delimiter.

jimmyca15:
@fpaganetto We abandoned this due to the complexity of supporting the ':' character in .NET Core. This character separates configuration sections and as you mentioned breaks the way the feature manager works today. It's not possible to deterministically enumerate all defined features if the `:` character is allowed. Without redesign at least.

fpaganetto:
Just to be clear, I want to have subsections on the ```appsetings.json``` with nested objects and access using the `:` on the FeatureGate annotation constant. Is that not going to be possible?

I don't understand if this is because of allowing the `:` character on the key name. That contradicts the PR https://github.com/microsoft/FeatureManagement-Dotnet/pull/95 where the usage of `:` is being rejected.

An example of what I want:
```csharp
[FeatureGate("Authentication:User:Read")]
public async Task<IActionResult> ReadUser(int id){}

[FeatureGate("Authentication:User:Create")]
public async Task<IActionResult> CreateUser(User user){}
```

On my appsettings.json:
```json
"EndpointEnabledStatus": {
    "Authentication": {
      "User": {
        "Create":  false,
        "Read" : true,
      }
    }
  }
```

jimmyca15:
@fpaganetto Thanks for the clarification. That is not going to be possible. The PR that you linked does need to be completed to make it clear that the ':' character is not supported. I will bump it for review.

PR #156, opened 2021-12-03
Fix PercentageFilter inequality bound

neiljohari:
NextDouble() ∈ [0, 1). We use a strict inequality, so a 0% rollout value sporadically returns true as 0⩽0.

The fix is to use a non-strict inequality, <. The upper bound still works: 0.999... < 1.

We noticed this internally because feature flags were occasionally tripping even with 0% rollout. This appears to be the root cause.

### DIFF
--- a/src/Microsoft.FeatureManagement/FeatureFilters/PercentageFilter.cs
+++ b/src/Microsoft.FeatureManagement/FeatureFilters/PercentageFilter.cs
@@ -46,7 +46,7 @@ namespace Microsoft.FeatureManagement.FeatureFilters
 
             if (result)
             {
-                result = (RandomGenerator.NextDouble() * 100) <= settings.Value;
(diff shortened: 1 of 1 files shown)


neiljohari:
Additionally, the CLA GHA didn't appear to trip. This PR is provided as part of my employment with https://github.com/roblox, I have the needed authorization to sign the CLA so please do send me a link to an agreement I can sign!

jimmyca15:
@neiljohari The license check mentions that all CLA requirements are met. Have you already contributed to other Microsoft repos?

neiljohari:
@jimmyca15 Thanks for the super speedy response! I don't believe I have previously contributed, I think this is my first one

jimmyca15 (approved):

jimmyca15:
I took a look. The license check did run. For small enough changes it doesn't prompt CLA sign.

Issue #87, opened 2020-07-21
IFeatureManagerSnapshot one flag, many contexts. 


PiotrJustyna:
Probably it's more a question than an issue (I hope), but is this expected behavior?

**Context:**

I'm evaluating one flag for two different contexts:

* `123`
* `456`

The flag should only return `true` for context `123`.

```csharp
var featureVFor123 = await _featureManagerSnapshot.IsEnabledAsync("FeatureV", "123");
var featureVFor456 = await _featureManagerSnapshot.IsEnabledAsync("FeatureV", "456");

return $"FeatureVFor123: {await featureVFor123}. FeatureVFor456: {await featureVFor456}.";
```

(where `_featureManagerSnapshot` is of type `IFeatureManagerSnapshot`)

returns:

`FeatureVFor123: true. FeatureVFor456: true.`

Which I would not expect (it's a different set of conditions for the flag). I would expect:

`FeatureVFor123: true. FeatureVFor456: false.`

It looks like the evaluated flag is getting cached and subsequent evaluations are ignoring different contexts. It does work correctly when I use `IFeatureManager _featureManager`.

Is this by design?

leonids2005:
Looks like this is by design :(.  Not perfect imo.

Here is the code 

        public async Task<bool> IsEnabledAsync<TContext>(string feature, TContext context)
        {
            //
            // First, check local cache
            if (_flagCache.ContainsKey(feature))
            {
                return _flagCache[feature];
            }

            bool enabled = await _featureManager.IsEnabledAsync(feature, context).ConfigureAwait(false);

            _flagCache[feature] = enabled;

            return enabled;
        }

jimmyca15:
@PiotrJustyna 

The IFeatureManagerSnapshot is purposefully naïve when it comes to feature contexts. We would need a way to check for equality between contexts to be able to do anything further, and this would've required every context to implement an interface that we provide. This is not always possible so we decided not to impose this limitation. Our thoughts were that in general, if IFeatureManagerSnapshot would be useful in combination with feature contexts, then the lifetime of the feature manager snapshot would be in line with the lifetime of the feature context.

jimmyca15:
Another option we have is to allow Object.Equals to play the role of establishing context equality. This may solve the problem.

[...]

PR #157, opened 2021-12-14
Update to dynamic features.

jimmyca15:
This PR contains multiple changes. The two biggest changes are the separation of feature flags and dynamic features and the renaming of types to respect dynamic feature terminology. Most other changes follow as consequence. Details of the included changes are below.

Documentation updates, including the new configuration schema, to follow in a subsequent PR.

* Update types to respect dynamic terminology. E.g. IFeatureManager -> IDynamicFeatureManager.
  * Updated property/method summaries to distinguish dynamic features and feature flags.
* Separate feature flags and dynamic features in configuration schema
  * Backward compatability support for old configuration schema.
* Breaking: Update custom feature definition provider to return feature flags and dynamic features.
  * Added DynamicFeatureDefinition class
  * renamed FeatureDefinition to FeatureFlagDefinition.
* Breaking: IFeatureManager.GetFeatureNamesAsync renamed to IFeatureManager.GetFeatureFlagNamesAsync
* Corrected namespace for built-in feature variant assigners.
  * Microsoft.FeatureManagement -> Microsoft.FeatureManagement.Assigners
* Added IDynamicFeatureManagerSnapshot
* Added test for custom feature definition provider.
* Added test for backcompat support for v1 configuration schema.

### Old configuration schema example

``` Javascript
{
    "FeatureManagement": {
        "FeatureFlag1": true,
        "FeatureFlag2": {
            "EnabledFor": [
                ...
            ]
        }
    }
}
```

### New configuration schema example

``` Javascript
{
    "FeatureManagement": {
        "FeatureFlags": {
            "FeatureFlag1": true,
            "FeatureFlag2": {
                "EnabledFor": [
                    ...
                ]
            }
        },
        "DynamicFeatures": {
            "DynamicFeature1": {
                "Assigner": "Targeting",
                "ConfigurationReference": "SomeSettings",
                "AssignmentParameters": {
                    ...
                }
            }
        }
    }
}
```


src/Microsoft.FeatureManagement/IDynamicFeatureManager.cs
```diff
@@ -0,0 +1,42 @@
...
+        IAsyncEnumerable<string> GetDynamicFeatureNamesAsync(CancellationToken cancellationToken = default);
+
+        /// <summary>
+        /// Retrieves a typed representation of the feature variant that should be used for a given dynamic feature.
+        /// </summary>
+        /// <typeparam name="T">The type that the feature variant's configuration should be bound to.</typeparam>
+        /// <param name="feature">The name of the dynamic feature.</param>
+        /// <param name="cancellationToken">The cancellation token to cancel the operation.</param>
+        /// <returns>A typed representation of the feature variant that should be used for a given dynamic feature.</returns>
+        ValueTask<T> GetVariantAsync<T>(string feature, CancellationToken cancellationToken);
```
zhenlan, line 29: Add default for cancellationToken

src/Microsoft.FeatureManagement/IDynamicFeatureManager.cs
```diff
@@ -0,0 +1,42 @@
...
+        /// <summary>
+        /// Retrieves a typed representation of the feature variant that should be used for a given dynamic feature.
+        /// </summary>
+        /// <typeparam name="T">The type that the feature variant's configuration should be bound to.</typeparam>
+        /// <typeparam name="TContext">The type of the context being provided to the dynamic feature manager for use during the process of choosing which variant to use.</typeparam>
+        /// <param name="feature">The name of the dynamic feature.</param>
+        /// <param name="context">A context providing information that can be used to evaluate which variant should be used for the dynamic feature.</param>
+        /// <param name="cancellationToken">The cancellation token to cancel the operation.</param>
+        /// <returns>A typed representation of the feature variant's configuration that should be used for a given feature.</returns>
+        ValueTask<T> GetVariantAsync<T, TContext>(string feature, TContext context, CancellationToken cancellationToken);
```
zhenlan, line 40: Add default for cancellationToken

examples/CustomAssignmentConsoleApp/Program.cs
```diff
@@ -35,7 +35,7 @@ public static async Task Main(string[] args)
             // Get the feature manager from application services
             using (ServiceProvider serviceProvider = services.BuildServiceProvider())
             {
-                IFeatureVariantManager variantManager = serviceProvider.GetRequiredService<IFeatureVariantManager>();
+                IDynamicFeatureManager variantManager = serviceProvider.GetRequiredService<IDynamicFeatureManager>();
```
zhenlan, line 38: >variantManager [](http://example.com/codeflow?start=39&length=14)

NIT: update variable name variantManager

examples/FeatureFlagDemo/Views/Shared/_Layout.cshtml
```diff
@@ -1,5 +1,5 @@
 ﻿@using Microsoft.FeatureManagement
-@inject IFeatureVariantManager variantManager;
+@inject IDynamicFeatureManager variantManager;
```
zhenlan, line 2: NIT: update variable name variantManager too?

examples/TargetingConsoleApp/Program.cs
```diff
@@ -40,7 +41,7 @@ public static async Task Main(string[] args)
             using (ServiceProvider serviceProvider = services.BuildServiceProvider())
             {
                 IFeatureManager featureManager = serviceProvider.GetRequiredService<IFeatureManager>();
-                IFeatureVariantManager variantManager = serviceProvider.GetRequiredService<IFeatureVariantManager>();
+                IDynamicFeatureManager variantManager = serviceProvider.GetRequiredService<IDynamicFeatureManager>();
```
zhenlan, line 44: NIT: update variable name variantManager to dynamicFeatureManager?

src/Microsoft.FeatureManagement/FeatureManagerSnapshot.cs
```diff
@@ -12,37 +12,99 @@ namespace Microsoft.FeatureManagement
...
             }
         }
 
+        public async IAsyncEnumerable<string> GetDynamicFeatureNamesAsync([EnumeratorCancellation]CancellationToken cancellationToken = default)
+        {
+            if (_dynamicFeatureNames == null)
+            {
+                var dynamicFeatureNames = new List<string>();
+
+                await foreach (string featureName in _featureManager.GetFeatureFlagNamesAsync(cancellationToken).ConfigureAwait(false))
```
zhenlan, line 58: It should be 
`_dynamicFeatureManager.GetDynamicFeatureNamesAsync`

src/Microsoft.FeatureManagement/FeatureManagerSnapshot.cs
```diff
@@ -12,37 +12,99 @@ namespace Microsoft.FeatureManagement
...
+            T variant = await _dynamicFeatureManager.GetVariantAsync<T>(feature, cancellationToken).ConfigureAwait(false);
+
+            _variantCache[cacheKey] = variant;
+
+            return variant;
+        }
+
+        public async ValueTask<T> GetVariantAsync<T, TContext>(string feature, TContext context, CancellationToken cancellationToken)
+        {
+            string cacheKey = GetVariantCacheKey<T>(feature);
```
zhenlan, line 92: Should the context be part of the cache key in this case too?

src/Microsoft.FeatureManagement/FeatureManager.cs
```diff
@@ -17,7 +16,7 @@ namespace Microsoft.FeatureManagement
     /// <summary>
     /// Used to evaluate whether a feature is enabled or disabled.
     /// </summary>
-    class FeatureManager : IFeatureManager, IFeatureVariantManager
+    class FeatureManager : IFeatureManager, IDynamicFeatureManager
```
zhenlan, line 19: I'm thinking whether we should split the implementation of two interfaces into separate classes. They are pretty much independent. This could be confusing for someone new to look at this code in the future. Thoughts?

src/Microsoft.FeatureManagement/IContextualFeatureVariantAssigner.cs
```diff
@@ -7,18 +7,18 @@
...
-        /// <param name="variantAssignmentContext">A variant assignment context that contains information needed to assign a variant for a feature.</param>
-        /// <param name="appContext">A context defined by the application that is passed in to the feature management system to provide contextual information for assigning a variant of a feature.</param>
+        /// <param name="variantAssignmentContext">A variant assignment context that contains information needed to assign a variant for a dynamic feature.</param>
+        /// <param name="appContext">A context defined by the application that is passed in to the feature management system to provide contextual information for assigning a variant of a dynamic feature.</param>
         /// <param name="cancellationToken">The cancellation token to cancel the operation.</param>
-        /// <returns>The variant that should be assigned for a given feature.</returns>
+        /// <returns>The variant that should be assigned for a given dynamic feature.</returns>
         ValueTask<FeatureVariant> AssignVariantAsync(FeatureVariantAssignmentContext variantAssignmentContext, TContext appContext, CancellationToken cancellationToken);
```
zhenlan, line 22: >cancellationToken [](http://example.com/codeflow?start=150&length=17)

Should we default the cancellationToken?

Looks like we don't have cancellationToken in I*FeatureFilter, should we add it?

src/Microsoft.FeatureManagement/IFeatureDefinitionProvider.cs
```diff
@@ -13,18 +13,33 @@ namespace Microsoft.FeatureManagement
...
-        /// Retrieves the definition for a given feature.
+        /// Retrieves the definition for a given feature flag.
         /// </summary>
-        /// <param name="featureName">The name of the feature to retrieve the definition for.</param>
+        /// <param name="featureName">The name of the feature flag to retrieve the definition for.</param>
         /// <param name="cancellationToken">The cancellation token to cancel the operation.</param>
-        /// <returns>The feature's definition.</returns>	
-        Task<FeatureDefinition> GetFeatureDefinitionAsync(string featureName, CancellationToken cancellationToken = default);
+        /// <returns>The feature flag's definition.</returns>	
+        Task<FeatureFlagDefinition> GetFeatureFlagDefinitionAsync(string featureName, CancellationToken cancellationToken = default);
```
zhenlan, line 21: >featureName [](http://example.com/codeflow?start=73&length=11)

NIT: featureName -> featureFlagName

src/Microsoft.FeatureManagement/IFeatureDefinitionProvider.cs
```diff
@@ -13,18 +13,33 @@ namespace Microsoft.FeatureManagement
     public interface IFeatureDefinitionProvider
```
zhenlan, line 13: Should we split it into separate interfaces for feature flags and dynamic features so users don't have to implement both if they only care about one but not the other?

src/Microsoft.FeatureManagement/IFeatureManager.cs
```diff
@@ -8,32 +8,32 @@
...
         /// <summary>
-        /// Checks whether a given feature is enabled.
+        /// Checks whether a given feature flag is enabled.
         /// </summary>
-        /// <param name="feature">The name of the feature to check.</param>
+        /// <param name="feature">The name of the feature flag to check.</param>
         /// <param name="cancellationToken">The cancellation token to cancel the operation.</param>
-        /// <returns>True if the feature is enabled, otherwise false.</returns>
+        /// <returns>True if the feature flag is enabled, otherwise false.</returns>
         Task<bool> IsEnabledAsync(string feature, CancellationToken cancellationToken = default);
```
zhenlan, line 28: feature -> featureFlag?

src/Microsoft.FeatureManagement/FeatureManagerSnapshot.cs
```diff
@@ -12,37 +12,99 @@ namespace Microsoft.FeatureManagement
...
             }
 
-            foreach (string featureName in _featureNames)
+            foreach (string featureName in _featureFlagNames)
             {
                 yield return featureName;
             }
         }
 
+        public async IAsyncEnumerable<string> GetDynamicFeatureNamesAsync([EnumeratorCancellation]CancellationToken cancellationToken = default)
```
avanigupta, line 52: >cancellationToken = default [](http://example.com/codeflow?start=116&length=27)

default not needed here. Cancellation token already has a default value in the interface.

src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
```diff
@@ -127,11 +175,9 @@ We support
 
             */
 
-            var enabledFor = new List<FeatureFilterConfiguration>();
+            Debug.Assert(configurationSection != null);
```
avanigupta, line 178: >Debug.Assert(configurationSection != null); [](http://example.com/codeflow?start=12&length=43)

Should we throw an exception instead of assert?

src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
```diff
@@ -171,48 +217,78 @@ We support
...
+            return new FeatureFlagDefinition()
+            {
+                Name = configurationSection.Key,
+                EnabledFor = enabledFor,
+            };
+        }
+
+        private DynamicFeatureDefinition ReadDynamicFeatureDefinition(IConfigurationSection configurationSection)
+        {
+            Debug.Assert(configurationSection != null);
```
avanigupta, line 231: >Debug.Assert(configurationSection != null); [](http://example.com/codeflow?start=12&length=43)

Same here. Null check may not even be necessary since the calling methods already check for null before invoking these two methods.

src/Microsoft.FeatureManagement/FeatureManagerSnapshot.cs
```diff
@@ -12,37 +12,99 @@ namespace Microsoft.FeatureManagement
...
+            string cacheKey = GetVariantCacheKey<T>(feature);
+
+            //
+            // First, check local cache
+            if (_flagCache.ContainsKey(feature))
+            {
+                return (T)_variantCache[cacheKey];
+            }
+
+            T variant = await _dynamicFeatureManager.GetVariantAsync<T>(feature, cancellationToken).ConfigureAwait(false);
```
jimmyca15, line 101: Should pass context.

src/Microsoft.FeatureManagement/FeatureManagerSnapshot.cs
jimmyca15: We didn't factor context into the snapshot for feature flags. There was a discussion for this in this issue: https://github.com/microsoft/FeatureManagement-Dotnet/issues/87

src/Microsoft.FeatureManagement/IFeatureDefinitionProvider.cs
jimmyca15: We split IFeatureManager and IDynamicFeatureManager so I'm not opposed to carrying that over here.

src/Microsoft.FeatureManagement/FeatureManager.cs
jimmyca15: I don't see the value immediately. What is the confusion that you had in mind?

src/Microsoft.FeatureManagement/FeatureManagerSnapshot.cs
zhenlan: Indeed. That carries some complexity. Let's see if it's brought up again in the future. Leave it as-is for now.

src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
jimmyca15: It's a private method so throwing here wouldn't be expected.

src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
jimmyca15: We often `Debug.Assert` use `Debug.Assert` in our private methods. It offers a couple benefits.

1. It makes it clear that the assertion can be relied on in the rest of the method execution. In this case, the reader will know the `configurationSection` is not null. 
2. When we look at this method we know that if we're calling it from another method we shouldn't pass null, and vice versa we can look at the places that call this and confirm that they're doing a null check, like you mentioned, before calling. 
3. If it is ever null for some reason, it will throw in debug mode so it can be fixed.

src/Microsoft.FeatureManagement/FeatureManager.cs
zhenlan: I guess the question can be asked in the other way - why do we implement both interfaces in one class? Would we do the same if we started from scratch? The functionalities of the two interfaces are parallel. It's not like one is a complement to another. In current naming, feels like the FeatureManager is really for IFeatureManager, and the IDynamicFeatureManager is a supplement (like IDisposable).

jimmyca15: > Would we do the same if we started from scratch?

I think that's a great way to look at it. I'd say we'd have them separated. Let me separate them.

jimmyca15: > In current naming, feels like the FeatureManager is really for IFeatureManager, and the IDynamicFeatureManager

For this point, it would be different if we renamed `IFeatureManager` to `IFeatureFlagManager`. But that was a breaking change that we didn't consider.

src/Microsoft.FeatureManagement/IContextualFeatureVariantAssigner.cs
jimmyca15: > Looks like we don't have cancellationToken in I*FeatureFilter, should we add it?

cancellation tokens on existing interfaces were added in PRs checked in to the main branch.
* https://github.com/microsoft/FeatureManagement-Dotnet/pull/131

For that reason I didn't touch them here. This PR is going into "feature/v3" branch and when "feature/v3" merges into main those conflicts will be resolved.

src/Microsoft.FeatureManagement/FeatureManager.cs
zhenlan: > For this point, it would be different if we renamed IFeatureManager to IFeatureFlagManager. But that was a breaking change that we didn't consider.

Totally, that would make more sense. Another option is to add what's in IDynamicFeatureManager to IFeatureManager directly. I think that's what we would do if we started from scratch, but we worried it will break all existing libraries built on top of the current feature management library.

jimmyca15: Yes, to be honest I'm not completely sold on either side. If you think it'll be more clear I'm happy to separate them. Otherwise, can keep in in `FeatureManager`

zhenlan: I think we should separate the implementation if we decide to separate the interface.

jimmyca15: In that case I'll go ahead and separate ConfigurationFeatureDefinitionProvider and FeatureManagerSnapshot as well.

src/Microsoft.FeatureManagement/IContextualFeatureVariantAssigner.cs
jimmyca15: Circling back to this, there were some missing default cancellation tokens. Thanks. I've added them.

src/Microsoft.FeatureManagement/FeatureManager.cs
jimmyca15: Updated.

src/Microsoft.FeatureManagement/DynamicFeatureManager.cs
```diff
@@ -0,0 +1,253 @@
...
+                    }
+
+                    return matchingAssigners.FirstOrDefault();
+                }
+            );
+
+            return assigner;
+        }
+
+        private static bool IsMatchingMetadataName(string metadataName, string desiredName, string suffix)
```
zhenlan, line 200: The comments in this function are still talking about filters. Is there a place we can put this helper function to so it can be shared?

jimmyca15: Sure. I put it in a new helper, `NameHelper`.

zhenlan:
> using System;

Missing copyright header

---
Refers to: src/Microsoft.FeatureManagement/NameHelper.cs:1 in ba664ee. [](commit_id = ba664ee550b4d1d1068eba7713a159bad3516879, deletion_comment = False)

zhenlan (approved):

Issue #34, opened 2019-12-11
Add a result filter for razor pages


haacked:
Right now, there's no nice declarative way (that I could find) to add a feature flag to a razor page.

```cs
[FeatureGate(FeatureFlags.FeatureA)]
public class IndexModel : PageModel {...}
```

Doesn't work because `FeatureGateAttribute` only overrides `OnActionExecutionAsync`. I haven't tested this myself, but I believe if you also implement `OnResultExecutingAsync` you'll cover Razor pages. At least according to the docs [Filter methods for Razor Pages in ASP.NET Core](https://docs.microsoft.com/en-us/aspnet/core/razor-pages/filter?view=aspnetcore-3.1#:~:targetText=Razor%20Page%20filters%20are%20similar,after%20model%20binding%20is%20complete.)

haacked:
Just as a follow-up, I hacked up a proof of concept to show this would work.

```cs
[AttributeUsage(AttributeTargets.Class, AllowMultiple = true)]
public class FeatureToggleAttribute : FeatureGateAttribute
{
    public FeatureToggleAttribute(params string[] features)
        : this(RequirementType.All, features)
    {
    }

    public FeatureToggleAttribute(RequirementType requirementType, params string[] features)
        : base(requirementType, features)
    {
    }

    public FeatureToggleAttribute(RequirementType requirementType, params object[] features)
        : base(requirementType, features)
    {
    }
    
    public FeatureToggleAttribute(params object[] features)
        : this(RequirementType.All, features)
    {
    }

    public override async Task OnResultExecutionAsync(ResultExecutingContext context, ResultExecutionDelegate next)
    {
        var fm = context.HttpContext.RequestServices.GetRequiredService<IFeatureManagerSnapshot>();
        bool flag;
        if (RequirementType == RequirementType.All)
            flag = await Features.All<string>((Func<string, Task<bool>>) (async feature =>
                await fm.IsEnabledAsync(feature).ConfigureAwait(false)));
        else
            flag = await Features.Any<string>((Func<string, Task<bool>>) (async feature =>
                await fm.IsEnabledAsync(feature).ConfigureAwait(false)));
        if (flag)
        {
            await next().ConfigureAwait(false);
        }
        else
        {
            context.HttpContext.Response.StatusCode = 404;
            await context.HttpContext.Response.CompleteAsync();
        }
    }
}
```

This code makes use of internal classes within this repo that I didn't copy. But it shows that the basic idea would work.

[...]

PR #165, opened 2022-03-23
Added PageFeatureGateAttribute. This enables razor pages to be gated by features.

jimmyca15:
See https://github.com/microsoft/FeatureManagement-Dotnet/issues/34 for discussion leading to this PR.

Currently the library supports controlling visibility of controllers with the `FeatureGateAttribute`. This works with MVC controllers and actions. It does not work with Razor pages due to differences in their execution pipeline. The discussion in the linked issue details how filtering can be performed for MVC Razor pages. Because of the differences in the execution pipeline for MVC controllers and Razor pages, the best way to support both scenarios is to have two separate attributes. This is why a new attribute is being added.

## Usage
```
namespace RazorPages.Pages
{
    //
    // The content of this Razor page will only be served
    // if the "Home" feature is enabled.
    [PageFeatureGate("Home")]
    public class IndexModel : PageModel
    {
        public void OnGet()
        {
        }
    }
}

```


jimmyca15:
Closed. There is no reason to create a new attribute for razor pages. action filter pipeline and page filter pipeline is separated. They can exist on the same attribute and not interfere with one another.

PR #166, opened 2022-03-23
Updated FeatureGateAttribute to be able to be used on Razor Pages.

jimmyca15:
See https://github.com/microsoft/FeatureManagement-Dotnet/issues/34 for discussion leading to this PR.

Currently the library supports controlling visibility of controllers with the `FeatureGateAttribute`. This works with MVC controllers and actions. It does not work with Razor pages. This is because the `FeatureGateAttribute` only runs as an MVC action filter. Code execution in the Razor Page pipeline is performed via page handlers.

By updating the `FeatureGateAttribute` to implement `IAsyncPageFilter` we are capable of supporting Razor Pages. MVC controllers only execute the action filter portion of `FeatureGateAttribute`, while Razor Pages only execute the page filter portion. 

## Usage
```
namespace RazorPages.Pages
{
    //
    // The content of this Razor page will only be served
    // if the "Home" feature is enabled.
    [FeatureGate("Home")]
    public class IndexModel : PageModel
    {
        public void OnGet()
        {
        }
    }
}

```


tests/Tests.FeatureManagement/Tests.FeatureManagement.csproj
```diff
@@ -1,32 +1,14 @@
...
-  <ItemGroup>
-    <None Remove="appsettings.json" />
-  </ItemGroup>
-
-  <ItemGroup>
-    <Content Include="appsettings.json">
-      <CopyToOutputDirectory>Always</CopyToOutputDirectory>
-      <CopyToPublishDirectory>PreserveNewest</CopyToPublishDirectory>
-    </Content>
-  </ItemGroup>
```
pratiksanglikar, line 18: Curious to know why we don't need to override these options anymore.

tests/Tests.FeatureManagement/Tests.FeatureManagement.csproj
```diff
@@ -1,32 +1,14 @@
-﻿<Project Sdk="Microsoft.NET.Sdk">
+﻿<Project Sdk="Microsoft.NET.Sdk.Web">
 
   <PropertyGroup>
-    <TargetFrameworks>netcoreapp2.1;netcoreapp3.1;net5.0</TargetFrameworks>
+    <TargetFrameworks>netcoreapp3.1;net5.0</TargetFrameworks>
```
pratiksanglikar, line 4: While we are at it, should we also add .NET6 to the targets since it is the latest LTS release of .NET.

tests/Tests.FeatureManagement/Pages/RazorTestAny.cshtml.cs
```diff
@@ -0,0 +1,19 @@
...
+using Microsoft.FeatureManagement.Mvc;
+
+namespace Tests.FeatureManagement.Pages
+{
+    [FeatureGate(RequirementType.Any, Features.ConditionalFeature, Features.ConditionalFeature2)]
+    public class RazorTestAnyModel : PageModel
+    {
+        public IActionResult OnGet()
+        {
+            return StatusCode(200);
```
pratiksanglikar, line 16: > StatusCode(200)

Should we use `HttpStatusCode.OK` instead? [HttpStatusCode](https://docs.microsoft.com/en-us/dotnet/api/system.net.httpstatuscode?view=netcore-3.1)

tests/Tests.FeatureManagement/Pages/RazorTestAll.cshtml.cs
```diff
@@ -0,0 +1,18 @@
...
+using Microsoft.FeatureManagement.Mvc;
+
+namespace Tests.FeatureManagement.Pages
+{
+    [FeatureGate(Features.ConditionalFeature, Features.ConditionalFeature2)]
+    public class RazorTestAllModel : PageModel
+    {
+        public IActionResult OnGet()
+        {
+            return StatusCode(200);
```
pratiksanglikar, line 15: > StatusCode(200)

Should we use `HttpStatusCode.OK` instead? [HttpStatusCode](https://docs.microsoft.com/en-us/dotnet/api/system.net.httpstatuscode?view=netcore-3.1)

src/Microsoft.FeatureManagement.AspNetCore/Microsoft.FeatureManagement.AspNetCore.csproj
```diff
@@ -33,11 +33,12 @@
     <ProjectReference Include="..\Microsoft.FeatureManagement\Microsoft.FeatureManagement.csproj" />
   </ItemGroup>
 
-  <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.0'" >
-    <PackageReference Include="Microsoft.AspNetCore.Mvc.TagHelpers" Version="2.1.3" />
+  <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.0'">
+	  <PackageReference Include="Microsoft.AspNetCore.Mvc.RazorPages" Version="2.1.11" />
+	  <PackageReference Include="Microsoft.AspNetCore.Mvc.TagHelpers" Version="2.1.3" />
```
pratiksanglikar, line 38: Nit: indent looks little off.

tests/Tests.FeatureManagement/Tests.FeatureManagement.csproj
jimmyca15: In order to use Razor Pages in the test project I had to update the sdk type to `Microsoft.NET.Sdk.Web`. The default settings for that sdk type suffice.

tests/Tests.FeatureManagement/Tests.FeatureManagement.csproj
jimmyca15: Yeah, it's a good idea. but that would be better in a separate PR to add .NET 6 targeting.

src/Microsoft.FeatureManagement.AspNetCore/Microsoft.FeatureManagement.AspNetCore.csproj
jimmyca15: Thanks, somehow a tab snuck in there. Fixed.

tests/Tests.FeatureManagement/Pages/RazorTestAll.cshtml.cs
jimmyca15: Updated to `OkResult`. Now status code is avoided completely.

tests/Tests.FeatureManagement/Pages/RazorTestAny.cshtml.cs
jimmyca15: Updated to OkResult. Now status code is avoided completely.

pratiksanglikar (approved):
Looks good to me!

tests/Tests.FeatureManagement/Tests.FeatureManagement.csproj
```diff
@@ -1,32 +1,14 @@
-﻿<Project Sdk="Microsoft.NET.Sdk">
+﻿<Project Sdk="Microsoft.NET.Sdk.Web">
 
   <PropertyGroup>
-    <TargetFrameworks>netcoreapp2.1;netcoreapp3.1;net5.0</TargetFrameworks>
```
avanigupta, line 4: >netcoreapp2.1; [](http://example.com/codeflow?start=22&length=14)

Since we removed netcoreapp2.1 from target frameworks, are we deprecating netstandard2.0 support with this change?

src/Microsoft.FeatureManagement.AspNetCore/FeatureGateAttribute.cs
```diff
@@ -120,5 +121,38 @@ public override async Task OnActionExecutionAsync(ActionExecutingContext context
...
             }
         }
+
+        /// <summary>
+        /// Called asynchronously before the handler method is invoked, after model binding is complete.
+        /// </summary>
+        /// <param name="context">The <see cref="PageHandlerExecutingContext"/>.</param>
+        /// <param name="next">The <see cref="PageHandlerExecutionDelegate"/>. Invoked to execute the next page filter or the handler method itself.</param>
+        /// <returns>A <see cref="Task"/> that on completion indicates the filter has executed.</returns>
+        public virtual async Task OnPageHandlerExecutionAsync(PageHandlerExecutingContext context, PageHandlerExecutionDelegate next)
```
avanigupta, line 131: Do these need to be virtual methods?

tests/Tests.FeatureManagement/Tests.FeatureManagement.csproj
jimmyca15: That's a good question. netcoreapp2.1 is no longer supported which is why I thought to remove it since I was updating tests. Your comment made me realize that by removing the netcoreapp2.1 target framework we are no longer testing the netstandard2.0 binary.

Support for netstandard 2.0 going forward is up for discussion, but for now in this version of Microsoft.FeatureManagement (2.x) we still have it so I still want netcoreapp2.1 to be testing that binary. I added back the netcoreapp2.1 target framework to avoid convoluting this PR.

zhenlan (approved):
:shipit:

avanigupta (approved):

PR #174, opened 2022-05-02
Allow for both IFeatureFilter and IContextualFeatureFilter with same alias

### LINKED ISSUE #98: Unable to create both IFeatureFilter and IContextualFeatureFilter with same FilterAlias.
When I create an implementation for both `IFeatureFilter` and `IContextualFeatureFilter` with same `FilterAlias`. I get the below exception 

`Multiple feature filters match the configured filter named` 

**Use case**
I created two feature filter with same filter alias as follows

```csharp
[FilterAlias("CompanyFilter")]
public class HttpCompanyFilter : IFeatureFilter
{
	{
		// Return flag based on http context;
	}
}
```

```csharp
[FilterAlias("CompanyFilter")]
public class CompanyContextFilter : IContextualFeatureFilter<CompanyContext>
{
	public Task<bool> EvaluateAsync(FeatureFilterEvaluationContext featureFilterContext, CompanyContext companyContext)
	{
		// Return flag based on CompanyContext parameter;
	}
}
```

In some situations where I know that HTTP context will always be available, I want to call the feature manager without passing any context. This should call a filter which implements `IFeatureFilter`

```csharp
await featureFlagManager.IsEnabledAsync("FeatureA");
```

However, in some cases, I do not have the HTTP context available, so I would like to call the feature manager by passing the app context this time. This should call a filter which implements `IContextualFeatureFilter<CompanyContext>`

```csharp
var context = new CompanyContext()
{
 Code = "SomeValue"
};

await featureFlagManager.IsEnabledAsync("FeatureA", context);
```

I also tried to create a single filter implementing both `IFeatureFilter` and `IContextualFeatureFilter`but apparently this is also not allowed by the framework.

jviau:
This PR introduces an opt-in mechanic to allow for having `IFeatureFilter` and `IContextualFeatureFilter` with the same alias. Specifically, you can have:

0 or 1 `IFeatureFilter`
0 or N `IContextualFeatureFilter<T>`, so long as there is at most 1 applicable filter for `T`. So, you can have multiple contextual filters with the same alias, so long as they all handle discrete app contexts.

By default, this functionality is **disabled** and can be enabled via `FeatureManagementOptions.AllowDuplicateContextualAlias`.

TODO:
- [ ] Add unit tests

Closes #98

### DIFF
--- a/src/Microsoft.FeatureManagement/FeatureManagementOptions.cs
+++ b/src/Microsoft.FeatureManagement/FeatureManagementOptions.cs
@@ -21,5 +21,11 @@ namespace Microsoft.FeatureManagement
         /// The default value is true.
         /// </summary>
         public bool IgnoreMissingFeatures { get; set; } = true;
+
+        /// <summary>
+        /// Gets or sets a value indicating whether to allow for the same alias name to be used for a <see cref="IFeatureFilter" /> and a
+        /// <see cref="IContextualFeatureFilter{TContext}" />. Default is false.
+        /// </summary>
+        public bool AllowDuplicateContextualAlias { get; set; }
     }
 }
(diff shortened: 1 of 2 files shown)


zhiyuanliang-ms:
Move to this PR #262

PR #175, opened 2022-05-13
Implement a default/global feature definition

hermogenes:
This PR introduces an option to have global/default feature definition.

The benefit is allow use definitions that would apply for all features without need to set it individually.

For more information check feature request #89


Sn3b:
Would this allow an easy way to enable all features?

PR #176, opened 2022-05-13
Add Impressions Listener

joshsirota:


src/Microsoft.FeatureManagement/IFeatureImpressionsListener.cs
```diff
@@ -0,0 +1,13 @@
...
+// Licensed under the MIT license.
+//
+namespace Microsoft.FeatureManagement
+{
+    /// <summary>
+    /// Marker interface for feature filters used to handle an impression event
+    /// </summary>
+    public interface IFeatureImpressionsListener
+    {
+        public bool HandleImpression(string feature, bool treatment);
```
hermogenes, line 11: Shouldn't it be ideally async?

src/Microsoft.FeatureManagement/FeatureManager.cs
```diff
@@ -160,9 +164,19 @@ private async Task<bool> IsEnabledAsync<TContext>(string feature, TContext appCo
                 await sessionManager.SetAsync(feature, enabled).ConfigureAwait(false);
             }
 
+            sendImpression(feature, enabled);
+
             return enabled;
         }
 
+        private void sendImpression(string feature, bool enabled)
```
hermogenes, line 172:
```suggestion
        private void SendImpression(string feature, bool enabled)
```

hermogenes, line 172: Naming convention
_Consider the other comment to change it to async then the signature would be slightly different._
```suggestion
        private void SendImpression(string feature, bool enabled)
```

zhiyuanliang-ms:
Hi, @joshsirota

Are you still tracking this PR? Could you give more context about this PR?

PR #179, opened 2022-05-18
Fixed targeting assignment precedence.

jimmyca15:
This PR addresses an issue regarding audience matching precedence when assigning variants via targeting.

If the current user is targeted by name directly, the associated variant should always win. This has the highest precedence.

Next is a group match.

Finally, if none of the above matches occur, default rollout percentage should be used to select a matching variant. Otherwise, the default variant should be selected.

1. User
2. Group
3. Default Rollout
4. Default Variant

This is already detailed in the README, but the existing code did not behave this way.

### DIFF
--- a/src/Microsoft.FeatureManagement/Targeting/ContextualTargetingFeatureVariantAssigner.cs
+++ b/src/Microsoft.FeatureManagement/Targeting/ContextualTargetingFeatureVariantAssigner.cs
@@ -67,24 +67,24 @@ namespace Microsoft.FeatureManagement.Assigners
 
             FeatureVariant variant = null;
 
-            double cumulativePercentage = 0;
+            var lookup = new Dictionary<FeatureVariant, TargetingFilterSettings>();
 
-            var cumulativeGroups = new Dictionary<string, double>(
-                _options.IgnoreCase ? StringComparer.OrdinalIgnoreCase :
-                                      StringComparer.Ordinal);
-
+            //
+            // Check users
             foreach (FeatureVariant v in featureDefinition.Variants)
             {
                 TargetingFilterSettings targetingSettings = v.AssignmentParameters.Get<TargetingFilterSettings>();
 
-                if (targetingSettings == null)
+                //
+                // Put in lookup table to avoid repeatedly creating targeting settings
+                lookup[v] = targetingSettings;
+
+                if (targetingSettings == null &&
+                    v.Default)
                 {
-                    if (v.Default)
-                    {
-                        //
-                        // Valid to omit audience for default variant
-                        continue;
-                    }
+                    //
+                    // Valid to omit audience for default variant
+                    continue;
                 }
 
                 if (!TargetingEvaluator.TryValidateSettings(targetingSettings, out string paramName, out string reason))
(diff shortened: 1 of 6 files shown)


jimmyca15:
@zhenlan @avanigupta @mrm9084 

Please help with reviewing this PR if you can.

src/Microsoft.FeatureManagement/Targeting/ContextualTargetingFeatureVariantAssigner.cs
```diff
@@ -67,69 +67,128 @@ public ValueTask<FeatureVariant> AssignVariantAsync(FeatureVariantAssignmentCont
 
             FeatureVariant variant = null;
 
-            double cumulativePercentage = 0;
+            var lookup = new Dictionary<FeatureVariant, TargetingFilterSettings>();
```
avanigupta, line 70: Will this dictionary be thread safe?

src/Microsoft.FeatureManagement/Targeting/ContextualTargetingFeatureVariantAssigner.cs
```diff
@@ -67,69 +67,128 @@ public ValueTask<FeatureVariant> AssignVariantAsync(FeatureVariantAssignmentCont
...
+            foreach (FeatureVariant v in featureDefinition.Variants)
+            {
+                TargetingFilterSettings targetingSettings = v.AssignmentParameters.Get<TargetingFilterSettings>();
+
+                //
+                // Put in lookup table to avoid repeatedly creating targeting settings
+                lookup[v] = targetingSettings;
+
+                if (targetingSettings == null &&
+                    v.Default)
```
avanigupta, line 83: Should this condition be OR instead of AND?
What is the purpose of specifying any audience for default variant?

src/Microsoft.FeatureManagement/Targeting/ContextualTargetingFeatureVariantAssigner.cs
```diff
@@ -67,69 +67,128 @@ public ValueTask<FeatureVariant> AssignVariantAsync(FeatureVariantAssignmentCont
...
-                    {
-                        //
-                        // Valid to omit audience for default variant
-                        continue;
-                    }
+                    continue;
                 }
 
-                if (!TargetingEvaluator.TryValidateSettings(targetingSettings, out string paramName, out string reason))
+                AccumulateGroups(targetingSettings.Audience, cumulativeGroups);
```
avanigupta, line 123: >targetingSettings.Audience [](http://example.com/codeflow?start=33&length=26)

Should we pass `targetingSettings.Audience.Groups` now that this method specifically accumulates groups?

src/Microsoft.FeatureManagement/Targeting/ContextualTargetingFeatureVariantAssigner.cs
```diff
@@ -67,69 +67,128 @@ public ValueTask<FeatureVariant> AssignVariantAsync(FeatureVariantAssignmentCont
 
             FeatureVariant variant = null;
```
avanigupta, line 68: `variant` is not used anymore.

src/Microsoft.FeatureManagement/Targeting/ContextualTargetingFeatureVariantAssigner.cs
jimmyca15: `Dictionary` itself is not thread safe, but since this is a function local variable we don't have to worry about multi-threaded access.

src/Microsoft.FeatureManagement/Targeting/ContextualTargetingFeatureVariantAssigner.cs
jimmyca15: It is intentionally AND. 

Here's an example setup with default variant that has an audience.

Option A is the default, Option B targets 50% of users by default rollout percentage.
Since A is the default, this setup means A will target 50% of the user base by default.
But we specifically want "Jeff" to use option A. So we setup an audience for option A and target Jeff.

src/Microsoft.FeatureManagement/Targeting/ContextualTargetingFeatureVariantAssigner.cs
jimmyca15: I think that's a great idea.

src/Microsoft.FeatureManagement/Targeting/TargetingEvaluator.cs
```diff
@@ -36,29 +39,95 @@ public static bool IsTargeted(TargetingFilterSettings settings, ITargetingContex
...
+        }
+
+        /// <summary>
+        /// Determine if a targeting context is targeted by presence in a group
+        /// </summary>
+        public static bool IsTargeted(
+            ITargetingContext targetingContext,
+            IEnumerable<GroupRollout> groups,
+            bool ignoreCase,
+            string hint)
```
avanigupta, line 106: >string hint [](http://example.com/codeflow?start=12&length=11)

Should we add null check for `hint`?

avanigupta:
>                         }

Nit: 
`return IsTargeted(...);`

---
Refers to: src/Microsoft.FeatureManagement/Targeting/TargetingEvaluator.cs:139 in 67b59b3. [](commit_id = 67b59b3f393e8bd4a6b0bcc802c3039ba96d5d48, deletion_comment = False)

src/Microsoft.FeatureManagement/Targeting/TargetingEvaluator.cs
jimmyca15: It's probably a good idea. But it looks like we don't enforce this at public entry point of feature manager. So to solve it right, I think I'll want to do a separate PR where I add enforcement through the call-chain.

src/Microsoft.FeatureManagement/Targeting/ContextualTargetingFeatureVariantAssigner.cs
```diff
@@ -67,69 +67,128 @@ public ValueTask<FeatureVariant> AssignVariantAsync(FeatureVariantAssignmentCont
...
 
-                AccumulateAudience(targetingSettings.Audience, cumulativeGroups, ref cumulativePercentage);
+            //
+            // Check default rollout percentage
+            foreach (FeatureVariant v in featureDefinition.Variants)
+            {
+                TargetingFilterSettings targetingSettings = lookup[v];
 
-                if (TargetingEvaluator.IsTargeted(targetingSettings, targetingContext, _options.IgnoreCase, featureDefinition.Name))
+                if (targetingSettings == null)
```
avanigupta, line 143: >if (targetingSettings == null) [](http://example.com/codeflow?start=16&length=30)

Should  we skip those variants that do not have any DefaultRolloutPercentage?

jimmyca15: I guess you mean to special case if `DefaultRolloutPercentage` is 0 ? We could, but the normal logic handles it fine.

jimmyca15:
>                         }

That changes the logic. 

Currently, if any of the groups return a positive targeting result, then we return true.

---
In reply to: [1136523253](https://github.com/microsoft/FeatureManagement-Dotnet/pull/179#issuecomment-1136523253) [](http://example.com/codeflow?ancestors=1136523253,1136523253)

---
Refers to: src/Microsoft.FeatureManagement/Targeting/TargetingEvaluator.cs:139 in 67b59b3. [](commit_id = 67b59b3f393e8bd4a6b0bcc802c3039ba96d5d48, deletion_comment = False)

avanigupta (approved):

PR #184, opened 2022-06-03
Added a pointer to the v2 README from the v3 readme.

jimmyca15:


README.md
```diff
@@ -20,6 +20,12 @@ Here are some of the benefits of using this library:
 
 **API Reference**: https://go.microsoft.com/fwlink/?linkid=2091700
 
+## Content Version
+
+The content of this README relates to version 3 of the Microsoft.FeatureManagement library.
```
avanigupta, line 25: How about calling it "version 3 and above"?
Similarly, "version 2 and lower"?

avanigupta (approved):

README.md
jimmyca15: I think at any point in time a given readme will only apply to a single major version. For example, the v2 readme mentioned here describes features that aren't available in v1. So I'd be weary of attaching the readme to more than one version.

avanigupta: Makes sense, thanks!

Issue #28, opened 2019-11-29
Feature filters are combined with a logical OR


RemyBoyer:
Hello,

While experimenting with the sdk, we noticed that when evaluating the filters and the contextual filters, the `FeatureManager` internal loop will break and return `true` for the first filter that returns `true`.

We are planning to use heavily the sdk, but we feel that our needs are mainly steering towards a logical AND.

For example, we will want to enable a feature for one specific country (contextual filter), but only for a defined percentage of them (filter). This would look like this in the app configuration store :

```
"conditions": {
    "client_filters": [
        {
            "name": "CountryContextual",
            "parameters": {
                "Country": [
                    "FR",
                    "IT"
                ]
            }
        },
        {
            "name": "Microsoft.Percentage",
            "parameters": {
                "Value": "50"
            }
        }
    ]
}
```

We imagine we might be able to create "meta" filters that would contain a logical operator and a list of sub filters, which can also be recursive. This would solve pretty much any use case we could imagine, but it seems to be a little over-complicated.

We could also create our own implementation of `IFeatureManager`, but the required dependency  `IFeatureSettingsProvider` is currently internal.

In our opinion, the most sensible default combination operator should be an AND.

What's your opinion on that? Do you think of any built-in or alternative way of reaching our goals?

jimmyca15:
Hello @RemyBoyer ,

Trying to introduce AND vs OR behavior in the evaluation of feature filters can be complex. It seems to me that a better approach to this issue would be to create the country filter to be percentage aware.

```
        {
            "name": "CountryContextual",
            "parameters": {
                "Countries": [
                    {
                      "Code": "FR",
                      "Percentage": 10
                    },
                    {
                      "Code": "IT",
                      "Percentage": 30
                    }
                ]
            }
        }
```

The `Microsoft.FeatureManagement` library is perfectly fine to handle a filter like this. You may find a disconnect when trying to set up a filter like this in the Azure App Configuration portal since you mentioned what it would look like in Azure App Configuration. When configuring feature filters in the Azure App Configuration portal, it is difficult to model arrays (see issues #27 and [147](https://github.com/Azure/AppConfiguration/issues/147)). It is even more difficult to model arrays of objects. This is something we are currently in the process of revamping.

Ignoring the current portal configuration issue that I mentioned, does that sound like an adequate solution?

RemyBoyer:
Hello @jimmyca15 ,

[...]

PR #192, opened 2022-06-22
Allow a feature flag to require all registered feature filters be enabled.

jimmyca15:
In response to #28.

## Combine feature filters with AND

By default, feature filters are evaluated using a logical OR operation. If any feature filter that is registered for a feature is enabled, then the feature is considered to be enabled. This PR adds the capability to customize this so that feature filters are combined with a logical AND operation. This means all registered feature filters would need to be enabled for the feature to be considered enabled.

### Configuration

Overriding feature filter evaluation to use AND can be done by specifying the `RequirementType` for a given feature flag.

**RequirementType**
* **Any**: Default. If any feature filter is enabled, the feature will be considered enabled.
* **All**: Requires all feature filters to be enabled to enable a feature

```json
{
    "FeatureManagement": {
        "FeatureFlags": {
            "MyAndFeature": {
                "RequirementType": "All",
                "EnabledFor": [
                    {
                        "Name": "Test",
                        "Parameters": {}
                    },
                    {
                        "Name": "Test2",
                        "Parameters": {}
                    }
                ]
            }
        }
    }
}
```

### DIFF
--- a/src/Microsoft.FeatureManagement/ConfigurationFeatureFlagDefinitionProvider.cs
+++ b/src/Microsoft.FeatureManagement/ConfigurationFeatureFlagDefinitionProvider.cs
@@ -22,6 +22,7 @@ namespace Microsoft.FeatureManagement
         private const string FeatureManagementSectionName = "FeatureManagement";
         private const string FeatureFlagDefinitionsSectionName = "FeatureFlags";
         private const string FeatureFiltersSectionName = "EnabledFor";
+        private const string RequirementTypeKeyword = "RequirementType";
         private readonly IConfiguration _configuration;
         private readonly ConcurrentDictionary<string, FeatureFlagDefinition> _featureFlagDefinitions;
         private IDisposable _changeSubscription;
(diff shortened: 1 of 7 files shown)


rossgrambo:
Closing in favor of https://github.com/microsoft/FeatureManagement-Dotnet/pull/221

src/Microsoft.FeatureManagement/ConfigurationFeatureFlagDefinitionProvider.cs
```diff
@@ -149,6 +152,20 @@ We support
             }
             else
             {
+                string rawRequirementType = configurationSection[RequirementTypeKeyword];
+
+                if (!string.IsNullOrEmpty(rawRequirementType))
+                {
+                    if (!Enum.TryParse(rawRequirementType, true, out RequirementType r))
```
avanigupta, line 159: We dont need a new variable for RequirementType.
We can do this:
`Enum.TryParse(rawRequirementType, true, out requirementType)`

Then we dont even need to do `requirementType = r` later.

Issue #123, opened 2021-04-17
TargetFilter only returns positive group evaluation result


yuanyi2121:
There is an issue with this line: https://github.com/microsoft/FeatureManagement-Dotnet/blob/df0f9d20db4c0a8f2c5bbb3362db614ff1c50944/src/Microsoft.FeatureManagement/Targeting/ContextualTargetingFilter.cs#L86

it only returns true when a group evaluation returned true, otherwise it will use the default percentage.

Example:
Setting: 
Default: 100%
Audience: GroupA = 100, GroupB = 0

Evaluate:
new TargetingContext
{ Groups = new List<string>() { "GroupB " } }

The return will be true because GroupB always evaluate as false, hence the default percentage applied.

jimmyca15:
That's the expected behavior.

yuanyi2121:
Hmm, if I have 100 groups and only want to turn a feature off for one group, how am I supposed to do that? I think the expected behavior should be if I want a group to be 0%, they should be 0%.

jimmyca15:
The 100 groups would need to be specified with rollout percentages of 100 with the singular disabled group being omitted.

yuanyi2121:
Yes, that's the point - right now I have to explicitly define 99 groups with 100% and one with 0% (or set 0% as default). And this is the better case where I can figure what the other 99 groups are. What if I want to disable a feature for a certain IP address?

The correct behavior should be, IMHO - if there are any group matches, the code should never fall back to the default settings, it should instead aggregate the result of all the matched groups.

jimmyca15:
@yuanyi2121 thanks for the feedback. We can consider adding an option to the targeting parameters to allow a behavior like that.

jimmyca15:
Resolved with the addition of exclusion in #218. The new exclusion feature is available in the latest stable release (2.6.1)

PR #218, opened 2023-03-22
Adds exclude logic to Targeting and Audience, adds unit test, updates example to include exclude

rossgrambo:
In response to #123.

# Adds Exclusion to Audience
Audiences have historically been a list of users, groups, and rollout percentages. This change adds an additional field in Audience called **Exclusion**. This field holds a list of **Users** and **Groups**. These users and groups are evaluated before the rest of the Audience. If there is a match, the feature flag immediately returns false. 

If the targeting filter has sibling filters, they will still be evaluated after a false result.

# Example
`appsettings.json`
```json
{
  "Name": "Targeting",
  "Parameters": {
    "Audience": {
      "Users": [
        "Jeff",
        "Anne"
      ],
      "Groups": [
        {
          "Name": "Management",
          "RolloutPercentage": 100
        },
        {
          "Name": "TeamMembers",
          "RolloutPercentage": 45
        }
      ],
      "DefaultRolloutPercentage": 20,
      "Exclusion": {
        "Users": [
          "Anne",
          "Phil"
        ],
        "Groups": [
          "Contractor"
        ]
      }
    }
  }
}
```

`Evaluation`
```csharp
.IsEnabledAsync(betaFeature, new TargetingContext { UserId = "Anne" }}) // returns false
.IsEnabledAsync(betaFeature, new TargetingContext { UserId = "Phil", Groups = new List<string>() { "Management" }}}) // returns false
.IsEnabledAsync(betaFeature, new TargetingContext { UserId = "Bill", Groups = new List<string>() { "Contractor" }}}) // returns false
.IsEnabledAsync(betaFeature, new TargetingContext { UserId = "Bill", Groups = new List<string>() { "Management", "Contractor" }}}) // returns false
```

### DIFF
--- a/src/Microsoft.FeatureManagement/Targeting/ContextualTargetingFilter.cs
+++ b/src/Microsoft.FeatureManagement/Targeting/ContextualTargetingFilter.cs
@@ -35,6 +35,7 @@ namespace Microsoft.FeatureManagement.FeatureFilters
         }
 
         private StringComparison ComparisonType => _options.IgnoreCase ? StringComparison.OrdinalIgnoreCase : StringComparison.Ordinal;
+        private StringComparer ComparerType => _options.IgnoreCase ? StringComparer.OrdinalIgnoreCase : StringComparer.Ordinal;
 
         /// <summary>
         /// Performs a targeting evaluation using the provided <see cref="TargetingContext"/> to determine if a feature should be enabled.
@@ -77,7 +78,7 @@ namespace Microsoft.FeatureManagement.FeatureFilters
                 // Check if the user is in a group within exclusion
                 if (targetingContext.Groups != null &&
                     settings.Audience.Exclusion.Groups != null &&
-                    settings.Audience.Exclusion.Groups.Any(group => targetingContext.Groups.Any(targetingGroup => targetingGroup.Equals(group, ComparisonType))))
+                    settings.Audience.Exclusion.Groups.Any(group => targetingContext.Groups.Contains(group, ComparerType)))
                 {
                     return Task.FromResult(false);
                 }
--- a/src/Microsoft.FeatureManagement/Targeting/ContextualTargetingFilter.cs
+++ b/src/Microsoft.FeatureManagement/Targeting/ContextualTargetingFilter.cs
@@ -5,6 +5,7 @@ using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Options;
 using System;
+using System.Collections.Generic;
 using System.Linq;
 using System.Security.Cryptography;
 using System.Text;
@@ -59,6 +60,27 @@ namespace Microsoft.FeatureManagement.FeatureFilters
             if (!TryValidateSettings(settings, out string paramName, out string message))
             {
                 throw new ArgumentException(message, paramName);
+            }
+
+            if (settings.Audience.Exclusion != null)
+            {
+                //
+                // Check if the user is in the exclusion directly
+                if (targetingContext.UserId != null &&
+                    settings.Audience.Exclusion.Users != null &&
+                    settings.Audience.Exclusion.Users.Any(user => targetingContext.UserId.Equals(user, ComparisonType)))
+                {
+                    return Task.FromResult(false);
+                }
+
+                //
+                // Check if the user is in a group within exclusion
+                if (targetingContext.Groups != null &&
+                    settings.Audience.Exclusion.Groups != null &&
+                    settings.Audience.Exclusion.Groups.Any(group => targetingContext.Groups.Any(targetingGroup => targetingGroup.Equals(group, ComparisonType))))
+                {
+                    return Task.FromResult(false);
+                }
             }
 
             //
--- a/examples/TargetingConsoleApp/Identity/InMemoryUserRepository.cs
+++ b/examples/TargetingConsoleApp/Identity/InMemoryUserRepository.cs
@@ -74,6 +74,19 @@ namespace Consoto.Banking.AccountService.Identity
                     "TeamMembers"
                 }
             },
+            new User
+            {
+                Id = "Anne",
+                Groups = Enumerable.Empty<string>()
+            },
+            new User
+            {
+                Id = "Chuck",
+                Groups =  new List<string>()
+                {
+                    "Contractor"
+                }
+            },
         };
 
         public Task<User> GetUser(string id)
--- a/examples/TargetingConsoleApp/appsettings.json
+++ b/examples/TargetingConsoleApp/appsettings.json
@@ -7,7 +7,8 @@
           "Parameters": {
             "Audience": {
               "Users": [
-                "Jeff"
+                "Jeff",
+                "Anne"
               ],
               "Groups": [
                 {
@@ -19,7 +20,16 @@
                   "RolloutPercentage": 45
                 }
               ],
-              "DefaultRolloutPercentage": 20
+              "DefaultRolloutPercentage": 20,
+              "Exclusion": {
+                "Users": [
+                  "Anne",
+                  "Phil"
+                ],
+                "Groups": [
+                  "Contractor"
+                ]
+              }
             }
           }
         }
--- a/src/Microsoft.FeatureManagement/Targeting/Audience.cs
+++ b/src/Microsoft.FeatureManagement/Targeting/Audience.cs
@@ -24,5 +24,10 @@ namespace Microsoft.FeatureManagement.FeatureFilters
         /// Includes users in the audience based off a percentage of the total possible audience. Valid values range from 0 to 100 inclusive.
         /// </summary>
         public double DefaultRolloutPercentage { get; set; }
+
+        /// <summary>
+        /// Excludes a basic audience from this audience.
+        /// </summary>
+        public BasicAudience Exclusion { get; set; }
     }
 }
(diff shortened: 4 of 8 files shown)


rossgrambo:
I should call out I went with the wording "Exclude" rather than "Except". Using Except/Exception in the code was not very clear. If we still want the field named Except, we can do that.

src/Microsoft.FeatureManagement/Targeting/ContextualTargetingFilter.cs
```diff
@@ -61,6 +61,27 @@ public Task<bool> EvaluateAsync(FeatureFilterEvaluationContext context, ITargeti
...
+                    settings.Audience.Exclude.Users.Any(user => targetingContext.UserId.Equals(user, ComparisonType)))
+                {
+                    return Task.FromResult(false);
+                }
+
+                //
+                // Check if the user is in a group that is being excluded
+                if (targetingContext.Groups != null &&
+                    settings.Audience.Exclude.Groups != null &&
+                    settings.Audience.Exclude.Groups.Intersect(targetingContext.Groups).Any())
```
jimmyca15, line 79: Intersect isn't necessary, we only need a single occurence.

src/Microsoft.FeatureManagement/Targeting/AudienceExclusion.cs
```diff
@@ -0,0 +1,23 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+//
+using System.Collections.Generic;
+
+namespace Microsoft.FeatureManagement.FeatureFilters
+{
+	/// <summary>
+	/// Excludes users and groups from the audience.
```
jimmyca15, line 9: The type itself should really just define users and groups. The exclusion part should come from it's presence as the `Exclude` property of `Audience.`

src/Microsoft.FeatureManagement/Targeting/AudienceExclusion.cs
```diff
@@ -0,0 +1,23 @@
...
+	/// </summary>
+	public class AudienceExclusion
```
jimmyca15, line 11: The name is a bit problematic for me. It's really just a simplified audience definition. Nothing about the properties make it `exclusion` related.

src/Microsoft.FeatureManagement/Targeting/ContextualTargetingFilter.cs
rossgrambo: Agreed, but .Intersect is O(N) while .Any is O(N^2) but faster when it's smaller and uses no memory. 

This answer has a good comparison of the two- https://stackoverflow.com/questions/17153688/intersect-and-any-or-contains-and-any-which-is-more-efficient-to-find-at-least

If you feel stronger about using .Any here, we can do that.

src/Microsoft.FeatureManagement/Targeting/AudienceExclusion.cs
rossgrambo: Completely agree. I was deciding if the library leaned towards generic or more specific object oriented. I saw TargetingContext and took that as leaning more towards specificity (but that's not a straightforward comparison). That, and, I couldn't land on a good name for a generic version.

What do you prefer?
`AudienceCompact`
`SimplifiedAudience`
`BasicAudience`
`StandardAudience`

src/Microsoft.FeatureManagement/Targeting/AudienceExclusion.cs
rossgrambo: Treating this comment as a duplicate of https://github.com/microsoft/FeatureManagement-Dotnet/pull/218#discussion_r1145479412

src/Microsoft.FeatureManagement/Targeting/ContextualTargetingFilter.cs
jimmyca15: But I also don't feel to strongly. If you like Intersect keep it, although I'd put a comment about why to use it over Any.

src/Microsoft.FeatureManagement/Targeting/AudienceExclusion.cs
jimmyca15: I also thought of `SimplifiedAudience` `BasicAudience`. I also considered what is different between this and our existing `Audience`. It's the absence of percentages. I then considered `DeterministicAudience`.

src/Microsoft.FeatureManagement/Targeting/ContextualTargetingFilter.cs
rossgrambo: I think that's good enough reasoning. I'll go ahead and switch it to Any.

src/Microsoft.FeatureManagement/Targeting/AudienceExclusion.cs
rossgrambo: I thought the same thing with absence of percentages but landed on `AudienceWithoutRollout`. `DeterministicAudience` is cleaner. 

I'm leaning towards `BasicAudience` because I think it's the most intuitive and least future-restrictive. I'll go ahead and make changes towards that. Let me know if you feel strongly about one of the other options.

zhenlan:
> I should call out I went with the wording "Exclude" rather than "Except". Using Except/Exception in the code was not very clear. If we still want the field named Except, we can do that.

I like it. I was thinking about it too. "Exclude" is a verb, a little bit weird as a property name. Do we like to use a noun, "Exclusion", instead?

jimmyca15:
> > I should call out I went with the wording "Exclude" rather than "Except". Using Except/Exception in the code was not very clear. If we still want the field named Except, we can do that.
> 
> I like it. I was thinking about it too. "Exclude" is a verb, a little bit weird as a property name. Do we like to use a noun, "Exclusion", instead?

Also offering that symmetry with exiting name in the configuration, `enabledFor`, would be `excluded`

zhenlan:
> Also offering that symmetry with existing name in the configuration, `enabledFor`, would be `excluded`

Honestly speaking, "except" works better with "enabledFor". I think that's the reason we picked the "except" in the first place. I am just concerned we may over complicate the design. If we choose to use "Exclusion" for the AppConfig schema, we might just use the same everywhere else. It will make the tooling work easier.

jimmyca15:
@zhenlan

> If we choose to use "Exclusion" for the AppConfig schema, we might just use the same everywhere else.

You lost me here. What are you referring to as the AppConfig schema? and what are you referring to as everywhere else?

From my thinking, AppConfig schema plays no role in the decision here. This name will only show up in the targeting configuration which is AppConfig agnostic.

rossgrambo:
>  If we choose to use "Exclusion" for the AppConfig schema, we might just use the same everywhere else.
I believe this is the only occurrence of the word. We're defining what the field name in an Audience should be. And then that will be reflected in both the backend and featuremanagement schemas. It will also likely be reflected in the portal UI. 

Agreed- `Except` is very clear. I think `Exclude` or `ExcludedFor` are still pretty readable. `Exclusion` makes it sound like a class or something I need more context to understand- but I'm likely just overthinking it. 

I also prefer `ExcludedFor` over `Excluded` as I think it hints that the contents are just users/groups. And strictly matches `EnabledFor`.

All of that said- I think I'll move forward with `Exclusion` unless people feel strongly against it.

rossgrambo:
Synced with Jimmy- We aligned on `Except`. 

In this context, we're just defining an Audience (not a feature flag). Except is clearly calling out that the audience is as defined- except for these users/groups. 

Also calling out that this field is the only place this idea of Except/Exclude/Exclusion exists. It's on the Audience object- so it will be seen in both the backend and FeatureManagement schemas, but there's no other uses of this word yet.

Not going to merge until we hear sign-off from you as well @zhenlan , but for now I'll go ahead and make the code changes towards "Except".

rossgrambo:
Changed the naming to `Exclusion`!

src/Microsoft.FeatureManagement/Targeting/ContextualTargetingFilter.cs
```diff
@@ -61,6 +61,27 @@ public Task<bool> EvaluateAsync(FeatureFilterEvaluationContext context, ITargeti
...
+                    settings.Audience.Exclusion.Users.Any(user => targetingContext.UserId.Equals(user, ComparisonType)))
+                {
+                    return Task.FromResult(false);
+                }
+
+                //
+                // Check if the user is in a group within exclusion
+                if (targetingContext.Groups != null &&
+                    settings.Audience.Exclusion.Groups != null &&
+                    settings.Audience.Exclusion.Groups.Any(group => targetingContext.Groups.Contains(group)))
```
jimmyca15, line 79: group comparison needs the comparison type for case sensitivity.

rossgrambo: Updated. Had to do a nested `.Any( .Any(` since `Linq.Contains(` takes a `StringComparer` while the rest of the code uses `StringComparison`. Converting between the two isn't available in `netstandard2.0`, which we still have as a target.

jimmyca15: Ah, I see. I don't mind to add StringComparison as private member alongside StringComparer as well to avoid the nested `Any`, if you feel it's cleaner.

If you stick with nested Any though, I would like it if we broke up the line a bit since it's a bit long in length. Suggested

```
                    settings.Audience.Exclusion.Groups.Any(
                        group => targetingContext.Groups.Any(
                            targetingGroup => targetingGroup.Equals(group, ComparisonType))))


```

jimmyca15 (approved):

avanigupta (approved):

jimmyca15 (approved):

PR #221, opened 2023-03-29
Adds Requirement Type All and Any to FeatureFlags

rossgrambo:
In response to issue https://github.com/microsoft/FeatureManagement-Dotnet/issues/28.
And as a rewrite of PR https://github.com/microsoft/FeatureManagement-Dotnet/pull/192.

# Add Requirement Type All and Any to FeatureFlags
 
Filters on Feature Flags have always used "Any" logic. Our execution goes through each filter and when a single one returns "true", we end execution and return true. This PR adds "All" logic. For "All", the execution instead goes through each filter and when a single one returns "false", we end execution, returning false. Otherwise we return true.

In short- this PR: 
* Moves RequirementType from AspNetCore to the main .Net library.
* Defaults all unspecified Feature Flags to use "Any" logic
* Adds "All" logic to Feature Flag evaluation

## Example
```json
"AllFilterFeature": {
  "RequirementType": "All",
  "EnabledFor": [
    {
      "Name": "Targeting",
      "Parameters": {
        "Audience": {
          "Users": [
            "Jeff",
            "Alicia"
          ]
        }
      }
    },
    {
      "Name": "TimeWindow",
      "Parameters": {
        "Start": "Wed, 1 April 2023 00:00:00 GMT",
        "End": "Wed, 30 April 2023 23:59:59 GMT"
      }
    }
  ]
```

## Evaluation
```csharp
// Called before TimeWindow:
.IsEnabledAsync(AllFilterFeature, new TargetingContext { UserId = "Jeff" }}) // returns false
.IsEnabledAsync(AllFilterFeature, new TargetingContext { UserId = "Anne" }}) // returns false

// Called during TimeWindow:
.IsEnabledAsync(AllFilterFeature, new TargetingContext { UserId = "Jeff" }}) // returns true
.IsEnabledAsync(AllFilterFeature, new TargetingContext { UserId = "Anne" }}) // returns false
```

### DIFF
--- a/src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
+++ b/src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
@@ -162,7 +162,7 @@ namespace Microsoft.FeatureManagement
                 {
                     if (!Enum.TryParse(rawRequirementType, true, out requirementType))
                     {
-                        throw new ArgumentException($"Invalid requirement type '{rawRequirementType}' for feature '{configurationSection.Key}'.");
+                        throw new FeatureManagementException(FeatureManagementError.InvalidConfiguration, $"Invalid requirement type '{rawRequirementType}' for feature '{configurationSection.Key}'.");
                     }
                 }
 
--- a/src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
+++ b/src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
@@ -158,12 +158,13 @@ namespace Microsoft.FeatureManagement
             {
                 string rawRequirementType = configurationSection[RequirementTypeKeyword];
 
-                if (!string.IsNullOrEmpty(rawRequirementType))
+                //
+                // If requirement type is specified, parse it and set the requirementType variable
+                if (!string.IsNullOrEmpty(rawRequirementType) && !Enum.TryParse(rawRequirementType, true, out requirementType))
                 {
-                    if (!Enum.TryParse(rawRequirementType, true, out requirementType))
-                    {
-                        throw new FeatureManagementException(FeatureManagementError.InvalidConfiguration, $"Invalid requirement type '{rawRequirementType}' for feature '{configurationSection.Key}'.");
-                    }
+                    throw new FeatureManagementException(
+                        FeatureManagementError.InvalidValue, 
+                        $"Invalid requirement type '{rawRequirementType}' for feature '{configurationSection.Key}'.");
                 }
 
                 IEnumerable<IConfigurationSection> filterSections = configurationSection.GetSection(FeatureFiltersSectionName).GetChildren();
--- a/src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
+++ b/src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
@@ -163,7 +163,7 @@ namespace Microsoft.FeatureManagement
                 if (!string.IsNullOrEmpty(rawRequirementType) && !Enum.TryParse(rawRequirementType, true, out requirementType))
                 {
                     throw new FeatureManagementException(
-                        FeatureManagementError.InvalidValue, 
+                        FeatureManagementError.InvalidConfigurationSetting, 
                         $"Invalid requirement type '{rawRequirementType}' for feature '{configurationSection.Key}'.");
                 }
 
--- a/src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
+++ b/src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
@@ -160,7 +160,7 @@ namespace Microsoft.FeatureManagement
 
                 //
                 // If requirement type is specified, parse it and set the requirementType variable
-                if (!string.IsNullOrEmpty(rawRequirementType) && !Enum.TryParse(rawRequirementType, true, out requirementType))
+                if (!string.IsNullOrEmpty(rawRequirementType) && !Enum.TryParse(rawRequirementType, ignoreCase: true, out requirementType))
                 {
                     throw new FeatureManagementException(
                         FeatureManagementError.InvalidConfigurationSetting, 
--- a/src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
+++ b/src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
@@ -18,6 +18,7 @@ namespace Microsoft.FeatureManagement
     sealed class ConfigurationFeatureDefinitionProvider : IFeatureDefinitionProvider, IDisposable
     {
         private const string FeatureFiltersSectionName = "EnabledFor";
+        private const string RequirementTypeKeyword = "RequirementType";
         private readonly IConfiguration _configuration;
         private readonly ConcurrentDictionary<string, FeatureDefinition> _definitions;
         private IDisposable _changeSubscription;
@@ -121,9 +122,15 @@ namespace Microsoft.FeatureManagement
             },
             myAlwaysDisabledFeature2: {
               enabledFor: false
-            }
+            },
+            myAllRequiredFilterFeature: {
+                requirementType: "all"
+                enabledFor: [ "myFeatureFilter1", "myFeatureFilter2" ],
+            },
 
             */
+
+            RequirementType requirementType = RequirementType.Any;
 
             var enabledFor = new List<FeatureFilterConfiguration>();
 
@@ -149,6 +156,16 @@ namespace Microsoft.FeatureManagement
             }
             else
             {
+                string rawRequirementType = configurationSection[RequirementTypeKeyword];
+
+                if (!string.IsNullOrEmpty(rawRequirementType))
+                {
+                    if (!Enum.TryParse(rawRequirementType, true, out requirementType))
+                    {
+                        throw new ArgumentException($"Invalid requirement type '{rawRequirementType}' for feature '{configurationSection.Key}'.");
+                    }
+                }
+
                 IEnumerable<IConfigurationSection> filterSections = configurationSection.GetSection(FeatureFiltersSectionName).GetChildren();
 
                 foreach (IConfigurationSection section in filterSections)
@@ -170,7 +187,8 @@ namespace Microsoft.FeatureManagement
             return new FeatureDefinition()
             {
                 Name = configurationSection.Key,
-                EnabledFor = enabledFor
+                EnabledFor = enabledFor,
+                RequirementType = requirementType
             };
         }
 
--- a/src/Microsoft.FeatureManagement/FeatureManagementError.cs
+++ b/src/Microsoft.FeatureManagement/FeatureManagementError.cs
@@ -24,7 +24,7 @@ namespace Microsoft.FeatureManagement
         MissingFeature,
 
         /// <summary>
-        /// A configuration had conflicting values.
+        /// There was a conflict in the feature management system.
         /// </summary>
         Conflict,
 
--- a/src/Microsoft.FeatureManagement/FeatureManagementError.cs
+++ b/src/Microsoft.FeatureManagement/FeatureManagementError.cs
@@ -9,18 +9,23 @@ namespace Microsoft.FeatureManagement
     public enum FeatureManagementError
     {
         /// <summary>
-        /// A feature filter that was listed for feature evaluation was not found.
-        /// </summary>
-        MissingFeatureFilter,
-
-        /// <summary>
         /// A feature filter configured for the feature being evaluated is an ambiguous reference to multiple registered feature filters.
         /// </summary>
         AmbiguousFeatureFilter,
 
         /// <summary>
+        /// A configuration error is present in the feature management system.
+        /// </summary>
+        InvalidConfiguration,
+
+        /// <summary>
         /// A feature that was requested for evaluation was not found.
         /// </summary>
-        MissingFeature
+        MissingFeature,
+
+        /// <summary>
+        /// A feature filter that was listed for feature evaluation was not found.
+        /// </summary>
+        MissingFeatureFilter
     }
 }
--- a/src/Microsoft.FeatureManagement/FeatureManagementError.cs
+++ b/src/Microsoft.FeatureManagement/FeatureManagementError.cs
@@ -9,14 +9,14 @@ namespace Microsoft.FeatureManagement
     public enum FeatureManagementError
     {
         /// <summary>
+        /// A feature filter that was listed for feature evaluation was not found.
+        /// </summary>
+        MissingFeatureFilter,
+
+        /// <summary>
         /// A feature filter configured for the feature being evaluated is an ambiguous reference to multiple registered feature filters.
         /// </summary>
         AmbiguousFeatureFilter,
-
-        /// <summary>
-        /// A configuration error is present in the feature management system.
-        /// </summary>
-        InvalidConfiguration,
 
         /// <summary>
         /// A feature that was requested for evaluation was not found.
@@ -24,8 +24,13 @@ namespace Microsoft.FeatureManagement
         MissingFeature,
 
         /// <summary>
-        /// A feature filter that was listed for feature evaluation was not found.
+        /// A configuration had conflicting values.
         /// </summary>
-        MissingFeatureFilter
+        Conflict,
+
+        /// <summary>
+        /// The given value was invalid.
+        /// </summary>
+        InvalidValue
     }
 }
--- a/src/Microsoft.FeatureManagement/FeatureManagementError.cs
+++ b/src/Microsoft.FeatureManagement/FeatureManagementError.cs
@@ -29,8 +29,8 @@ namespace Microsoft.FeatureManagement
         Conflict,
 
         /// <summary>
-        /// The given value was invalid.
+        /// The given configuration setting was invalid.
         /// </summary>
-        InvalidValue
+        InvalidConfigurationSetting
     }
 }
--- a/src/Microsoft.FeatureManagement/FeatureManagementOptions.cs
+++ b/src/Microsoft.FeatureManagement/FeatureManagementOptions.cs
@@ -11,6 +11,7 @@ namespace Microsoft.FeatureManagement
         /// <summary>
         /// Controls the behavior of feature evaluation when dependent feature filters are missing.
         /// If missing feature filters are not ignored an exception will be thrown when attempting to evaluate a feature that depends on a missing feature filter.
+        /// This option will throw an error if used in combination with RequirementType.All as the execution may unintentionally enable a flag.
         /// The default value is false.
         /// </summary>
         public bool IgnoreMissingFeatureFilters { get; set; }
--- a/src/Microsoft.FeatureManagement/FeatureManagementOptions.cs
+++ b/src/Microsoft.FeatureManagement/FeatureManagementOptions.cs
@@ -11,7 +11,7 @@ namespace Microsoft.FeatureManagement
         /// <summary>
         /// Controls the behavior of feature evaluation when dependent feature filters are missing.
         /// If missing feature filters are not ignored an exception will be thrown when attempting to evaluate a feature that depends on a missing feature filter.
-        /// This option will throw an error if used in combination with RequirementType.All as the execution may unintentionally enable a flag.
+        /// This option is invalid when used in combination with <see cref="RequirementType.All"/>
         /// The default value is false.
         /// </summary>
         public bool IgnoreMissingFeatureFilters { get; set; }
--- a/src/Microsoft.FeatureManagement/FeatureManager.cs
+++ b/src/Microsoft.FeatureManagement/FeatureManager.cs
@@ -76,6 +76,12 @@ namespace Microsoft.FeatureManagement
 
             if (featureDefinition != null)
             {
+                if (featureDefinition.RequirementType == RequirementType.All && _options.IgnoreMissingFeatureFilters)
+                {
+                    string errorMessage = $"The 'IgnoreMissingFeatureFilters' flag cannot use used in combination with a feature of requirement type 'All'.";
+                    throw new FeatureManagementException(FeatureManagementError.MissingFeatureFilter, errorMessage);
+                }
+
                 //
                 // If the requirement type is Any, we end on a true. Requirement type All will end on a false
                 bool targetEvaluation = featureDefinition.RequirementType == RequirementType.Any;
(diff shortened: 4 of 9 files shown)


src/Microsoft.FeatureManagement/FeatureManager.cs
```diff
@@ -76,6 +76,13 @@ private async Task<bool> IsEnabledAsync<TContext>(string feature, TContext appCo
...
             if (featureDefinition != null)
             {
+                //
+                // Default to true for requirement type All
+                if (featureDefinition.RequirementType == RequirementType.All)
+                {
+                    enabled = true;
+                }
+
                 //
```
jimmyca15, line 86: The `Any` check in

`                if (featureDefinition.EnabledFor.Any(featureFilter => string.Equals(featureFilter.Name, "AlwaysOn", StringComparison.OrdinalIgnoreCase)))
`

Doesn't work for requirementtype.all

rossgrambo: Do you mean it shouldn't work, or in it's current state, it doesn't work?

src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
```diff
@@ -149,6 +156,16 @@ We support
             }
             else
             {
+                string rawRequirementType = configurationSection[RequirementTypeKeyword];
+
+                if (!string.IsNullOrEmpty(rawRequirementType))
+                {
+                    if (!Enum.TryParse(rawRequirementType, out requirementType))
```
avanigupta, line 163: This needs to be case insensitive.

src/Microsoft.FeatureManagement/FeatureManager.cs
jimmyca15: I guess both. Right now the presence of the AlwaysOn filter will short circuit the evaluation and enable the feature. Even for `all`

src/Microsoft.FeatureManagement/FeatureManager.cs
```diff
@@ -76,6 +76,13 @@ private async Task<bool> IsEnabledAsync<TContext>(string feature, TContext appCo
 
             if (featureDefinition != null)
             {
+                //
+                // Default to true for requirement type All
+                if (featureDefinition.RequirementType == RequirementType.All)
+                {
+                    enabled = true;
```
avanigupta, line 83: What do we want the behavior to be when `IgnoreMissingFeatureFilters` is true and no feature filter is found?
Currently, we treat this as a disabled feature since RequirementType is Any. But when RequirementType is All, should that be an enabled feature or disabled?

avanigupta: I think it should be disabled. It should have the same effect as this configuration:
```json
"FeatureManagement": {
 "DisabledFeature": {
            "EnabledFor": []
        }
}
```
In that case, the code needs modification to treat a feature with no filters as disabled feature.

rossgrambo: I was also curious about this. In the current state, it would be evaluated true. 

I think that would be unexpected, but it felt messier to have the AND logic be false- until one filter evaluates to true. Do we know what the use case for using IgnoreMissingFeatureFilters is?

rossgrambo: Okay! That makes sense, I'll adjust it.

src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
```diff
@@ -149,6 +156,16 @@ We support
...
+                    {
+                        throw new FormatException($"Invalid requirement type '{rawRequirementType}' for feature '{configurationSection.Key}'.");
```
jimmyca15, line 165: Do we use FormatException elsewhere? I'd like if we could use `FeatureManagementException`.

src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
jimmyca15: What would be your reasoning that it would need to be case insensitive?

avanigupta: We have case insensitive comparison while parsing other feature flag fields, like feature name, filter name, etc. Just want to maintain the same behavior for requirement type.
I imagine people might try to use RequirementType "ALL", "All" or "all" interchangeably.

jimmyca15: Yes, I think it would stick out if this was case sensitive given the other examples you mentioned.

src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
rossgrambo: Good call, I shouldn't introduce a new exception type. I switched it to `ArgumentException` instead.

Looking through our exceptions, all we have is-
```
FeatureManagementException
    FeatureManagementError.MissingFeatureFilter
    FeatureManagementError.AmbiguousFeatureFilter
    FeatureManagementError.MissingFeature
ArgumentNullException
ArgumentException
```

So I think ArgumentException makes the most sense. It's currently used by ValidateSettings in TargetingFilter.

A lot more error types get added by the v3 changes. Perhaps once those are back in, we can use the InvalidConfiguration error type. Or I can add it now if we want.

src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
```diff
@@ -149,6 +156,16 @@ We support
             }
             else
             {
+                string rawRequirementType = configurationSection[RequirementTypeKeyword];
+
+                if (!string.IsNullOrEmpty(rawRequirementType))
+                {
+                    if (!Enum.TryParse(rawRequirementType, true, out requirementType))
+                    {
+                        throw new ArgumentException($"Invalid requirement type '{rawRequirementType}' for feature '{configurationSection.Key}'.");
```
avanigupta, line 165: Technically, rawRequirementType is not an argument so we shouldn't throw ArgumentException.
I think it should be okay to add a new error type for `InvalidConfiguration` in FeatureManagementException.

rossgrambo:
I addressed the comments, but I also re-wrote my changes to the FeatureManager to have less edge-cases. 

Both Any and And now have the same default: false 
Which allows me to remove the edge case for an empty list of filters with a default of true.
The default changes to "true" when the first filter returns true. In the Any case, we stop and return. In the All case, we now have a true default and we continue.

I also combined some of the inner logic so the Contextual and non-Contextual filters have less duplicated code.

src/Microsoft.FeatureManagement/FeatureManager.cs
```diff
@@ -77,68 +77,93 @@ private async Task<bool> IsEnabledAsync<TContext>(string feature, TContext appCo
...
-                            string errorMessage = $"The feature filter '{featureFilterConfiguration.Name}' specified for feature '{feature}' was not found.";
-
-                            if (!_options.IgnoreMissingFeatureFilters)
+                            throw new FeatureManagementException(FeatureManagementError.MissingFeatureFilter, errorMessage);
+                        }
+                        else
+                        {
+                            if (featureDefinition.RequirementType == RequirementType.All)
                             {
+                                errorMessage = $"The 'IgnoreMissingFeatureFilters' flag cannot override a missing filter on a feature with requirement type 'All'. {errorMessage}";
```
avanigupta, line 110: I think we should block this combination even if filters are not missing. It produces a more deterministic restriction which is not dependent on users' feature filters, which may change in future.

src/Microsoft.FeatureManagement/FeatureManager.cs
```diff
@@ -77,68 +77,93 @@ private async Task<bool> IsEnabledAsync<TContext>(string feature, TContext appCo
...
 
-                                break;
-                            }
+                    //
+                    // IContextualFeatureFilter
+                    if (useAppContext)
+                    {
+                        ContextualFeatureFilterEvaluator contextualFilter = GetContextualFeatureFilter(featureFilterConfiguration.Name, typeof(TContext));
+
+                        if (contextualFilter != null) {
```
avanigupta, line 137: nit: opening bracket in new line

src/Microsoft.FeatureManagement/FeatureManager.cs
```diff
@@ -77,68 +77,93 @@ private async Task<bool> IsEnabledAsync<TContext>(string feature, TContext appCo
...
-                else
+                //
+                // For all enabling filters listed in the feature's state, evaluate them according to requirement type
+                foreach (FeatureFilterConfiguration featureFilterConfiguration in featureDefinition.EnabledFor)
                 {
                     //
-                    // For all enabling filters listed in the feature's state calculate if they return true
-                    // If any executed filters return true, return true
+                    // If the filter is AlwaysOn and we're targeting true, return true
+                    if (string.Equals(featureFilterConfiguration.Name, "AlwaysOn", StringComparison.OrdinalIgnoreCase) && targetEvaluation)
```
avanigupta, line 89: AlwaysOn filter should always evaluate to true. If the filter is AlwaysOn and requirement is All, we might fail to find IFeatureFilterMetadata and fail the evaluation.

src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
```diff
@@ -149,6 +156,16 @@ We support
             }
             else
             {
+                string rawRequirementType = configurationSection[RequirementTypeKeyword];
+
+                if (!string.IsNullOrEmpty(rawRequirementType))
+                {
+                    if (!Enum.TryParse(rawRequirementType, true, out requirementType))
```
jimmyca15, line 163: Those `if` statement's can be combined.

src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
```diff
@@ -149,6 +156,16 @@ We support
...
+                    {
+                        throw new FeatureManagementException(FeatureManagementError.InvalidConfiguration, $"Invalid requirement type '{rawRequirementType}' for feature '{configurationSection.Key}'.");
```
jimmyca15, line 165: Can you new-line the params. The line got kind of long.

src/Microsoft.FeatureManagement/FeatureManagementOptions.cs
```diff
@@ -11,6 +11,7 @@ public class FeatureManagementOptions
         /// <summary>
         /// Controls the behavior of feature evaluation when dependent feature filters are missing.
         /// If missing feature filters are not ignored an exception will be thrown when attempting to evaluate a feature that depends on a missing feature filter.
+        /// This option will throw an error if used in combination with RequirementType.All as the execution may unintentionally enable a flag.
```
jimmyca15, line 14: "This option will throw an error" It's not that the option throws. The option is invalid when combined with `all`.

jimmyca15: You should be able to use `cref` I believe to link to the `RequirementType.All` type.

src/Microsoft.FeatureManagement/FeatureManagementError.cs
```diff
@@ -9,18 +9,23 @@ namespace Microsoft.FeatureManagement
...
         /// </summary>
-        MissingFeatureFilter,
+        AmbiguousFeatureFilter,
 
         /// <summary>
-        /// A feature filter configured for the feature being evaluated is an ambiguous reference to multiple registered feature filters.
+        /// A configuration error is present in the feature management system.
         /// </summary>
-        AmbiguousFeatureFilter,
+        InvalidConfiguration,
```
jimmyca15, line 19: It seems like every error here could be classified as an `InvalidConfiguration`. In a way it seems like a catch all error-code.

Thinking -- how can we be more specific with the error code to the problem we're adding it for.

src/Microsoft.FeatureManagement/FeatureManagementError.cs
```diff
@@ -9,18 +9,23 @@ namespace Microsoft.FeatureManagement
     public enum FeatureManagementError
     {
         /// <summary>
-        /// A feature filter that was listed for feature evaluation was not found.
+        /// A feature filter configured for the feature being evaluated is an ambiguous reference to multiple registered feature filters.
         /// </summary>
-        MissingFeatureFilter,
+        AmbiguousFeatureFilter,
```
jimmyca15, line 14: Be careful about changing the order of enums. They have underlying numeric values. If someone was serializing these errors as number, this would break.

src/Microsoft.FeatureManagement/FeatureManager.cs
```diff
@@ -76,69 +76,100 @@ private async Task<bool> IsEnabledAsync<TContext>(string feature, TContext appCo
...
             {
-                //
-                // Check if feature is always on
-                // If it is, result is true, goto: cache
-
-                if (featureDefinition.EnabledFor.Any(featureFilter => string.Equals(featureFilter.Name, "AlwaysOn", StringComparison.OrdinalIgnoreCase)))
+                if (featureDefinition.RequirementType == RequirementType.All && _options.IgnoreMissingFeatureFilters)
                 {
-                    enabled = true;
+                    string errorMessage = $"The 'IgnoreMissingFeatureFilters' flag cannot use used in combination with a feature of requirement type 'All'.";
```
jimmyca15, line 81: No need for the variable. It's only used in a single place.

src/Microsoft.FeatureManagement/FeatureManager.cs
```diff
@@ -76,69 +76,100 @@ private async Task<bool> IsEnabledAsync<TContext>(string feature, TContext appCo
...
                     //
-                    // For all enabling filters listed in the feature's state calculate if they return true
-                    // If any executed filters return true, return true
-
-                    foreach (FeatureFilterConfiguration featureFilterConfiguration in featureDefinition.EnabledFor)
+                    // Handle AlwaysOn filters
+                    if (string.Equals(featureFilterConfiguration.Name, "AlwaysOn", StringComparison.OrdinalIgnoreCase))
                     {
-                        IFeatureFilterMetadata filter = GetFeatureFilterMetadata(featureFilterConfiguration.Name);
+                        enabled = true;
```
jimmyca15, line 97: `enabled = true;`

I believe having to do this comes because we start with false because of the empty feature filter case? If so, how about we special case it so we can avoid it? Something like

```
if (!featureDefinition.EnabledFor.Any())
{

}
else
{
    enabled = featureDefinition.RequirementType == RequirementType.All;
    
     foreach (...)
     {
        ...
     }
}
```

src/Microsoft.FeatureManagement/FeatureManager.cs
```diff
@@ -76,69 +76,100 @@ private async Task<bool> IsEnabledAsync<TContext>(string feature, TContext appCo
...
 
-                        if (filter == null)
+                        if (enabled == targetEvaluation)
+                        {
+                            break;
+                        } else
```
jimmyca15, line 102: else should be on a new line.

src/Microsoft.FeatureManagement/FeatureManager.cs
rossgrambo: In this particular spot, changing the logic would not affect this code much, if at all, as we still need to special case `AlwaysOn` as an expected filter that will not be found as a IFeatureFilter. `AlwaysOn` will always need to `return true` in the `Any` case and `continue` in the `All` case because continuing the code would error. 

If you check my first commit, I originally had that pattern, but it adds more `if's` and overall reads less cleanly-
1. It adds an `if` at the beginning where we default to either true or false depending on requirement type.
2. It adds an `if` for the empty list

The only benefit would only remove the `// Set enabled to true if this filter evaluated to true` section

src/Microsoft.FeatureManagement/FeatureManager.cs
```diff
@@ -76,69 +76,100 @@ private async Task<bool> IsEnabledAsync<TContext>(string feature, TContext appCo
...
+                    if (filter == null)
+                    {
+                        string errorMessage = $"The feature filter '{featureFilterConfiguration.Name}' specified for feature '{feature}' was not found.";
+
+                        if (!_options.IgnoreMissingFeatureFilters)
                         {
-                            ContextualFeatureFilterEvaluator contextualFilter = GetContextualFeatureFilter(featureFilterConfiguration.Name, typeof(TContext));
+                            throw new FeatureManagementException(FeatureManagementError.MissingFeatureFilter, errorMessage);
+                        }
+                        else
```
avanigupta, line 118: `else` is not required

src/Microsoft.FeatureManagement/FeatureManager.cs
jimmyca15: Not just that. You don't need the `evaluation` variable, and the implementation could really be a short circuit.

src/Microsoft.FeatureManagement/FeatureManager.cs
```diff
@@ -122,37 +144,40 @@ private async Task<bool> IsEnabledAsync<TContext>(string feature, TContext appCo
...
-                        if (filter is IFeatureFilter featureFilter && await featureFilter.EvaluateAsync(context).ConfigureAwait(false))
+                        if (filter is IFeatureFilter featureFilter && 
+                            await featureFilter.EvaluateAsync(context).ConfigureAwait(false) == targetEvaluation)
                         {
-                            enabled = true;
+                            enabled = targetEvaluation;
 
                             break;
-                        }
+                        };
```
jimmyca15, line 164: Stray colon

src/Microsoft.FeatureManagement/FeatureManager.cs
rossgrambo: Synced with @jimmyca15. Reverted my "default to false" changes introduced in https://github.com/microsoft/FeatureManagement-Dotnet/pull/221/commits/0ef1d89b62224b0e6c660a4c42b7c9e40c230f81 and made some readability updates.

src/Microsoft.FeatureManagement/FeatureManagementError.cs
rossgrambo: Undid the re-ordering

src/Microsoft.FeatureManagement/FeatureManagementError.cs
rossgrambo: Removed Invalid Configuration and added 

```
        /// <summary>
        /// A configuration had conflicting values.
        /// </summary>
        Conflict,

        /// <summary>
        /// The given value was invalid.
        /// </summary>
        InvalidValue
```

src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
```diff
@@ -149,6 +156,17 @@ We support
             }
             else
             {
+                string rawRequirementType = configurationSection[RequirementTypeKeyword];
+
+                //
+                // If requirement type is specified, parse it and set the requirementType variable
+                if (!string.IsNullOrEmpty(rawRequirementType) && !Enum.TryParse(rawRequirementType, true, out requirementType))
```
jimmyca15, line 163: `Enum.TryParse(rawRequirementType, true, out requirementType)`,

For boolean parameters where we pass a boolean literal, I'd recommend to use parameter name to help the reader know what that flag is for.

Like `Enum.TryParse(rawRequirementType, ignoreCase: true, out requirementType)`,

src/Microsoft.FeatureManagement/FeatureManagementError.cs
```diff
@@ -21,6 +21,16 @@ public enum FeatureManagementError
         /// <summary>
         /// A feature that was requested for evaluation was not found.
         /// </summary>
-        MissingFeature
+        MissingFeature,
+
+        /// <summary>
+        /// A configuration had conflicting values.
```
jimmyca15, line 27: The error code isn't limited to configuration. It could be any conflict we encounter in our feature management system.

src/Microsoft.FeatureManagement/FeatureManager.cs
```diff
@@ -70,28 +70,52 @@ private async Task<bool> IsEnabledAsync<TContext>(string feature, TContext appCo
...
+                {
+                    throw new FeatureManagementException(
+                        FeatureManagementError.Conflict, 
+                        $"The 'IgnoreMissingFeatureFilters' flag cannot use used in combination with a feature of requirement type 'All'.");
+                }
 
-                if (featureDefinition.EnabledFor.Any(featureFilter => string.Equals(featureFilter.Name, "AlwaysOn", StringComparison.OrdinalIgnoreCase)))
+                //
+                // Treat an empty list of enabled filters as a disabled feature
+                if (featureDefinition.EnabledFor == null || featureDefinition.EnabledFor.Count() == 0)
```
jimmyca15, line 88: `featureDefinition.EnabledFor.Count() == 0` I'd recommend `featureDefinition.EnabledFor.Any()`. Otherwise you may need to enumerate an entire enumerable to get the entire count.

src/Microsoft.FeatureManagement/FeatureManager.cs
```diff
@@ -122,37 +144,40 @@ private async Task<bool> IsEnabledAsync<TContext>(string feature, TContext appCo
...
-                            if (contextualFilter != null && await contextualFilter.EvaluateAsync(context, appContext).ConfigureAwait(false))
+                            if (contextualFilter != null && 
+                                await contextualFilter.EvaluateAsync(context, appContext).ConfigureAwait(false) == targetEvaluation)
                             {
-                                enabled = true;
+                                enabled = targetEvaluation;
 
                                 break;
-                            }
+                            };
```
jimmyca15, line 153: There's a stray semi colon here.

src/Microsoft.FeatureManagement/FeatureManager.cs
```diff
@@ -122,37 +144,40 @@ private async Task<bool> IsEnabledAsync<TContext>(string feature, TContext appCo
...
+                        if (filter is IFeatureFilter featureFilter && 
+                            await featureFilter.EvaluateAsync(context).ConfigureAwait(false) == targetEvaluation)
                         {
-                            enabled = true;
+                            enabled = targetEvaluation;
 
                             break;
                         }
                     }
+
```
jimmyca15, line 166: Accidental whitespace looks like.

src/Microsoft.FeatureManagement/FeatureManagementError.cs
jimmyca15: This was resolved, but I don't see any reply or related change.

rossgrambo: My fault, didn't save it in visual studio so it wasn't included in the commit. Added it now. I went with: `There was a conflict in the feature management system.`

jimmyca15 (approved):

avanigupta (approved):

PR #226, opened 2023-04-13
Update README with targeting exclusion and requirement type snippets.

rossgrambo:
Adds sections to the README for 

1. Targeting Exclusion
2. RequirementType `All` vs `Any`

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,7 +24,10 @@ Here are some of the benefits of using this library:
 Feature flags are composed of two parts, a name and a list of feature-filters that are used to turn the feature on.
 
 ### Feature Filters
-Feature filters define a scenario for when a feature should be enabled. When a feature is evaluated for whether it is on or off, its list of feature-filters are traversed until one of the filters decides the feature should be enabled. At this point the feature is considered enabled and traversal through the feature filters stops. If no feature filter indicates that the feature should be enabled, then it will be considered disabled.
+Feature filters define a scenario for when a feature should be enabled. When a feature is evaluated for whether it is on or off, its list of feature-filters are traversed until one of the filters decides the feature should be enabled or disabled. Traversal works in one of two ways:
+
+1. For filters with a `RequirementType` of `Any` (the default): The feature-filters are traversed until one of the filters returns true. If no filter returns true then the feature is disabled.
+2. For filters with a `RequirementType` of `All`: The feature-filters are traversed until one of the filters returns false. If no feature filter returns false and there is at least 1 filter - then the feature is enabled.
 
 As an example, a Microsoft Edge browser feature filter could be designed. This feature filter would activate any features it is attached to as long as an HTTP request is coming from Microsoft Edge.
 
@@ -67,11 +70,29 @@ The feature management library supports appsettings.json as a feature flag sourc
                 }
             ]
         }
+        "FeatureW": {
+            "RequirementType": "All",
+            "EnabledFor": [
+                {
+                    "Name": "TimeWindow",
+                    "Parameters": {
+                        "Start": "Wed, 01 May 2019 13:59:59 GMT",
+                        "End": "Mon, 01 July 2019 00:00:00 GMT"
+                    }
+                },
+                {
+                    "Name": "Percentage",
+                    "Parameters": {
+                        "Value": "50"
+                    }
+                }
+            ]
+        }
     }
 }
 ```
 
-The `FeatureManagement` section of the json document is used by convention to load feature flag settings. In the section above, we see that we have provided three different features. Features define their feature filters using the `EnabledFor` property. In the feature filters for `FeatureT` we see `AlwaysOn`. This feature filter is built-in and if specified will always enable the feature. The `AlwaysOn` feature filter does not require any configuration so it only has the _Name_ property. `FeatureU` has no filters in its `EnabledFor` property and thus will never be enabled. Any functionality that relies on this feature being enabled will not be accessible as long as the feature filters remain empty. However, as soon as a feature filter is added that enables the feature it can begin working. `FeatureV` specifies a feature filter named `TimeWindow`. This is an example of a configurable feature filter. We can see in the example that the filter has a parameter's property. This is used to configure the filter. In this case, the start and end times for the feature to be active are configured.
+The `FeatureManagement` section of the json document is used by convention to load feature flag settings. In the section above, we see that we have provided four different features. Features define their feature filters using the `EnabledFor` property. In the feature filters for `FeatureT` we see `AlwaysOn`. This feature filter is built-in and if specified will always enable the feature. The `AlwaysOn` feature filter does not require any configuration so it only has the _Name_ property. `FeatureU` has no filters in its `EnabledFor` property and thus will never be enabled. Any functionality that relies on this feature being enabled will not be accessible as long as the feature filters remain empty. However, as soon as a feature filter is added that enables the feature it can begin working. `FeatureV` specifies a feature filter named `TimeWindow`. This is an example of a configurable feature filter. We can see in the example that the filter has a parameter's property. This is used to configure the filter. In this case, the start and end times for the feature to be active are configured. `FeatureW` specifies a requirement type of `All`, meaning all of it's filters must return true for the feature to be active. In this case, the feature is active between the specified time window and for 50% of users.
 
 ### On/Off Declaration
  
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -24,7 +24,10 @@ Here are some of the benefits of using this library:
 Feature flags are composed of two parts, a name and a list of feature-filters that are used to turn the feature on.
 
 ### Feature Filters
-Feature filters define a scenario for when a feature should be enabled. When a feature is evaluated for whether it is on or off, its list of feature-filters are traversed until one of the filters decides the feature should be enabled. At this point the feature is considered enabled and traversal through the feature filters stops. If no feature filter indicates that the feature should be enabled, then it will be considered disabled.
+Feature filters define a scenario for when a feature should be enabled. When a feature is evaluated for whether it is on or off, its list of feature-filters are traversed until one of the filters decides the feature should be enabled or disabled. Traversal works in one of two ways:
+
+1. For filters with a `RequirementType` of `Any` (the default): The feature-filters are traversed until one of the filters returns true. If no filter returns true then the feature is disabled.
```
avanigupta, line 29: > until one of the filters returns true

Instead of saying that it "returns true", can we say "is enabled"?
Returning a particular value is implementation detail.

So the sentence would be something like:
"Feature filters are traversed until one of the filters decides that the feature should be enabled. If no filter indicates that the feature should be enabled, then it will be considered disabled."

For RequirementType All:
"Feature filters are traversed until one of the filters decides that the feature should be disabled. If no filter indicates that the feature should be disabled, then it will be considered enabled."

I dont think we need to mention that there should be at least one filter to enable the feature. That information is already communicated elsewhere in the doc and its not specific to Requirement Type.

README.md
```diff
@@ -67,11 +70,29 @@ The feature management library supports appsettings.json as a feature flag sourc
...
...on above, we see that we have provided four different features. Features define their feature filters using the `EnabledFor` property. In the feature filters for `FeatureT` we see `AlwaysOn`. This feature filter is built-in and if specified will always enable the feature. The `AlwaysOn` feature filter does not require any configuration so it only has the _Name_ property. `FeatureU` has no filters in its `EnabledFor` property and thus will never be enabled. Any functionality that relies on this feature being enabled will not be accessible as long as the feature filters remain empty. However, as soon as a feature filter is added that enables the feature it can begin working. `FeatureV` specifies a feature filter named `TimeWindow`. This is an example of a configurable feature filter. We can see in the example that the filter has a parameter's property. This is used to configure the filter. In this case, the start and end times for the feature to be active are configured. `FeatureW` specifies a requirement type of `All`, meaning all of it's filters must return true for the feature to be active. In this case, the feature is active between the specified time window and for 50% of users.
```
avanigupta, line 95: > We can see in the example that the filter has a parameter's property

Parameters (no apostrophe) 

Also, can you put Name and Parameters in code quotes to match other property names?

avanigupta, line 95: Same comment here about "returning true" in FeatureW explanation - we can say "enabled" instead.
And same comment about`RequirementType` in quotes.

README.md
```diff
@@ -440,7 +461,7 @@ This filter provides the capability to enable a feature based on a time window.
...
+This filter provides the capability to enable a feature for a target audience. An in-depth explanation of targeting is explained in the [targeting](./README.md#Targeting) section below. The filter parameters include an audience object which describes users, groups, excluded users/groups, and a default percentage of the user base that should have access to the feature. Each group object that is listed in the target audience must also specify what percentage of the group's members should have access. If a user is specified in the exclusion object, either directly or in a group, the feature will be disabled. Otherwise, if a user is specified in the users section directly, or if the user is in the included percentage of any of the group rollouts, or if the user falls into the default rollout percentage then that user will have the feature enabled.
```
avanigupta, line 464: > If a user is specified in the exclusion object, either directly or in a group, the feature will be disabled.

 If a user is specified in the exclusion section, either directly or if the user is in an excluded group, the feature will be disabled.

README.md
```diff
@@ -463,21 +484,36 @@ This filter provides the capability to enable a feature for a target audience. A
...
+                        ]
+                    }
                 }
             }
         }
     ]
 }
 ```
 
+### RequirementType
```
avanigupta, line 503: This section is talking mostly about feature filters. I think RequirementType can be moved to the top, closer to Feature Flag Declaration section.

README.md
```diff
@@ -551,6 +587,30 @@ services.Configure<TargetingEvaluationOptions>(options =>
...
+        }
+    ],
+    "DefaultRolloutPercentage": 0
+    "Exclusion": {
+        "Users": [
+            "Mark"
+        ]
+    }
+}
+```
```
avanigupta, line 612: Can you add a short example of what the filter will evaluate to for different users?

README.md
rossgrambo: Looks like some spelling mistakes were already there, I might do a full spell check pass in a different PR.

README.md
```diff
@@ -67,11 +70,29 @@ The feature management library supports appsettings.json as a feature flag sourc
                 }
             ]
         }
+        "FeatureW": {
```
jimmyca15, line 73: If we keep adding to this list it will get a bit unweildy. I'd like to see if we can limit the new features to their own sections. In this case, moving this feature you've put together to the requirement type section.

README.md
```diff
@@ -24,7 +24,10 @@ Here are some of the benefits of using this library:
 Feature flags are composed of two parts, a name and a list of feature-filters that are used to turn the feature on.
 
 ### Feature Filters
-Feature filters define a scenario for when a feature should be enabled. When a feature is evaluated for whether it is on or off, its list of feature-filters are traversed until one of the filters decides the feature should be enabled. At this point the feature is considered enabled and traversal through the feature filters stops. If no feature filter indicates that the feature should be enabled, then it will be considered disabled.
+Feature filters define a scenario for when a feature should be enabled. When a feature is evaluated for whether it is on or off, its list of feature-filters are traversed until one of the filters decides the feature should be enabled or disabled. Traversal works in one of two ways:
```
jimmyca15, line 27: Inline with my comment about add-ons in new sections. I wanted to propose keeping the wording the same for this section but just mentioning that it can be overriden with requirement type and linking to the new requirement type section, also moving the new content here into the requirement type section.

README.md
```diff
@@ -24,7 +24,10 @@ Here are some of the benefits of using this library:
...
+Feature filters define a scenario for when a feature should be enabled. When a feature is evaluated for whether it is on or off, its list of feature-filters are traversed until one of the filters decides the feature should be enabled or disabled. Traversal works in one of two ways:
+
+1. For features with a `RequirementType` of `Any` (the default): The feature-filters are traversed until one of the filters decides that the feature should be enabled. If no filter indicates that the feature should be enabled, then it will be considered disabled.
+2. For features with a `RequirementType` of `All`: The feature-filters are traversed until one of the filters decides that the feature should be disabled. If no filter indicates that the feature should be disabled, then it will be considered enabled.
```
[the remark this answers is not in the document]
rossgrambo: Haha, I previously had noted that, but Avani called out that we already address that elsewhere. https://github.com/microsoft/FeatureManagement-Dotnet/pull/226#discussion_r1166119204

README.md
rossgrambo: I feel like we can take one of two approaches-

1. Describe filters in the Any state and then treat RequirementType "All" as an override.
2. Describe filters as having either All or Any logic and describing each option.

When I was writing this- I was using option **2**. Sounds like you're suggesting option **1**. I think this goes along with your other suggestions as well- for moving the RequirementType related info to the RequirementType section. 

I think you're right and we should describe it as an override for now, as it's easier to understand sequentially. I'll re-write it towards that.

README.md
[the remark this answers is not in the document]
jimmyca15: Where is that other place it is addressed?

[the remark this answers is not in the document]
rossgrambo: https://github.com/microsoft/FeatureManagement-Dotnet#feature-filters

> If no feature filter indicates that the feature should be enabled, then it will be considered disabled.

And https://github.com/microsoft/FeatureManagement-Dotnet#feature-flag-declaration

> FeatureU has no filters in its EnabledFor property and thus will never be enabled.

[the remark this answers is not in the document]
jimmyca15: I still worry. It's not quite clear that the same is true for the all requirement. If one were in doubt, and they read this description of how all works, then I have a strong feeling that they would be led astray.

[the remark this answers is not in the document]
avanigupta: I'm okay with adding it here explicitly.

jimmyca15 (approved):

PR #228, opened 2023-04-20
Adds a simplified index

rossgrambo:


README.md
```diff
@@ -20,6 +20,16 @@ Here are some of the benefits of using this library:
...
 
+## Index
+* [Feature Flags](#feature-flags)
+* [Feature Filters](#feature-filters)
+* [Registration](#registration)
+* [Consumption](#consumption)
+* [Targeting](#targeting)
+* [Caching](#caching)
+* [Custom Feature Providers](#custom-feature-providers)
+* [Contributing](#contributing)
```
jimmyca15, line 31: I would recommend leaving off this piece. It's not really relevant to referencing feature management functionality.

README.md
```diff
@@ -20,6 +20,16 @@ Here are some of the benefits of using this library:
 
 **API Reference**: https://go.microsoft.com/fwlink/?linkid=2091700
 
+## Index
+* [Feature Flags](#feature-flags)
+* [Feature Filters](#feature-filters)
+* [Registration](#registration)
```
jimmyca15, line 26: This is a bit odd to see. I didn't realize this was a top level header. Makes me wonder when I read, am I registering for a subscription of some sort?

I realized that we did a bit of restructuring of the top level headers in the same update that we added the index which was retracted. See how in [this branch](https://github.com/microsoft/FeatureManagement-Dotnet/blob/release/v3/README.md#Feature-Flags), the registration top level header is gone. I think we should take that same change. The resulting index in that branch is a bit more informative as a result.

rossgrambo: Yeah I agree. I wasn't sure how much restructuring we wanted to do with this- in case the 3.0.0 changes were still on the horizon for restructuring. I can work on cleaning it up though.

jimmyca15: If possible to take all of the changes I'd go for it. It was unfortunate they were reverted as a part of the 3.0 revert.

jimmyca15 (approved):

PR #180, opened 2022-05-23
Added capability to cache filter and assignment parameters.

jimmyca15:
This PR adds a filter parameters binder interface to allow feature manager to cache bound feature filter and feature variant assignment parameters for performance optimization.

### Why

Feature filters are given filter parameters in the form of `IConfiguration` that instruct them how to perform feature evaluation. Often these parameters are bound to a strongly typed settings object before use. Observed cost of binding `IConfiguration` to perform feature evaluation is high. By caching the result of the binding, a noticeable performance increase can be achieved. This PR gives Implementers of IFeatureFilter and IFeatureVariantAssigner the option to take cache these settings with the help of the feature management system. This is possible through the `IFilterParametersBinder` and `IAssignmentParametersBinder` interfaces.

``` c#
/// <summary>
/// An interface used by the feature management system to pre-bind feature filter parameters to a settings type.
/// <see cref="IFeatureFilter"/>s can implement this interface to take advantage of caching of settings by the feature management system.
/// </summary>
public interface IFilterParametersBinder
{
    /// <summary>
    /// Binds a set of feature filter parameters to a settings object.
    /// </summary>
    /// <param name="parameters">The configuration representing filter parameters to bind to a settings object</param>
    /// <returns>A settings object that is understood by the implementer of <see cref="IFilterParametersBinder"/>.</returns>
    object BindParameters(IConfiguration parameters);
}
```

### Example Usage

``` c#
class MyFilter : IFeatureFilter, IFilterParametersBinder
{
    public object BindParameters(IConfiguration filterParameters)
    {
        //
        // Called before evaluation occurs, only if settings are not already in cache.
        return filterParameters.Get<FilterSettings>();
    }

    public Task<bool> EvaluateAsync(FeatureFilterEvaluationContext context, CancellationToken cancellationToken)
    {
        //
        // Get pre-bound/cached settings
        FilterSettings settings = (FilterSettings)context.Settings;

        //
        // Use them for evaluation
        ...
    }
}
```

### Breaking Change

This PR introduces a possible breaking change for custom feature definition providers that implement `IFeatureFlagDefinitionProvider` or `IDynamicFeatureDefinitionProvider`. To ensure that any filters that implement `IFilterParametersBinder` see updates to feature filter parameters or feature variant assignment parameters when they are changed, definition providers should ensure a new object reference is used for the parameters in the feature definition. Otherwise, the filters will continue to use the same settings from the cache.

### Perf

Rough performance assessment showed improvements of ~100 times at least for complex settings objects.

PR #229, opened 2023-04-24
Adds 'Settings' to Context and uses it as a cache of parameters

rossgrambo:
Rewrite of https://github.com/microsoft/FeatureManagement-Dotnet/pull/180

This PR adds a filter parameters binder interface to allow feature manager to cache bound feature filter and feature variant assignment parameters for performance optimization.

### Why

Feature filters are given filter parameters in the form of `IConfiguration` that instruct them how to perform feature evaluation. Often these parameters are bound to a strongly typed settings object before use. Observed cost of binding `IConfiguration` to perform feature evaluation is high. By caching the result of the binding, a noticeable performance increase can be achieved. This PR gives Implementers of IFeatureFilter and IFeatureVariantAssigner the option to cache these settings with the help of the feature management system. This is possible through the `IFilterParametersBinder` and `IAssignmentParametersBinder` interfaces.

``` c#
/// <summary>
/// An interface used by the feature management system to pre-bind feature filter parameters to a settings type.
/// <see cref="IFeatureFilter"/>s can implement this interface to take advantage of caching of settings by the feature management system.
/// </summary>
public interface IFilterParametersBinder
{
    /// <summary>
    /// Binds a set of feature filter parameters to a settings object.
    /// </summary>
    /// <param name="parameters">The configuration representing filter parameters to bind to a settings object</param>
    /// <returns>A settings object that is understood by the implementer of <see cref="IFilterParametersBinder"/>.</returns>
    object BindParameters(IConfiguration parameters);
}
```

### Example Usage

Filters offered in this library will benefit from this caching change for anyone using this version moving forward. For custom filters, the class must implement the IFilterParametersBinder interface. Below is an example.

``` c#
class MyFilter : IFeatureFilter, IFilterParametersBinder
{
    {
        //
        // Called before evaluation occurs, only if settings are not already in cache.
        return filterParameters.Get<FilterSettings>();
    }

    public Task<bool> EvaluateAsync(FeatureFilterEvaluationContext context, CancellationToken cancellationToken)
    {
        //
        // Get pre-bound/cached settings
        FilterSettings settings = (FilterSettings)context.Settings;

        //
        // Use them for evaluation
        ...
    }
}
```

### Breaking Change

This PR introduces a possible breaking change for custom feature definition providers that implement `IFeatureFlagDefinitionProvider`. To ensure that any filters that implement `IFilterParametersBinder` see updates to feature filter parameters when they are changed, definition providers should ensure a new object reference is used for the parameters in the feature definition. Otherwise, the filters will continue to use the same settings from the cache.

### Perf

Rough performance assessment showed improvements of ~100 times at least for complex settings objects.


src/Microsoft.FeatureManagement/FeatureFilters/IFilterParametersBinder.cs
```diff
@@ -0,0 +1,21 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+//
+using Microsoft.Extensions.Configuration;
+
+namespace Microsoft.FeatureManagement
```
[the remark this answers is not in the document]
jimmyca15: Agree. Not in a sense that namespace has to correlate with folder structure, but its that the intent of this folder was implementations of feature filters. This type should be next to IFeatureFilter which it's meant to decorate.

src/Microsoft.FeatureManagement/FeatureManager.cs
```diff
@@ -57,6 +74,10 @@ await foreach (FeatureDefinition featureDefintion in _featureDefinitionProvider.
                 yield return featureDefintion.Name;
             }
         }
```
jimmyca15, line 76: missing a new line between methods.

src/Microsoft.FeatureManagement/FeatureManager.cs
```diff
@@ -14,15 +16,24 @@ namespace Microsoft.FeatureManagement
     /// <summary>
     /// Used to evaluate whether a feature is enabled or disabled.
     /// </summary>
-    class FeatureManager : IFeatureManager
+    class FeatureManager : IFeatureManager, IDisposable
     {
+        private readonly TimeSpan SettingsCachePeriod = TimeSpan.FromSeconds(5);
```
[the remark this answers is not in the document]
jimmyca15: Makes sense. Users could set it to zero to turn off the caching altogether.

src/Microsoft.FeatureManagement/FeatureManager.cs
```diff
@@ -38,6 +49,12 @@ class FeatureManager : IFeatureManager
             _filterMetadataCache = new ConcurrentDictionary<string, IFeatureFilterMetadata>();
             _contextualFeatureFilterCache = new ConcurrentDictionary<string, ContextualFeatureFilterEvaluator>();
             _options = options?.Value ?? throw new ArgumentNullException(nameof(options));
+            _cache = new MemoryCache(
+                Options.Create(
```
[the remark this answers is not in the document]
jimmyca15: I think I wrote that line originally. If there is an overload that doesn't require IOptions then it was an oversight on my part.

src/Microsoft.FeatureManagement/FeatureManager.cs
```diff
@@ -187,6 +213,46 @@ await featureFilter.EvaluateAsync(context).ConfigureAwait(false) == targetEvalua
...
+            }
+
+            object settings;
+
+            //
+            // Check if settings already bound from configuration
+            ConfigurationCacheItem cacheItem = (ConfigurationCacheItem)_cache.Get(context.FeatureName);
+
+            if (cacheItem == null ||
+                cacheItem.Parameters != context.Parameters)
```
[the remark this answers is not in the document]
rossgrambo: It only compares the ref. Which is a breaking change if someone has a custom provider that updates the parameters object instead of changing the ref. Our provider creates a new ref when updating, so it only affects custom providers.

src/Microsoft.FeatureManagement/FeatureManager.cs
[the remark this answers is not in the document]
zhenlan: It's my understanding that this change is theoretically a breaking change. Do we feel comfortable releasing this without a major version bump given a customer has a way to get back to the old behavior?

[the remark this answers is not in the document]
jimmyca15: @zhenlan I'd feel more comfortable with a major version bump.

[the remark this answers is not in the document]
zhenlan: Makes sense.

I am not sure how soon we will have the next major version release, so I just throw another idea. What if we release this change with a default cache of zero in v2 (so people can opt-in) and change the default to non-zero in the next major version release?

src/Microsoft.FeatureManagement/FeatureManager.cs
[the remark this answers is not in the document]
zhenlan: I understand it's not optimal. How costly is it to do the value comparison? Alternatively, we don't do the value comparison every time but only when we are about to expire the cache. It could avoid the breaking change.

src/Microsoft.FeatureManagement/FeatureManager.cs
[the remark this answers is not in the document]
jimmyca15: > What if we release this change with a default cache of zero in v2 (so people can opt-in) and change the default to non-zero in the next major version release?

I think that's a solid idea.

src/Microsoft.FeatureManagement/FeatureManager.cs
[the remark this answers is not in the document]
rossgrambo: The way things are written now, we'd have to do the check on every IsEnabledAsync() call, because the goal is to catch this before the cache expires. If the cache expires or is about to expire, there's no issue, as we'd just recompute using the same object. 

Which means in the current state- it's only a breaking change for up to 5 seconds. Even then, it doesn't break anything. It simply means their app will update every 5 seconds instead of sooner. Unless they extended the cache without updating their custom provider, which is the fear.

The check would essentially be a checksum on the parameters, which is fairly fast. Which I thought was a good idea, but we'd essentially be re-adding the same delay that this cache was hoping to solve (parsing the json to a typed object). We could cache the checksum, but if that cache was 5 seconds long, it's the exact same scenario we're already in.

src/Microsoft.FeatureManagement/FeatureManager.cs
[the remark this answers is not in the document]
rossgrambo: Agreed, that's a great idea. I'll go ahead and do that.

src/Microsoft.FeatureManagement/FeatureManagementOptions.cs
```diff
@@ -22,5 +24,12 @@ public class FeatureManagementOptions
         /// The default value is true.
         /// </summary>
         public bool IgnoreMissingFeatures { get; set; } = true;
+
+        /// <summary>
+        /// Controls the duration of the cache from Parameters to Settings.
```
jimmyca15, line 29: There a few points that I think are worth mentioning in this description. I wrote up something that I felt covered them all.

Controls the cache lifetime of settings bound by <cref=IFilterParametersBinder> in the feature management cache.
By default, this cache is off, with a ttl of <cref=TimeSpan.Zero>.
To enable caching of filter parameters, a non-zero ttl should be provided. A recommendation is five seconds.
Increasing the value may cause an observed increase in memory footprint as items live longer.
Lowering the value will decrease performance benefits yielded by caching bound parameters.

src/Microsoft.FeatureManagement/FeatureManagementOptions.cs
```diff
@@ -22,5 +24,12 @@ public class FeatureManagementOptions
...
+        /// If the cache period is set to TimeSpan.Zero, the cache will be disabled.
+        /// The default value is TimeSpan.Zero.
+        /// </summary>
+        public TimeSpan SettingsCachePeriod { get; set; } = TimeSpan.Zero;
```
jimmyca15, line 33: I'd recommend FilterSettingsCacheTtl.

src/Microsoft.FeatureManagement/FeatureManager.cs
```diff
@@ -187,6 +212,48 @@ await featureFilter.EvaluateAsync(context).ConfigureAwait(false) == targetEvalua
...
+                _cache.Set(
+                    context.FeatureName,
+                    new ConfigurationCacheItem
+                    {
+                        Settings = settings,
+                        Parameters = context.Parameters
+                    },
+                    new MemoryCacheEntryOptions
+                    {
+                        AbsoluteExpiration = _options.SettingsCachePeriod.Equals(TimeSpan.Zero) ?
```
jimmyca15, line 244: In .NET the more common comparison method is `==`

jimmyca15: If the cache period is 0, we shouldn't even be putting things in the cache.

jimmyca15:
This is our first option that can be invalid (ttl < 0). Do you mind to add some validation in FeatureManager to validate that cache ttl is greater or equal to 0.

src/Microsoft.FeatureManagement/FeatureManager.cs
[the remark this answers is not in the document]
rossgrambo: Agreed, I think that fear has been mitigated. But we should still call it out in our release/readme so no one gets burned by enabling the cache.

src/Microsoft.FeatureManagement/FeatureManager.cs
```diff
@@ -187,6 +212,49 @@ await featureFilter.EvaluateAsync(context).ConfigureAwait(false) == targetEvalua
...
+                    _cache.Set(
+                        context.FeatureName,
+                        new ConfigurationCacheItem
+                        {
+                            Settings = settings,
+                            Parameters = context.Parameters
+                        },
+                        new MemoryCacheEntryOptions
+                        {
+                            AbsoluteExpirationRelativeToNow = _options.FilterSettingsCacheTtl
```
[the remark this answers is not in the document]
rossgrambo: `SlidingExpiration` continues to slide if it's used, right? So, if it was set to 30 seconds and a user kept refreshing / hitting the same endpoint within 30 seconds. It'd be used indefinitely (unless something else expired it). 

If I understand that correctly, then I could see using it here if we pick a very small value. Like 1 second, so it lasts throughout whatever computation they are doing. 

Because we're enabling developers to pick the cache duration, I'd rather stick with `AbsoluteExpirationRelativeToNow`. Because I think its functionality is what they'd expect.

Do you think `SlidingExpiration` might make more sense here?

src/Microsoft.FeatureManagement/FeatureManager.cs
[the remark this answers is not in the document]
zhenlan: Thanks for the explanation, Ross. 

I hate to dump another "cache expiration" concept (besides what our config provider already had) on customers. From a customer perspective, it's hard to tell why 5s is a good number. As a customer, I would expect everything just works, and if anything changes, the cache updates itself automatically. This is why I was trying to see how bad it is if we make the value/checksum comparison even upon every IsEnabledAsync call. If it doesn't work, can we figure out something better?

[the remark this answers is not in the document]
jimmyca15: Or, if it's the 5 second piece that's confusing, we can just make the switch a boolean.

rossgrambo:
Added validation for TTL. Added an error type and a validation method. https://github.com/microsoft/FeatureManagement-Dotnet/pull/229/commits/95abd6897fc0179dc7049e0928888a2c59ddb044

Let me know if you prefer a different structure or a different error name.

jimmyca15:
> Added validation for TTL. Added an error type and a validation method. [95abd68](https://github.com/microsoft/FeatureManagement-Dotnet/commit/95abd6897fc0179dc7049e0928888a2c59ddb044)
> 
> Let me know if you prefer a different structure or a different error name.

For constructor error I'd expect simple ArgumentException to suffice.

src/Microsoft.FeatureManagement/FeatureManager.cs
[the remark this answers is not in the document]
rossgrambo: So I ran some numbers. Running IsEnabledAsync on a TargetingFilter 10,000 times each:
No Cache: ~2000ms
Cache: 56ms

I wonder if we can get away with just setting the cache to 1 second or even 0.5 seconds. And then just accepting that "FeatureManagement library has up to a 1 second delay when using a custom provider".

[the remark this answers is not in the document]
rossgrambo: Reason being, each instance of turning Params to a typed object in the non-cached world only takes ~0.5ms. 

Which means even if our cache only lives for 100ms, we'd only be adding 0.5% additional time to their total computation time in the worst case. So we get pretty significant gains with small caches.

src/Microsoft.FeatureManagement/FeatureManager.cs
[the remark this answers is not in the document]
avanigupta: I misunderstood. You're right, this should be absolute expiration!

src/Microsoft.FeatureManagement/FeatureManager.cs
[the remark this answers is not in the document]
zhenlan: I'm wondering what the perf looks like if we don't have cache expiration at all, but we evict the cache as soon as a change is detected.

If it's our feature definition provider, we do a reference check (as you did already). If it's a custom provider, we do a full value/checksum comparison.

rossgrambo:
I just realized that this cacheing doesn't work well with multiple filters. Essentially, we cache FeatureName -> Filter.Parameters. 

If I have a 2 Targeting Filters on a single Feature. The parameters of the first one get cached, and then the second filter busts the cache (because the key is the same, but "Parameters" are not the same ref) and it writes a new cache item for those parameters. 

A solution here could be to cache parameters for each Filter, not for the Feature. Something like "Feature.1" pointing to the first filter? That should be safe because we know the ref's will be different on each Parameters object, so even if the order of Filters gets moved around, the cache would bust accordingly. 

What do y'all think?

src/Microsoft.FeatureManagement/FeatureManager.cs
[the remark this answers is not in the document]
rossgrambo: Synced with @zhenlan, 

I got some deep equals metrics. For others, this is a comparison of doing a deep equals on Parameters each call vs binding the parameters in each call. Because I fear Deep Equals are too expensive to be helpful in this scenario.

The timing of BindParameters 10000 times is ~600ms
The timing of DeepEquals 10000 times is ~560ms

Rewriting to the cache 10000 times adds ~30ms. which I think should be added to the BindParameters total time.

This isn't enough of a performance boost over Binding to justify doing that check on each IsEnabledAsync. Here's my version of DeepEquals, let me know if you have any improvements that might make it viable:

```
        private bool AreConfigurationsEqual(IConfiguration config1, IConfiguration config2)
        {
            List<string> keys1 = config1.GetChildren().Select(x => x.Key).ToList();
            List<string> keys2 = config2.GetChildren().Select(x => x.Key).ToList();

            if (!keys1.SequenceEqual(keys2))
            {
                return false;
            }

            foreach (string key in keys1)
            {
                IConfigurationSection section1 = config1.GetSection(key);
                IConfigurationSection section2 = config2.GetSection(key);

                if (section1.Value == null)
                {
                    if (!AreConfigurationsEqual(section1, section2))
                    {
                        return false;
                    }
                } else if (section1.Value != section2.Value)
                {
                    return false;
                }
            }

            return true;
        }
```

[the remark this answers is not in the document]
zhenlan: Thanks for making the comparison, @rossgrambo. It doesn't look like it is worth it for custom providers. So, let's scope this optimization to our built-in providers, as Jimmy suggested. We can open it up to custom providers in the next major release (please open an issue to track it).

rossgrambo:
This is ready for review again. 

Solved the multiple filters cache problem by adding a "FilterIndex" and appending it to the key of the cache. 

Also added an internal interface that allows us to cache params only from only our providers. Had to expose this internal interface to the Tests project via `InternalsVisibleTo` and sign the tests project as well.

src/Microsoft.FeatureManagement/FeatureManagementOptions.cs
```diff
@@ -1,6 +1,8 @@
 ﻿// Copyright (c) Microsoft Corporation.
 // Licensed under the MIT license.
 //
+using System;
```
zhenlan, line 4: Is this change necessary, as I don't see other changes to the file?

rossgrambo: Good catch, no not anymore. I'll remove it.

src/Microsoft.FeatureManagement/FeatureManager.cs
```diff
@@ -187,6 +215,52 @@ await featureFilter.EvaluateAsync(context).ConfigureAwait(false) == targetEvalua
...
+                        cacheKey,
+                        new ConfigurationCacheItem
+                        {
+                            Settings = settings,
+                            Parameters = context.Parameters
+                        },
+                        new MemoryCacheEntryOptions
+                        {
+                            SlidingExpiration = TimeSpan.FromMinutes(5),
+                            AbsoluteExpirationRelativeToNow = TimeSpan.FromDays(1)
```
zhenlan, line 252: Can we define these constants somewhere instead of hardcoding inline?

src/Microsoft.FeatureManagement/ConfigurationWrapper.cs
```diff
@@ -0,0 +1,38 @@
...
+using System.Collections.Generic;
+using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.Primitives;
+
+namespace Microsoft.FeatureManagement
+{
+    /// <summary>
+    /// Wraps an instance of IConfiguration.
+    /// </summary>
+    class ConfigurationWrapper : IConfiguration
```
zhenlan, line 14: Can you please help me understand what we use this wrapper class for?

src/Microsoft.FeatureManagement/IFilterParametersBinder.cs
```diff
@@ -0,0 +1,21 @@
...
+//
+using Microsoft.Extensions.Configuration;
+
+namespace Microsoft.FeatureManagement
+{
+    /// <summary>
+    /// An interface used by the feature management system to pre-bind feature filter parameters to a settings type.
+    /// <see cref="IFeatureFilter"/>s can implement this interface to take advantage of caching of settings by the feature management system.
+    /// </summary>
+    public interface IFilterParametersBinder
```
zhenlan, line 12: We are caching the parameters, not the whole feature definition. I'm wondering if it makes sense to name this interface something like `ICacheableFeatureFilterParameters`, and the requirement is that the `BindParameters` returns a new object instance every call. This way, we don't need the `IFeatureDefinitionProviderCacheable`. In your current code, implementation of this interface is a prerequisite for caching anyway. Thoughts?

zhenlan, line 12: We can make this public in the future. For the release, I would hide it as internal for now so we can change our minds later.

src/Microsoft.FeatureManagement/FeatureManager.cs
rossgrambo: This Wrapper class allows us to update the "ref" when configurations are updated- regardless if the configuration updated it's reference or not. I believe that's it's only purpose- unless Jimmy had other reasons in mind.

src/Microsoft.FeatureManagement/ConfigurationWrapper.cs
rossgrambo: This Wrapper class allows us to update the "ref" when configurations are updated- regardless if the configuration updated it's reference or not. I believe that's it's only purpose- unless Jimmy had other reasons in mind.*

src/Microsoft.FeatureManagement/IFilterParametersBinder.cs
zhenlan: Not sure I follow. This is a new interface, isn't it? Typical feature filters implement `IFeatureFilter`, not this one, right?

rossgrambo: Correct- so essentially:

A **filter** will not use caching, unless the filter implements this interface. 

Meaning the TimeWindowFilter that used to be:
```
public class TimeWindowFilter : IFeatureFilter
```
Has and will continue to be a valid filter. However we want this filter to handle caching, so it becomes:
```
public class TimeWindowFilter : IFeatureFilter, IFilterParametersBinder
```

And then separately, we also added a private interface for the **Provider**. So our provider declares if it can handle caching. 

If the Provider can handle caching, it will use it, but only for Filters that can handle caching (aka- implement this interface)

zhenlan: Right. I don't see the need for a provider to declare cacheable at this point. We have agreed not to expose the cacheability to custom filters in this version yet.

rossgrambo: No, we agreed not to expose cacheable to custom providers, as they're the location we can't guarantee a ref update. Custom Filters don't have the same risk.

zhenlan: What's the customer value of exposing this interface without cacheability?

rossgrambo: If they are using our default provider- they can use Custom Filters with caching enabled.

src/Microsoft.FeatureManagement/ConfigurationWrapper.cs
zhenlan: Got it. Thanks. Would you mind adding some comments to clarify the necessity of this?

rossgrambo: Good idea, I'll do that

src/Microsoft.FeatureManagement/IFilterParametersBinder.cs
zhenlan: Thanks for the clarification. Is this interface useful in scenarios other than caching?

rossgrambo: Not that I can think of. Only other benefit is a well-defined way to go from IConfiguration to the filter specific CustomFilterSettings object. But I don't think that it would be used for that.

zhenlan: Right. It also appears to me this interface is only useful if a filter binds parameters to an object and cares about caching. A custom filter can work directly with the `IConfiguration` of parameters without worrying about all this. Can we name the interface so it's clearer why a user wants to implement it?

jimmyca15: The interface exposes a single method, `BindParameters`. I'd argue `IFilterParametersBinder` is aptly named. Any additional information about usage seems better placed in a summary.

rossgrambo: After re-reading the summaries- I can't come up with a better description that provides any extra info. I'm going to resolve this conversation for now- but let me know if you can come up with a better summary. 

Pasting the current summaries here for easy reference:
```
    /// <summary>
    /// An interface used by the feature management system to pre-bind feature filter parameters to a settings type.
    /// <see cref="IFeatureFilter"/>s can implement this interface to take advantage of caching of settings by the feature management system.
    /// </summary>
    public interface IFilterParametersBinder
    {
        /// <summary>
        /// Binds a set of feature filter parameters to a settings object.
        /// </summary>
        /// <param name="parameters">The configuration representing filter parameters to bind to a settings object</param>
        /// <returns>A settings object that is understood by the implementer of <see cref="IFilterParametersBinder"/>.</returns>
        object BindParameters(IConfiguration parameters);
    }
```

jimmyca15: @zhenlan

Thoughts?

jimmyca15: @zhenlan are you good with this?

jimmyca15: @rossgrambo 

Discussed offline with @zhenlan. We agreed to keep the interface the way it is now.

jimmyca15: @rossgrambo 

Discussed with @zhenlan. We agreed to have it public.

src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
```diff
@@ -15,7 +15,7 @@ namespace Microsoft.FeatureManagement
     /// <summary>
     /// A feature definition provider that pulls feature definitions from the .NET Core <see cref="IConfiguration"/> system.
     /// </summary>
-    sealed class ConfigurationFeatureDefinitionProvider : IFeatureDefinitionProvider, IDisposable
+    sealed class ConfigurationFeatureDefinitionProvider : IFeatureDefinitionProvider, IDisposable, IFeatureDefinitionProviderCacheable
```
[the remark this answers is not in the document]
jimmyca15: Perhaps also in that types file as well.

src/Microsoft.FeatureManagement/FeatureManager.cs
```diff
@@ -187,6 +218,52 @@ await featureFilter.EvaluateAsync(context).ConfigureAwait(false) == targetEvalua
             return enabled;
         }
 
+        private void BindSettings(IFeatureFilterMetadata filter, FeatureFilterEvaluationContext context)
+        {
+            IFilterParametersBinder binder = filter as IFilterParametersBinder;
+
+            if (binder == null)
```
jimmyca15, line 225: What do you think about having a separate short circuit like

```
if (!(_featureDefinitionProvider is IFeatureDefinitionProviderCacheable))
{
    context.Settings = binder.BindParameters(context.Parameters);

    return;
}
```

Suggesting it because any interaction with cache key or cache doesn't make sense if the definition provider isn't cachable.

rossgrambo: Good idea, makes the cache if statement a little cleaner as a result.

src/Microsoft.FeatureManagement/FeatureManager.cs
```diff
@@ -135,6 +160,7 @@ private async Task<bool> IsEnabledAsync<TContext>(string feature, TContext appCo
                         var context = new FeatureFilterEvaluationContext()
                         {
                             FeatureName = feature,
+                            FilterIndex = filterIndex,
```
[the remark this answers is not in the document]
rossgrambo: I was thinking it would be relevant for telemetry as well (as we'll mark which filter returned true). But I no longer think that we actually need the index for those changes. I'll go ahead with your suggestion.

jimmyca15 (approved):

Issue #244, opened 2023-06-27
CustomFilter breaks when upgrading to 2.6.0 with null reference exception


timvandenhof:
Affected version: 2.6.0
Non-affected version: 2.5.1

# Description
When we upgraded Microsoft.FeatureManagement.AspNetCore from 2.5.0 to 2.6.0, we noticed a runtime null reference exception. The null reference exception occurs inside `TryValidateSettings` of the `ContextualTargettingFilter` class.
This seems to be caused by `context.settings` being `null`.

<details>
  <summary>Stacktrace via Application Insights</summary>

```
System.NullReferenceException:
   at Microsoft.FeatureManagement.FeatureFilters.ContextualTargetingFilter.TryValidateSettings (Microsoft.FeatureManagement, Version=2.6.0.0, Culture=neutral, PublicKeyToken=69dad7634abb75e4)
   at Microsoft.FeatureManagement.FeatureFilters.ContextualTargetingFilter.EvaluateAsync (Microsoft.FeatureManagement, Version=2.6.0.0, Culture=neutral, PublicKeyToken=69dad7634abb75e4)
   at OURNAMESPACE.CustomFeatureFlagFilter+<EvaluateTargetingAsync>d__5.MoveNext (OURLIBRARY, Version=3.115.1.0, Culture=neutral, PublicKeyToken=null: OURPATH\CustomFeatureFlagFilter.cs:45)
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw (System.Private.CoreLib, Version=7.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e)
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess (System.Private.CoreLib, Version=7.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification (System.Private.CoreLib, Version=7.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e)
   at System.Runtime.CompilerServices.TaskAwaiter`1.GetResult (System.Private.CoreLib, Version=7.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e)
[...]

PR #245, opened 2023-06-27
Add additional null checking to ContextualTargetingFilter

timvandenhof:
Draft fix for issue #244.
This PR needs additional tests to verify this actually fixes the issue mentioned.
Not sure if the `?.` is desired to be used, as the if statement itself is very explicit in null checking.


timvandenhof:
Will cancel this PR, in favor of #246.

PR #246, opened 2023-06-27
Adds edge case for EvaluateAsync call that doesn't use context from FeatureManager

rossgrambo:
Handles edge case from https://github.com/microsoft/FeatureManagement-Dotnet/issues/244

The parameters to settings binding happens in the FeatureManager before calling each individual filter. This lead to the assumption that all filters would have the settings object available. 

In this case, the filter is called with context that was not binded by FeatureManager. This meant settings could be null.

This fix adds a check for that situation and binds if needed. When hitting this scenario, the app doesn't benefit from the caching changes described in https://github.com/microsoft/FeatureManagement-Dotnet/releases/tag/2.6.0-preview2

### DIFF
--- a/src/Microsoft.FeatureManagement/FeatureFilters/TimeWindowFilter.cs
+++ b/src/Microsoft.FeatureManagement/FeatureFilters/TimeWindowFilter.cs
@@ -43,7 +43,9 @@ namespace Microsoft.FeatureManagement.FeatureFilters
         /// <returns>True if the feature is enabled, false otherwise.</returns>
         public Task<bool> EvaluateAsync(FeatureFilterEvaluationContext context)
         {
-            TimeWindowFilterSettings settings = (TimeWindowFilterSettings)context.Settings;
+            //
+            // Assume settings was bound from feature manager, if not, bind from parameters
+            TimeWindowFilterSettings settings = (TimeWindowFilterSettings)context.Settings ?? (TimeWindowFilterSettings)BindParameters(context.Parameters);
 
             DateTimeOffset now = DateTimeOffset.UtcNow;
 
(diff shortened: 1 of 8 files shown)


src/Microsoft.FeatureManagement/FeatureFilters/TimeWindowFilter.cs
```diff
@@ -43,7 +43,9 @@ public object BindParameters(IConfiguration filterParameters)
         /// <returns>True if the feature is enabled, false otherwise.</returns>
         public Task<bool> EvaluateAsync(FeatureFilterEvaluationContext context)
         {
-            TimeWindowFilterSettings settings = (TimeWindowFilterSettings)context.Settings;
+            //
+            // Assume settings was bound from feature manager, if not, bind from parameters
```
jimmyca15, line 47: > bound from feature manager

I don't like how the comment mentions an expected caller. Public method can be called in any way.

I'd suggest
> Check if prebound settings available, otherwise bind from parameters.

tests/Tests.FeatureManagement/CustomFilterTargeting.cs
```diff
@@ -0,0 +1,31 @@
...
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Options;
+using Microsoft.FeatureManagement;
+using Microsoft.FeatureManagement.FeatureFilters;
+using System;
+using System.Threading.Tasks;
+
+namespace Tests.FeatureManagement
+{
+    class CustomFilterTargeting : IFeatureFilter
```
jimmyca15, line 14: Usually filter names end with Filter. So `CustomTargetingFilter`. Any reason to stray?

rossgrambo: No great reason. Just wanted to note it's a custom filter- but uses a targeting filter within. Was originally `CustomFilterWithTargeting`.

`CustomTargetingFilter` sounded like it's a different type of targeting. But happy to use that name to be consistent.

jimmyca15 (approved):

PR #251, opened 2023-08-11
Add edge case for exact 100 bucket

rossgrambo:
Adds an edge case for the `1 / MaxInt` chance that a person is placed in exactly at 100%, which currently isn't handled by the existing `<`. Using `<=` would cause other percents to fall in multiple buckets, so the edge case only applies to 100, where there is no higher bucket.

### DIFF
--- a/src/Microsoft.FeatureManagement/Targeting/ContextualTargetingFilter.cs
+++ b/src/Microsoft.FeatureManagement/Targeting/ContextualTargetingFilter.cs
@@ -142,6 +142,13 @@ namespace Microsoft.FeatureManagement.FeatureFilters
         /// <returns>A boolean representing if the context identifier should be targeted</returns>
         private bool IsTargeted(string contextId, double percentage)
         {
+            //
+            // Handle edge case of exact 100 bucket
+            if (percentage == 100)
+            {
+                return true;
+            }
+
             byte[] hash;
 
             using (HashAlgorithm hashAlgorithm = SHA256.Create())


src/Microsoft.FeatureManagement/Targeting/ContextualTargetingFilter.cs
```diff
@@ -156,6 +156,13 @@ private bool IsTargeted(string contextId, double percentage)
 
             double contextPercentage = (contextMarker / (double)uint.MaxValue) * 100;
 
+            //
+            // Handle edge case of exact 100 bucket
```
jimmyca15, line 160: If percentage is 100, you can immediately return true at the beginning of the method.

rossgrambo: If I do the immediate return, it will be ripped out and turned into something like this as a part of the variants changes. I think this is cleaner with that future in mind.

jimmyca15: This PR has no notion of variants.

jimmyca15: The only thing I can imagine variants would play into is whether or not we should have this PR. If we do want it, even though variants are on the horizon, I don't expect the implementation should depend on that future change

jimmyca15: My concern is based off the idea that at any point in time, any user should be able to visit the repo, and the code should make sense. Having PRs that mix in the concerns of future changes can lead to confusing point-in-time source code.

PR #256, opened 2023-08-31
Support recurring time window filter for TimeWindowFilter

zhiyuanliang-ms:
Implemented the CrontabField and CrontabExpression classes and wrote some unit tests


src/Microsoft.FeatureManagement/FeatureFilters/Crontab/CrontabExpression.cs
```diff
@@ -0,0 +1,82 @@
...
+using System;
+
+namespace Microsoft.FeatureManagement.FeatureFilters.Crontab
+{
+    /// <summary>
+    /// The Crontab expression with Minute, Hour, Day of month, Month, Day of week fields.
+    /// </summary>
+    public class CrontabExpression
+    {
+        static private readonly int _numberOfFields = 5;
```
juniwang, line 14: `private static`

src/Microsoft.FeatureManagement/FeatureFilters/Crontab/CrontabExpression.cs
```diff
@@ -0,0 +1,82 @@
...
+        private readonly CrontabField _monthField = new CrontabField(CrontabFieldKind.Month);
+        private readonly CrontabField _dayOfWeekField = new CrontabField(CrontabFieldKind.DayOfWeek);
+
+        /// <summary>
+        /// Checks whether the given expression can be parsed by the Crontab expression.
+        /// </summary>
+        /// <param name="expression">The expression to parse.</param>
+        /// <param name="message">The error message.</param>
+        /// <returns>True if the expression is valid and parsed successfully, otherwise false.</returns>
+        public bool Parse(string expression, out string message)
```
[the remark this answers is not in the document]
juniwang: Also `Parse` gives me the impression that it should return the parsed object. Maybe `TryParse`?

```
public CrontabExpression Parse(string expression)
{
    // implementation
}
```

or:

```
public bool TryParse(string expression, out CrontabExpression crobtab)
{
    // implementation
}
```

src/Microsoft.FeatureManagement/FeatureFilters/Crontab/CrontabExpression.cs
```diff
@@ -0,0 +1,82 @@
...
+        {
+            if (expression == null)
+            {
+                message = $"Expression is null. Five fields in the sequence of Minute, Hour, Day of month, Month, and Day of week are required.";
+                return false;
+            }
+
+            message = "";
+            char[] whitespaceDelimiters = {' ', '\n', '\t'};
```
juniwang, line 37: this can be const or static readonly.

src/Microsoft.FeatureManagement/FeatureFilters/Crontab/CrontabExpression.cs
```diff
@@ -0,0 +1,82 @@
...
+                _minuteField,
+                _hourField,
+                _dayOfMonthField,
+                _monthField,
+                _dayOfWeekField
+            };
+
+            for (int i = 0; i < fieldObjs.Length; i++)
+            {
+                if (fieldObjs[i].Parse(fields[i], out message) == false)
```
juniwang, line 57: similar as above

src/Microsoft.FeatureManagement/FeatureFilters/Crontab/CrontabField.cs
```diff
@@ -0,0 +1,217 @@
...
+        public bool Parse(string content, out string message)
+        {
+            message = "";
+            _bits.SetAll(false);
+
+            //
+            // The field can be a list which is a set of numbers or ranges separated by commas.
+            // Ranges are two numbers/name separated with a hyphen or an asterisk which represents all possible values in the field.
+            // Step values can be used in conjunction with ranges after a slash.
+            string validSegmentPattern = @"^(?:[0-9]+|[a-zA-Z]{3}|(?:\*|(?:[0-9]+|[a-zA-Z]{3})-(?:[0-9]+|[a-zA-Z]{3}))(/[0-9]+)?)$";
```
juniwang, line 80: are we using one single regex to test all fields? this makes the regex very complex and hard to understand. Any way to make it more comprehensible?

juniwang: Also the pattern should be constant. Or a static Regex for reuse.

src/Microsoft.FeatureManagement/FeatureFilters/Crontab/CrontabField.cs
```diff
@@ -0,0 +1,217 @@
...
+                    return GetDayOfWeekNumber(str);
+                }
+                else
+                {
+                    return -1;
+                }
+            }
+        }
+
+        static private int GetMonthNumber(string name)
```
juniwang, line 172: `private static`

src/Microsoft.FeatureManagement/FeatureFilters/Crontab/CrontabField.cs
```diff
@@ -0,0 +1,217 @@
...
+                "AUG" => 8,
+                "SEP" => 9,
+                "OCT" => 10,
+                "NOV" => 11,
+                "DEC" => 12,
+                _ => -1
+            };
+        }
+
+        static private int GetDayOfWeekNumber(string name)
```
juniwang, line 192: `private static`

src/Microsoft.FeatureManagement/FeatureFilters/Crontab/CrontabField.cs
zhiyuanliang-ms: I can add more comments to explain this regex. But I think using one single regex can make the code tidy and simple.

src/Microsoft.FeatureManagement/FeatureFilters/Crontab/CrontabField.cs
```diff
@@ -0,0 +1,242 @@
...
+                            return false;
+                        }
+
+                        if (_kind == CrontabFieldKind.DayOfWeek && last == 0 && last != first) // Corner case for Sunday: both 0 and 7 can be interpreted to Sunday
+                        {
+                            last = 7; // Mon-Sun should be intepreted to 1-7 instead of 1-0
+                        }
+                    }
+
+                    for (int num = first; num <= last; num += step)
```
amerjusupovic, line 145: Related to other comment on unit test, but are we intentionally ignoring `31-1/100` with this logic?

tests/Tests.FeatureManagement/Crontab.cs
```diff
@@ -0,0 +1,75 @@
...
+using Xunit;
+
+namespace Tests.FeatureManagement
+{
+    public class CrontabTest
+    {
+        [Theory]
+        [InlineData("* * * * *", true)]
+        [InlineData("1 2 3 Apr Fri", true)]
+        [InlineData("00-59/3,1,2-2 01,3,20-23 */10,31-1/100 Apr,1-Feb,oct-DEC/1 Sun-Sat/2,1-7", true)]
```
amerjusupovic, line 16: Could you explain the `31-1/100` part of the day value? Does cron support the day of the month wrapping around from 31 to 1?

src/Microsoft.FeatureManagement/FeatureFilters/Crontab/CrontabField.cs
zhiyuanliang-ms: Yes. This is the behavior of NCronTab when it is parsing "* * 31-1/100 * *".

tests/Tests.FeatureManagement/Crontab.cs
zhiyuanliang-ms: The IBM doc and man page of crontab(5) both don't mention whether this is valid.
https://crontab.guru/ this website will consider this invalid. (By the way, this website recognized the step value 100, it only denied range like 100-1)
However, the NCronTab library will treat "31-1/100" as a valid expression and it will be considered as the 31st day of the month.
I think the range like 31-1 and large step value like 100 can be interpreted as the 31st to 1st day of month per 100 days, which will give us only the 31st day of the month.
So I believe this should be valid and supported. This is not a big deal or a critical error.
I can add some extra logics to check whether the step value is within the max value of crontab field and check whether the range is from a smaller number to a a larger number.
What do you think? Should we treat "31-1/100" as a invalid one?

src/Microsoft.FeatureManagement/FeatureFilters/Crontab/CrontabExpression.cs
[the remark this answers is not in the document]
zhiyuanliang-ms: Implemented public static CrontabExpression Parse(string expression) and private bool TryParse(string expression, out string message)
The Parse method will call TryParse and throw exception when parsing an invalid expression.

src/Microsoft.FeatureManagement/FeatureFilters/Crontab/CrontabExpression.cs
zhiyuanliang-ms: Done

src/Microsoft.FeatureManagement/FeatureFilters/Crontab/CrontabExpression.cs
zhiyuanliang-ms: Done

src/Microsoft.FeatureManagement/FeatureFilters/Crontab/CrontabExpression.cs
```diff
@@ -0,0 +1,82 @@
...
+        /// <summary>
+        /// Checks whether the Crontab expression is satisfied by the given timestamp.
+        /// </summary>
+        /// <param name="time">The timestamp to check.</param>
+        /// <returns>True if the Crontab expression is satisfied by the give timestamp, otherwise false.</returns>
+        public bool IsSatisfiedBy(DateTimeOffset time)
+        {
+            bool isMatched;
+            isMatched = _dayOfWeekField.Match((int)time.DayOfWeek)
+                      & _monthField.Match((int)time.Month)
```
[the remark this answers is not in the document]
zhiyuanliang-ms: Done

src/Microsoft.FeatureManagement/FeatureFilters/Crontab/CrontabField.cs
zhiyuanliang-ms: Re-implemented the CrontabField.TryParse() without using regex.

src/Microsoft.FeatureManagement/FeatureFilters/Crontab/CrontabField.cs
zhiyuanliang-ms: Done

src/Microsoft.FeatureManagement/FeatureFilters/Crontab/CrontabField.cs
zhiyuanliang-ms: Done

tests/Tests.FeatureManagement/Crontab.cs
amerjusupovic: I didn't realize NCronTab treated that expression as valid, so it makes more sense to stick with what the NCronTab implementation is and treat that expression as just the 31st day. As long as it's clear that we're sticking with this library then I think it's ok.

zhiyuanliang-ms: I think "first-last/step" should be interpreted to something like a for loop: for (int i = first; i<= last; i+=step). If we consider it in this way, everything will make sense.

zhiyuanliang-ms: Update: The range like "31-1" should be forbidden. The reason is as follow:
1. The portal should validate the Crontab expression typed by user. If such reversed range is allowed, user may be confused.
2. Although Quartz and NCronTab will treat this as a valid expression, it seems like a bug or a corner case they didn't handle. After more tests, I found that the behavior of NextOccurence(), IsSatisfiedBy(DateTimeOffset) will be extremly strange when both libraries were handling the Crontab like "* * 31-1/100 * *".
For example:
UTs for NCronTab:
```
[InlineData("* * 31-29/2 * *", "Tue, 29 Aug 2023 00:00:00 +00:00", true)]
[InlineData("* * 31-1/30 * *", "Thu, 31 Aug 2023 01:00:00 +00:00", true)]
[InlineData("* * 31-1/30 * *", "Fri, 1 Sep 2023 01:00:00 +00:00", true)]
[InlineData("* * 31-1/31 * *", "Fri, 1 Sep 2023 01:00:00 +00:00", true)]
[InlineData("* * 31-1/31 * *", "Thu, 31 Aug 2023 01:00:00 +00:00", false)]
[InlineData("* * 31-1/100 * *", "Thu, 31 Aug 2023 01:00:00 +00:00", false)]
[InlineData("* * 1-31/100 * *", "Fri, 1 Sep 2023 01:00:00 +00:00", true)]
[InlineData("* * 1-31/100 * *", "Thu, 31 Aug 2023 01:00:00 +00:00", false)]
public void IsCronSatisfiedByTimestamp(string cron, string timeString, bool expected)
{
    Assert.Equal(expected, IsSatisfiedBy(cron, timeString));
}
private static bool IsSatisfiedBy(string cron, string timeString)
{
    CrontabSchedule schedule = CrontabSchedule.Parse(cron, new CrontabSchedule.ParseOptions { IncludingSeconds = false });

    DateTime dateTime = DateTime.Parse(timeString);
    DateTime withoutSeconds = new DateTime(dateTime.Year, dateTime.Month, dateTime.Day, dateTime.Hour, dateTime.Minute, 0);
    DateTime timeAfter = schedule.GetNextOccurrence(withoutSeconds.AddMinutes(-1));
    if (timeAfter.Equals(withoutSeconds))
    {
        return true;
    } 
    else
    {
        return false;
    } 
}
```

src/Microsoft.FeatureManagement/FeatureFilters/TimeWindowFilterSettings.cs
```diff
@@ -14,12 +17,19 @@ public class TimeWindowFilterSettings
...
         /// If no end time is specified the time window is considered to never end.
         /// </summary>
         public DateTimeOffset? End { get; set; } // E.g. "Wed, 01 May 2019 23:00:00 GMT"
+
+        /// <summary>
+        /// An optional list which specifies the recurring time windows used to determine when a feature configured to use the <see cref="TimeWindowFilter"/> feature filter should be enabled.
+        /// The recurring time windows are represented in the form of Crontab expression.
+        /// If any recurring time window filter is specified, to enable the feature flag, the current time also needs to be within at least one of the recurring time windows.
+        /// </summary>
+        public List<CrontabExpression> Filters { get; set; } = new List<CrontabExpression>(); // E.g. ["* 18-19 * * Mon"] which means the recurring time window of 18:00~20:00 on Monday
```
jimmyca15, line 33: I don't see any reason to define public type for cron tab expression. I expect string to suffice.

jimmyca15, line 33: I think we should use Cron Expression nomenclature that seems to be a norm. The man pages we refer to as our spec actually call it a "cron command" by the way.

src/Microsoft.FeatureManagement/FeatureFilters/TimeWindowFilterSettings.cs
```diff
@@ -14,12 +17,19 @@ public class TimeWindowFilterSettings
...
+        public DateTimeOffset? Start { get; set; } // E.g. "Wed, 01 May 2019 22:59:30 GMT+0800"
 
         /// <summary>
         /// An optional end time used to determine when a feature configured to use the <see cref="TimeWindowFilter"/> feature filter should be enabled.
         /// If no end time is specified the time window is considered to never end.
         /// </summary>
         public DateTimeOffset? End { get; set; } // E.g. "Wed, 01 May 2019 23:00:00 GMT"
+
+        /// <summary>
+        /// An optional list which specifies the recurring time windows used to determine when a feature configured to use the <see cref="TimeWindowFilter"/> feature filter should be enabled.
```
jimmyca15, line 29:
```suggestion
        /// An optional list of cron expressions that can be used to filter out what time within the time window is applicable.
        /// This setting can be used to set up a recurring time window.
```

src/Microsoft.FeatureManagement/FeatureFilters/TimeWindowFilterSettings.cs
jimmyca15, line 33: > E.g. ["* 18-19 * * Mon"] which means the recurring time window of 18:00~20:00 on Monday

This looks more like it should belong in a doc than as a comment here. Also this seems kind of like a stray comment, was it intended to be part of the method summary?

jimmyca15, line 33: Why list instead of IEnumerable?

jimmyca15: Ah I see, because we have comments like this on the other properties. Those seem odd. I would suggest the opposite, that we remove the stray comments from the existing properties.

src/Microsoft.FeatureManagement/FeatureFilters/Crontab/CrontabFieldKind.cs
```diff
@@ -0,0 +1,42 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+//
+namespace Microsoft.FeatureManagement.FeatureFilters.Crontab
+{
+    /// <summary>
+    /// Define an enum of all required fields of the Crontab expression.
+    /// </summary>
+    public enum CrontabFieldKind
```
jimmyca15, line 9: I'm not expecting any new public types as part of this enhancement.

src/Microsoft.FeatureManagement/FeatureFilters/TimeWindowFilter.cs
```diff
@@ -3,13 +3,17 @@
...
+using Microsoft.FeatureManagement.FeatureFilters.Crontab;
 using System;
+using System.Collections.Generic;
 using System.Threading.Tasks;
 
 namespace Microsoft.FeatureManagement.FeatureFilters
 {
     /// <summary>
-    /// A feature filter that can be used to activate a feature based on a time window.
+    /// A feature filter that can be used to activate a feature based on time window.
```
jimmyca15, line 14:
```suggestion
    /// A feature filter that can be used to activate a feature based on a singular or recurring time window.
```

src/Microsoft.FeatureManagement/FeatureFilters/TimeWindowFilter.cs
```diff
@@ -33,7 +37,26 @@ public TimeWindowFilter(ILoggerFactory loggerFactory)
         /// <returns><see cref="TimeWindowFilterSettings"/> that can later be used in feature evaluation.</returns>
         public object BindParameters(IConfiguration filterParameters)
         {
-            return filterParameters.Get<TimeWindowFilterSettings>() ?? new TimeWindowFilterSettings();
+            DateTimeOffset? Start = filterParameters["Start"] != null ? filterParameters.GetValue<DateTimeOffset>("Start") : (DateTimeOffset?)null;
```
jimmyca15, line 40: I like to avoid repeated literals. I.e. "Start"

jimmyca15: So GetValue<DateTimeOffset?>() doesn't work?

src/Microsoft.FeatureManagement/FeatureFilters/TimeWindowFilter.cs
```diff
@@ -3,13 +3,17 @@
 //
```
jimmyca15, line 3: The file uses new lines between statements. Let's preserve the styling with our new additions.

src/Microsoft.FeatureManagement/FeatureFilters/TimeWindowFilter.cs
```diff
@@ -33,7 +37,26 @@ public TimeWindowFilter(ILoggerFactory loggerFactory)
         /// <returns><see cref="TimeWindowFilterSettings"/> that can later be used in feature evaluation.</returns>
         public object BindParameters(IConfiguration filterParameters)
         {
-            return filterParameters.Get<TimeWindowFilterSettings>() ?? new TimeWindowFilterSettings();
+            DateTimeOffset? Start = filterParameters["Start"] != null ? filterParameters.GetValue<DateTimeOffset>("Start") : (DateTimeOffset?)null;
+            DateTimeOffset? End = filterParameters["End"] != null ? filterParameters.GetValue<DateTimeOffset>("End") : (DateTimeOffset?)null;
+
+            TimeWindowFilterSettings settings = new TimeWindowFilterSettings()
```
jimmyca15, line 43:
```suggestion
            var settings = new TimeWindowFilterSettings
```

src/Microsoft.FeatureManagement/FeatureFilters/TimeWindowFilter.cs
```diff
@@ -33,7 +37,26 @@ public TimeWindowFilter(ILoggerFactory loggerFactory)
...
+            {
+                Start = Start,
+                End = End
+            };
+
+            List<string> Filters = filterParameters.GetSection("Filters").Get<List<string>>();
```
jimmyca15, line 49: local variables should use lowercase param names.


```suggestion
            IEnumerable<string> filters = filterParameters.GetSection("Filters").Get<IEnumerable<string>>();
```

src/Microsoft.FeatureManagement/FeatureFilters/TimeWindowFilter.cs
```diff
@@ -33,7 +37,26 @@ public TimeWindowFilter(ILoggerFactory loggerFactory)
...
+            if (Filters != null)
+            {
+                foreach (string expression in Filters)
+                {
+                    CrontabExpression crontabExpression = CrontabExpression.Parse(expression);
+                    settings.Filters.Add(crontabExpression);
```
jimmyca15, line 55: I would recommend to build the filters than assign it to settings, rather than assuming initial state of settings Filters.

src/Microsoft.FeatureManagement/FeatureFilters/TimeWindowFilter.cs
```diff
@@ -49,14 +72,43 @@ public Task<bool> EvaluateAsync(FeatureFilterEvaluationContext context)
 
             DateTimeOffset now = DateTimeOffset.UtcNow;
 
-            if (!settings.Start.HasValue && !settings.End.HasValue)
+            if (!settings.Start.HasValue && !settings.End.HasValue && settings.Filters.Count == 0)
             {
-                _logger.LogWarning($"The '{Alias}' feature filter is not valid for feature '{context.FeatureName}'. It must have have specify either '{nameof(settings.Start)}', '{nameof(settings.End)}', or both.");
+                _logger.LogWarning($"The '{Alias}' feature filter is not valid for feature '{context.FeatureName}'. It must specify at least one of '{nameof(settings.Start)}', '{nameof(settings.End)}' and '{nameof(settings.Filters)}'.");
```
jimmyca15, line 77:
```suggestion
                _logger.LogWarning($"The '{Alias}' feature filter is not valid for feature '{context.FeatureName}'. It must specify at least one of '{nameof(settings.Start)}', '{nameof(settings.End)}' or '{nameof(settings.Filters)}'.");
```

src/Microsoft.FeatureManagement/FeatureFilters/Crontab/CrontabExpression.cs
```diff
@@ -0,0 +1,115 @@
...
+            }
+            return true;
+        }
+
+        /// <summary>
+        /// Checks whether the Crontab expression is satisfied by the given timestamp.
+        /// </summary>
+        /// <param name="time">The timestamp to check.</param>
+        /// <returns>True if the Crontab expression is satisfied by the give timestamp, otherwise false.</returns>
+        public bool IsSatisfiedBy(DateTimeOffset time)
```
jimmyca15, line 80: Please order public before private.

src/Microsoft.FeatureManagement/FeatureFilters/Crontab/CrontabExpression.cs
```diff
@@ -0,0 +1,115 @@
...
+        /// <summary>
+        /// Check whether the given expression can be parsed by a Crontab expression.
+        /// If the expression is invalid, an ArgumentException will be thrown.
+        /// </summary>
+        /// <param name="expression">The expression to parse.</param>
+        /// <returns>A parsed CrontabExpression.</returns>
+        public static CrontabExpression Parse(string expression)
+        {
+            CrontabExpression crontabExpression = new CrontabExpression();
+            if (crontabExpression.TryParse(expression, out string message) == false)
```
jimmyca15, line 34: TryParse is a common pattern in .NET. Usually it contains an out parameter of what it parsed. This doesn't seem to fit.

src/Microsoft.FeatureManagement/FeatureFilters/Crontab/CrontabExpression.cs
```diff
@@ -0,0 +1,115 @@
...
+            {
+                throw new ArgumentException($"Crontab expression: \"{expression}\" is invalid. " + message);
```
jimmyca15, line 36: Argument exception takes argument name as a parameter.

src/Microsoft.FeatureManagement/FeatureFilters/Crontab/CrontabExpression.cs
```diff
@@ -0,0 +1,115 @@
...
+{
+    /// <summary>
+    /// The Crontab expression with Minute, Hour, Day of month, Month, Day of week fields.
+    /// </summary>
+    public class CrontabExpression
+    {
+        private static readonly int NumberOfFields = 5;
+        private static readonly char[] WhitespaceDelimiters = {' ', '\n', '\t'};
+
+        private readonly CrontabField[] CrontabFields = new CrontabField[]
```
jimmyca15, line 16:
```suggestion
        private CronTabField Minute { get; set; }
        
        private CronTabField Hour { get; set; }

        ...
```

src/Microsoft.FeatureManagement/FeatureFilters/Crontab/CrontabExpression.cs
```diff
@@ -0,0 +1,115 @@
...
+
+            if (fields.Length != NumberOfFields)
+            {
+                message = $"Five fields in the sequence of Minute, Hour, Day of month, Month, and Day of week are required.";
+                return false;
+            }
+
+            for (int i = 0; i < NumberOfFields; i++)
+            {
+                if (CrontabFields[i].TryParse(fields[i], out message) == false)
```
jimmyca15, line 67: CronField field;

if (!CronField.TryParse(Kind.Minute, fields[0], out field))
{
    return false;
}

Minute = field;

src/Microsoft.FeatureManagement/FeatureFilters/Crontab/CrontabExpression.cs
```diff
@@ -0,0 +1,115 @@
...
+            the month and day of month fields shall specify the days that match. 
+
+            If both month and day of month are specified as an <asterisk>, but day of week is an element or list, 
+            then only the specified days of the week match. 
+
+            If either the month or day of month is specified as an element or list, and the day of week is also specified as an element or list, 
+            then any day matching either the month and day of month, or the day of week, shall be matched.
+             */
+            bool isDayMatched;
+            if (!CrontabFields[4].IsAsterisk && // the day of week is specified
```
jimmyca15, line 95:
```suggestion
            if (!DayOfWeek.MatchAll && // the day of week is specified 
```

src/Microsoft.FeatureManagement/FeatureFilters/TimeWindowFilterSettings.cs
zhiyuanliang-ms: If the user sets the "Start" and "End" parameter with invalid timestamp string which cannot be converted into a DateTimeOffset, an exception will be thrown during BindParameters(). If we just store the raw Crontab string here, only when the filter is evaluated, the invalid Crontab will be detected and an exception will be thrown.

I think the job for the IFilterParametersBinder.BindParameters() is to pre-process the configuration paramters to some settings that filter can use directly. For example, the "Start" and "End" parameters are just string in the configuration, but it will become a DataTimeOffset in the TimeWindowFilterSettings.

zhiyuanliang-ms: The Audience class for the targeting filter has a property public List<GroupRollout> Groups { get; set; }. 
I can change it to IEnumerable.

src/Microsoft.FeatureManagement/FeatureFilters/TimeWindowFilter.cs
zhiyuanliang-ms: Ah, my bad, I didn't expect GetValue<> to be so powerful. 
It seems like GetValue<DateTimeOffset>() is enough. Thank you for the suggestion.

src/Microsoft.FeatureManagement/FeatureFilters/TimeWindowFilter.cs
zhiyuanliang-ms: I moved the code of parsing Cron to the EvaluateAsync() instead if BindParameters().
Now, the TimeWindowFilterSettings will store the raw string of Crons.

src/Microsoft.FeatureManagement/FeatureFilters/Crontab/CrontabExpression.cs
zhiyuanliang-ms: Great suggestion!

src/Microsoft.FeatureManagement/FeatureFilters/TimeWindowFilter.cs
zhiyuanliang-ms: Sorry, I'm not sure I fully grasp your suggestion. Could you please clarify this?

src/Microsoft.FeatureManagement/FeatureFilters/TimeWindowFilterSettings.cs
zhiyuanliang-ms: I have removed these comments.

src/Microsoft.FeatureManagement/FeatureFilters/Cron/CronExpression.cs
```diff
@@ -0,0 +1,99 @@
...
+        private static readonly int NumberOfFields = 5;
+        private static readonly char[] WhitespaceDelimiters = {' ', '\n', '\t'};
+
+        private readonly CronField _minute;
+        private readonly CronField _hour;
+        private readonly CronField _dayOfMonth;
+        private readonly CronField _month;
+        private readonly CronField _dayOfWeek;
+
+        public CronExpression(CronField minute, CronField hour, CronField dayOfMonth, CronField month, CronField dayOfWeek)
```
jimmyca15, line 22: Can this constructor be private?

src/Microsoft.FeatureManagement/FeatureFilters/Cron/CronField.cs
```diff
@@ -0,0 +1,241 @@
...
+{
+    /// <summary>
+    /// The CronField uses a BitArray to record which values are included in the field.
+    /// </summary>
+    internal class CronField
+    {
+        /// <summary>
+        /// Whether the CronField is an asterisk.
+        /// </summary>
+        public bool MatchAll { get; private set; } = false;
```
jimmyca15, line 17: I know I suggested this name, but I think `MatchesAll` would be better. Also initializing boolean to `false` is unnecessary.


```suggestion
        public bool MatchesAll { get; private set; }
```

jimmyca15: Also, please put properties after constructor.

src/Microsoft.FeatureManagement/FeatureFilters/Cron/CronField.cs
```diff
@@ -0,0 +1,241 @@
...
+            _kind = kind;
+
+            (_minValue, _maxValue) = kind switch
+            {
+                CronFieldKind.Minute => (0, 59),
+                CronFieldKind.Hour => (0, 23),
+                CronFieldKind.DayOfMonth => (1, 31),
+                CronFieldKind.Month => (1, 12),
+                CronFieldKind.DayOfWeek => (0, 7),
+                _ => throw new ArgumentException("Invalid crontab field kind.")
```
jimmyca15, line 39: Should have param name specified.

src/Microsoft.FeatureManagement/FeatureFilters/Cron/CronExpression.cs
```diff
@@ -0,0 +1,99 @@
...
+        public static CronExpression Parse(string expression)
+        {
+            if (expression == null)
+            {
+                throw new ArgumentNullException(nameof(expression));
+            }
+
+            string InvalidCronExpressionErrorMessage = $"The provided Cron expression: '{expression}' is invalid.";
+
+            string[] fields = expression.Split(WhitespaceDelimiters, StringSplitOptions.RemoveEmptyEntries);
```
jimmyca15, line 46: Please be consistent about new line usage. The line above is separated by a blank line. Below it is not. I suggest we stick with adding blank line between statements.

src/Microsoft.FeatureManagement/FeatureFilters/Cron/CronExpression.cs
```diff
@@ -0,0 +1,99 @@
...
+        /// <returns>A parsed CronExpression.</returns>
+        public static CronExpression Parse(string expression)
+        {
+            if (expression == null)
+            {
+                throw new ArgumentNullException(nameof(expression));
+            }
+
+            string InvalidCronExpressionErrorMessage = $"The provided Cron expression: '{expression}' is invalid.";
+
```
jimmyca15, line 45: I'd like if we can avoid defining white space characters and instead use [IsWhiteSpace](https://learn.microsoft.com/en-us/dotnet/api/system.char.iswhitespace?view=net-7.0#system-char-iswhitespace(system-char))

I suggest

```
var fields = new string[NumberOfFields];

int i = 0, pos = 0;

while (i < expression.Length)
{
    if (char.IsWhiteSpace(expression[i]))
    {
        continue;
    }

    if (pos >= fields.Length)
    {
        //
        // Invalid length
    }

    //
    // Grab fields here

    pos++;
}
```

src/Microsoft.FeatureManagement/FeatureFilters/Cron/CronExpression.cs
```diff
@@ -0,0 +1,99 @@
...
+        /// <summary>
+        /// Checks whether the Crontab expression is satisfied by the given timestamp.
+        /// </summary>
+        /// <param name="time">The timestamp to check.</param>
+        /// <returns>True if the Crontab expression is satisfied by the give timestamp, otherwise false.</returns>
+        public bool IsSatisfiedBy(DateTimeOffset time)
+        {
+            /*
+            The curent time is said to be satisfied by the Crontab when the 'minute', 'hour', and 'month of the year' fields match the current time,
```
jimmyca15, line 74: Let's replace all occurence of cron tab with cron expression.

src/Microsoft.FeatureManagement/FeatureFilters/Cron/CronExpression.cs
```diff
@@ -0,0 +1,99 @@
...
+                            && _month.Match((int)time.Month);
+            }
+            else
+            {
+                isDayMatched = _dayOfMonth.Match((int)time.Day)
+                            && _dayOfWeek.Match((int)time.DayOfWeek)
+                            && _month.Match((int)time.Month);
+            }
+
+            bool isTimeSpanMatched = _hour.Match((int)time.Hour) && _minute.Match((int)time.Minute);
```
jimmyca15, line 92:
```suggestion
            return isDayMatched &&
                _hour.Match((int)time.Hour) &&
                _minute.Match((int)time.Minute);
```

src/Microsoft.FeatureManagement/FeatureFilters/Cron/CronField.cs
```diff
@@ -0,0 +1,241 @@
...
+                {
+                    return false;
+                }
+            }
+
+            result = cronField;
+            return true;
+        }
+
+        private int TryGetNumber(string str)
```
jimmyca15, line 169: A method `TryGetNumber` should be `private bool TryGetNumber(string value, out int number)`

src/Microsoft.FeatureManagement/FeatureFilters/Cron/CronField.cs
```diff
@@ -0,0 +1,241 @@
...
+            // Step values can be used in conjunction with ranges after a slash.
+            string[] segments = content.Split(',');
+            foreach (string segment in segments)
+            {
+                if (segment == null)
+                {
+                    return false;
+                }
+
+                int value = cronField.TryGetNumber(segment);
```
jimmyca15, line 96: I'm not a fan of calling private method on object. Can we avoid that?

src/Microsoft.FeatureManagement/FeatureFilters/Cron/CronField.cs
```diff
@@ -0,0 +1,241 @@
...
+                        cronField.MatchAll = true;
+                    }
+
+                    string[] parts = segment.Split('/');
+                    if (parts.Length > 2) // multiple slashs
+                    {
+                        return false;
+                    }
+
+                    int step = parts.Length == 2 ? cronField.TryGetNumber(parts[1]) : 1;
```
jimmyca15, line 116: All try should return bool

src/Microsoft.FeatureManagement/FeatureFilters/Cron/CronExpression.cs
zhiyuanliang-ms: Ah, my bad. I forget to replace crontab everywhere.

src/Microsoft.FeatureManagement/FeatureFilters/Cron/CronField.cs
zhiyuanliang-ms: Yes. I can pass the field kind as an additional parameter and make TryGetNumber to be a static private method.

src/Microsoft.FeatureManagement/FeatureFilters/Cron/CronExpression.cs
zhiyuanliang-ms: Sounds good!

src/Microsoft.FeatureManagement/FeatureFilters/Cron/CronExpression.cs
```diff
@@ -0,0 +1,121 @@
...
+                }
+
+                int start = i; // Start of a field
+
+                while (i < expression.Length && !char.IsWhiteSpace(expression[i]))
+                {
+                    i++;
+                }
+
+                fields[pos] = expression.Substring(start, i - start);
```
jimmyca15, line 70: Even better if we can use `ReadOnlySpan<char>` to avoid allocating new strings.

zhiyuanliang-ms: CronField.TryParse(CronFieldKind kind, string content, out CronField result) will take a string as parameter. This method will also call string.Split() and string.Contains(). Should I change them and use ReadOnlySpan<char> instead?

README.md
```diff
@@ -489,6 +489,52 @@ This filter provides the capability to enable a feature based on a time window.
...
+        {
+            "Name": "Microsoft.TimeWindow",
+            "Parameters": {
+                "Start": "2023-09-06T00:00:00+08:00",
+                "Filters": [
+                    // 18:00-20:00 on weekdays
+                    "* 18-19 * * Mon-Fri",
+                    // 18:00-22:00 on weekends
+                    "* 18-21 * * Sat,Sun"
+                ]
```
zhenlan, line 506: Do we really have to support an array of crontab expressions, or is one sufficient for most scenarios? Even if someone really needs more than one, they can always add more Time Window filters with different configurations.

I'm concerned this complicates the usage of the recurring Time Window filter. I can already see this complicates the UI design.

BTW, having a 'Filters' parameter in a feature filter is also confusing.

zhiyuanliang-ms: The scenario that turn on a feature flag in 18:00-20:00 on every weekend and 18:00-22:00 on every national holiday requires a list to specify recurring time windows with different duration.

zhiyuanliang-ms: I agree that the parameter name 'Filters' is kind of confusing. But the cron expression does serve as "filters" since we also have the parameter "Start" and "End" which should be considered as the timestamp when the filter starts to work("FilterStart") and when the filter ends working("FilterEnd").
The reason why we use "Filters" instead of "Recurrence" is that we will treat an empty list of "Filters" as not specified.
When "Filters" is not specifed, the TimeWindowFilter will revert to the original fixed time window filter.
The name "Filters" makes more sense.

Let's take a look at two choices:
Choice A:
"Start" : "2023-9-19-00:00",
"End": "2024-9-19-00:00",
"Filters": [ ]

Choice B:
"Start" : "2023-9-19-00:00",
"End": "2024-9-19-00:00",
"Recurrence": [ ]
If "Filters" or "Recurrence" is not specified, there is no confusion.
If "Filters" or "Recurrence" is not empty, the filter will only enable the feature during the recurring time window listed in the list "Filters" or "Recurrence". However, for the “Recurrence" choice, the parameter "Start" and "End" serve as the "RecurrenceStart" and "RecurrentEnd" instead of "FilterStart" and "FilterEnd". This will make the scenario that the "Recurrence" is specified as an empty list confusing. The function of "Start" and "End" will be inconsistent if we use name "Recurrence".
The user who uses FeatureManagement-Dotnet is expected to read the README before using it. The user who uses FeatureManagement through portal will never know the backend data schema for TimeWindowFilter unless they use "Advanced Edit"(in this case, the user should be considered as a pro who has read our doc). On the Portal, the scenarios of setting a fixed time window and setting recurring time window will have different UI and will not mention the "Filters" parameter.

For me, I will vote for the name "Filters" because it makes the role of "Start" and "End" parametes consistent. I also believe it will not cause confusion for the user who uses FeatureManagement-Dotnet lib directly.

zhenlan: > The scenario when the user want to activate a feature flag during 18:00-20:00 on every weekend and during 18:00-22:00 on holidays requires a list to specify recurring time windows with different durations. 

I'm sure we can come up with even more complex scenarios. The question is how realistic they are. Outlook doesn't support infinite recurrence patterns, just one pattern at a time. Without the array, the parameters can be simplified to "Start", "End" and "RecurrencePattern". The UI design doesn't need an add button for more than one recurrence pattern anymore.

> Using multiple timewindow filters may not be a good choice, since the feature flag may also use other filters(e.g. targeting), in this case, the "RequirementType" is "All".

We know the "RequirementType" doesn't support all possible logical operations. It was a conscious decision. It's a balance between complexity and the most common scenarios. The choice of that design shouldn't be the reason for us to complicate the design of each feature filter. We have a reasonable workaround. We know there are scenarios that won't be supported out-of-box no matter how. That's where customers should consider custom filters.

zhiyuanliang-ms: If we do not want to use list for "Recurrence", then we cannot use cron expression any more. The reason is that we have to use at least three cron expressions to describe the time span 18:30`20:30. e.g. "30-59 18 * * *", "* 19 * * *“ and "0-29 20 * * *".
I have proposed a lot of schema design to describe the recurrence. Some of them are deprecated early and are not included in the design doc.
The best way to describe recurrence is to specify the duration and recurrence pattern (i.e. when we want to make the recuurence happen).
At first, I want to use two parameters "Duration" and "RecurWhen". 
'''javascript
"TimeWindowDuration": { 
  "Type": "Hour", 
  "Length": 4 
},  


"RecurWhen": [
 "0 18 * * * 5"
]
```

Comment (zhenlan) [reply README.md:506]: I'm not sure I follow what you said here. I did NOT suggest we abandon the usage of cron expression. I am only suggesting we support one cron expression per Time Window filter instead of an array of them. Here is an example.

``` JavaScript
"EnhancedPipeline": {
    "EnabledFor": [
        {
            "Name": "Microsoft.TimeWindow",
            "Parameters": {
                "Start": "2023-09-06T00:00:00+08:00",
                "RecurrencePattern": "* 18-19 * * Mon-Fri" // 18:00-20:00 on weekdays
            }
        }
    ]
}
```

src/Microsoft.FeatureManagement/FeatureFilters/Cron/CronFieldKind.cs
```diff
@@ -0,0 +1,41 @@
...
+
+        /// <summary>
+        /// Field Name: Month
+        /// Allowed Values: 1-12 (or use the first three letters of the month name)
+        /// </summary>
+        Month,
+
+        /// <summary>
+        /// Field Name: Day of week
+        /// Allowed Values: 0-7 (0 or 7 is Sunday, or use the first three letters of the day name)
```
zhenlan, line 37: > 0 or 7 is Sunday

Is this from the standard?

zhiyuanliang-ms: This is according to the linux man page https://man7.org/linux/man-pages/man5/crontab.5.html
Some cron implementations may not allow 0 to represent Sunday.

README.md
zhiyuanliang-ms: What if I want the recurrence happen during 18:30`20:30 every day?
You cannot represent 18:30-20:30 in a single cron.

zhiyuanliang-ms:
Update:
We will re-think the design. Current design will inevitably result in the portal UI becoming complex.

README.md
juniwang: > > The scenario when the user want to activate a feature flag during 18:00-20:00 on every weekend and during 18:00-22:00 on holidays requires a list to specify recurring time windows with different durations.
> 
> I'm sure we can come up with even more complex scenarios. The question is how realistic they are. Outlook doesn't support infinite recurrence patterns, just one pattern at a time. Without the array, the parameters can be simplified to "Start", "End" and "RecurrencePattern". The UI design doesn't need an add button for more than one recurrence pattern anymore.
> 
> > Using multiple timewindow filters may not be a good choice, since the feature flag may also use other filters(e.g. targeting), in this case, the "RequirementType" is "All".
> 
> We know the "RequirementType" doesn't support all possible logical operations. It was a conscious decision. It's a balance between complexity and the most common scenarios. The choice of that design shouldn't be the reason for us to complicate the design of each feature filter. We have a reasonable workaround. We know there are scenarios that won't be supported out-of-box no matter how. That's where customers should consider custom filters.

Hi @zhenlan, I thinks scenario like `18:00-20:00 everyday` is a very common scenario, not something that rarely happens. And current design does not allow us to use one single cron to describe it. Since they are closed tied together, adding two cron expressions makes more sense to me than adding two TimeWindow filters.  If we don't support cron array, probablly customers will have to add more Feature filters, which means inceasing size/complexity in another level.

zhenlan: @juniwang Yes, we chatted offline, and I understand this now. This limitation imposes challenges on our portal to properly break down `18:30-20:30` into 3 crontabs and reconstruct it back from crontabs to a human-readable time duration. We need to rethink our design choices.

rossgrambo:
We spoke about requiring the `start` to be within the recurring `range.start` to `range.end` window.

If we do that, `range.start` is not needed. Unless the primary `start` and `end` aren't required to be the first one...?

zhiyuanliang-ms:
My thought is that the "Start" and "End" define a time window, which is also the first ocurrence.
This primary time window will provide two information for the recurrence:
The start time of the time window and the duration of the time window.
The "Recurrence" specifies the date/day when the recurrence of the time window will happen.

For me, I think the primary time window is unnecessary to follow the recurrence pattern and range.
For example, we have set a time window on 9.22 Friday 18:00-20:00 and let it repeat on every Monday start from 2023-9-31.

My impletation for the validation of the parameter "Recurrence.Range.StartDate" will be like
`private static bool TryValidateStartDate(TimeWindowFilterSettings settings, out string paramName, out string reason)
{
    Recurrence recurrence = settings.Recurrence;

    paramName = $"{nameof(settings.Recurrence)}.{nameof(recurrence.Range)}.{nameof(recurrence.Range.StartDate)}";

    reason = null;

    if (recurrence.Range.StartDate == null)
    {
        reason = RequiredParameter;

        return false;
    }

    if (recurrence.Range.StartDate < settings.Start)
    {
        reason = OutOfRange;

        return false;
    }

    return true;
}`

zhenlan:
> For example, we have set a time window on 9.22 Friday 18:00-20:00 and let it repeat on every Monday start from 2023-9-31。

For the discussion, in this example (sorry, there is no 9/31 :) so I change it to 9/25)
```
settings.Start = 9/22 Friday 18:00
settings.End = 9/22 Friday 20:00
recurrence.Range.StartDate = 9/25 Monday 12:00am
recurrence.Range.EndDate = null
recurrence.type = weekly, Monday
```

IMO, we have two options
1. We use `settings.Start/End` to calculate the start time and duration and throw away the date part afterwards, so the first occurrence is 9/25 Monday 18:00-20:00
2. The `settings.Start/End` must be the first occurrence that matches the pattern, so this is treated as invalid input. Portal will show error; runtime will either throw or treat it as disabled.

I don't like treating `settings.Start/End` as the first occurrence regardless and then calculating the following occurrences (likely still based on option 1 above).

jimmyca15:
@zhenlan


date ends up being important if you have an event that spans across midnight.

rossgrambo:
> date ends up being important if you have an event that spans across midnight.

It's important to be defined on the field. But if you're extracting duration from it as Zhenlan suggested, you don't need date anymore, as start + duration tells you everything.

But regardless, I think @zhiyuanliang-ms 's version makes the most sense. 
1. TimeWindow start and end defines a time when we will return true. (Already the state of things)
2. Completely separately, here's a way to setup recurrence. We will use the start time and end time of the time window.

If we really want to support hourly recurrance then bullet 2 would need to change. But without that, I think that's the cleanest way to do it.

jimmyca15:
@rossgrambo that's how I imagine it working.

zhenlan:
Users shouldn't have a recurring time window that has any occurrence outside the range.

jimmyca15:
@zhiyuanliang-ms 

Zhenlan, Ross, and I discussed this. We came up with the following points.

* We don't need `recurrence.range.startDate`
It's already covered by our time window `start` field. We just mandate that the `start` date represents the first occurrence.

* `start` has to be a valid occurrence within the recurrence pattern. After all, it marks the first occurrence.

* Some extra validation we discussed -- The date part of `start` has to be strictly less than `endDate`

zhiyuanliang-ms:
@jimmyca15 Got it.

zhiyuanliang-ms:
@rossgrambo @zhenlan @jimmyca15 Since we have decided to remove "Recurrence.Range.StartDate", I am wondering whether we still need "Recurrence.Range.RecurrenceTimeZone". We can mandate that the time zone of "Start" will be applied to the recurrence.

zhiyuanliang-ms:
I noticed that the current behavior of Portal is that the "Start" and "End" will be converted into GMT format string rather than using the time zone selected by the user. The time zone information will not be provided in "Start" or "End". 
It doesn't matter for the fixed time window. But for the recurring time window,  time zone matters. So I guess we may still need the "RecurrenceTimeZone". 

If the "RecurrenceTimeZone" is specified, it will be applied to the "Start" and then we will check whether the "Start" timestamp is valid for the Recurrence.Pattern under the "RecurrenceTimeZone".
If the "RecurrenceTimeZone" is not specified, the UTC time offset of "Start" will be applied to the recurrence.

Besides, I believe Portal should provide the time zone information in the datetime string by using format like"2023-09-23T08:00+08:00" rather than "Sat, 23 Sep 2023 00:00:00 GMT", since we allow user to select the timezone on the Portal.

zhiyuanliang-ms:
> I noticed that the current behavior of Portal is that the "Start" and "End" will be converted into GMT format string rather than using the time zone selected by the user. The time zone information will not be provided in "Start" or "End".

I think the current behavior of Portal will break the recurrence pattern.
The recurrence pattern only defines on which day/date the recurrrence will occur, the start time information is provided by the "Start".

If user selects 2023-9-24 6:00-8:00 and time zone +08:00, this time zone will be converted into "Sat, 23 Sep 2023 22:00:00 GMT to "Sun, 24 Sep 2023 00:00:00 GMT".
In this case, the "RecurrenceTimeZone" should be a required parameter instead of an optional one.

zhiyuanliang-ms:
image
The parameter "FirstDayOfWeek" looks redundant for me. I cannot figure out what it is used for.
image
Any idea about this?
I plan to remove this parameter.

rossgrambo:
> The parameter "FirstDayOfWeek" looks redundant for me

It allows you to pick what day is the first day of the week. If you pass sunday, nothing changes (at least that's what i'm used to). If you pass monday, then the value in index might be affected. For example:

```
// Lets say the date is Sun, Oct 1st.

DaysOfWeek: "Sunday", "Wednesday"
FirstDayOfWeek: "Monday",
Index: "First"
```
With this, the "first" week is considered ending on Sun, Oct 1st. Mon, Oct 2nd would be the "second" week. Europe generally starts weeks on Monday while the US starts weeks on Sunday. 

I think we should include it.

> In this case, the "RecurrenceTimeZone" should be a required parameter instead of an optional one.

I'm a little lost here. If we use the Start time zone and Sept 23rd would be "converted" to Sept 24th, that's fine. There's no actual conversion. If they use UTC and sept 23rd, that's the start. If they use +08:00, and pick sept 23rd, then that's the start instead. These are different times in UTC.

This means:
```
"Start": "2023-9-24 6:00 +8:00",
"Recurrance": {
  "dayOfMonth": "25"
}
```

Would repeat at "Sat, 25 Sep 2023 06:00:00 GMT+0800" / "Sun, 24 Sep 2023 22:00:00 GMT"

Let me know if I'm missing something

zhiyuanliang-ms:
image
@rossgrambo 

> // Lets say the date is Sun, Oct 1st.
> 
> DaysOfWeek: "Sunday", "Wednesday"
> FirstDayOfWeek: "Monday",
> Index: "First"
This will be translated to the first Sunday or Wednesday of the month. From the above example, I guess the "Index" describes the day of week directly instead of the week. And "FirstDayOfWeek" is only required for type "Weekly, not for "RelativeMonthly" or "RelativeYearly".

zhiyuanliang-ms:
> I'm a little lost here. If we use the Start time zone and Sept 23rd would be "converted" to Sept 24th, that's fine. There's no actual conversion. If they use UTC and sept 23rd, that's the start. If they use +08:00, and pick sept 23rd, then that's the start instead. These are different times in UTC.

> Would repeat at "Sat, 25 Sep 2023 06:00:00 GMT+0800" / "Sun, 24 Sep 2023 22:00:00 GMT"
> Let me know if I'm missing something

"Sat, 25 Sep 2023 06:00:00 GMT+0800" and "Sun, 24 Sep 2023 22:00:00 GMT" are the same timestamp, they are representing the same physical time. 

Currently, no matter what time zone you choose on the Portal, the timestamp will be converted into GMT time(which is under UTC+0:00)
For the fixed time window, the current behavior of Portal will not affect how the filter works. The fixed time window is still the same.

For the recurrence, according to Jimmy's comment, we should add this check for the "Start".
> start has to be a valid occurrence within the recurrence pattern. After all, it marks the first occurrence.
It is possible that under UTC+08:00 the "Start" is on Monday and ,when it is converted to GMT time, it becomes Sunday.
The user sets the recurrence Pattern as "Weekly on Monday" The "Start" will not be a valid first occurrence.
The feature management library cannot know the time zone information from the GMT format "Start" and "End" if the "RecurrenceTImeZone" is missing.

rossgrambo:
I see, so it's a concern for the portal, as it's already doing the conversion, but you're worried the rest of the settings won't be accurate anymore.

I think there's 3 options:
1. Stop the portal from converting to GMT, as we now care more about the timezone
2. Convert the other settings as well. If the GMT conversion changes the date, in the settings, -1/+1 from the daysOfMonth & daysOfWeeks. 
    1. We'd have to support "-1" for days like cron
    2. I think there'd be some weirdness on Feb 29th, but haven't thought it through.
3. Add the recurrence time zone. This would be applied to the date and then checked for matching the recurrence. 

I'd prefer 1, but we can go with 3 if we're scared of the impact of 1.

zhiyuanliang-ms:
I like the option 1. But we need to investigate whether there is a format with the time zone info which can be parsed by spring, dotnet and javascript.

rossgrambo:
Closing this PR for now in favor of https://github.com/microsoft/FeatureManagement-Dotnet/pull/266

PR #260, opened 2023-09-12
Update FeatureGatedAsyncActionFilter.cs

WeihanLi:
use `context.HttpContext.RequestServices` directly


zhiyuanliang-ms:
@jimmyca15 It seems like this change is reasonable. Should this PR be approved?

jimmyca15 (approved):

zhiyuanliang-ms:
@WeihanLi This PR has been approved. Feel free to merge it.

WeihanLi:
@zhiyuanliang-ms thanks, but I think I do not have permission to merge and seemed the status is not updated which also block merge, could you please help check, thanks very much

Issue #254, opened 2023-08-22
Potential bug: Propagating App Service environment variables through IFeatureManager


aatrisgn:
Hi,

I am experiencing some weird behaviour utilizing IFeatureManager on a regular Linux S1 App Service with Azure App Configuration integration via managed identity.

In short, it seems like IFeatureManager propagates some configuration of our App Service instance when no Feature Toggles are defined in Azure App Configuration for our application.

We have two scenarios for our usage:
1. An application has one or more feature flags enabled in Azure App Configuration
2. An application has  no feature flags enabled in Azure App Configuration

We are exposing the feature flags via a configurations endpoint for our frontend application. Our feature flags are not sensitive information, but only for UX.

Our feature flags will only be defined in Azure App Configuration for deployed Azure environments. Locally, we handle it differently.

**Problem**
What we experience are the following three things:
1. An environment has one or more feature flags enabled:
![image](https://github.com/microsoft/FeatureManagement-Dotnet/assets/49903629/be4ffe48-e102-4631-b89f-ce7abfc2b72f)

Feature flag is correctly exposed to our frontend.

2. An environment has disabled feature flags:
![image](https://github.com/microsoft/FeatureManagement-Dotnet/assets/49903629/7d5a9864-f007-4e59-a3b3-cd0ac5e15ee4)

We do not expose any feature flags - This is fine behaviour as well.

3. We have not deployed any feature flags (**Where the problem lies**):
![image](https://github.com/microsoft/FeatureManagement-Dotnet/assets/49903629/bfcb671c-576f-43ed-b3b2-7343d4845a32)

This is problematic. We do not want to expose what seems to look like App Service configuration for our feature flags. It seems weird to me, that this is exposed via IFeatureManager. I can read in the documentation that IFeatureManager exposes environment settings as well, but I wouldn't expect it to do so, unless explicitly defined.

Our application code for Azure App Configuration looks like the following:

    webApplicationBuilder.Services.AddFeatureManagement();
    
    if (appConfigEndpoint != null) {
        webApplicationBuilder.Configuration.AddAzureAppConfiguration(options =>
            options.Connect(
                new Uri(appConfigEndpoint),
                new ManagedIdentityCredential(managedIdentityId))
        .UseFeatureFlags(featureFlagOptions => featureFlagOptions.CacheExpirationInterval = TimeSpan.FromMinutes(5)));
    }

We expose our Feature Flags to our frontend via the following:

[...]

Issue #144, opened 2021-09-17
Empty appsettings.json `FeatureManagement` section returns internal `GetFeatureNamesAsync`


simonauner:
Not sure if the title reflects the issue correctly, but I have a dotnet app that uses Microsoft.FeatureManagement 2.3.0.

It has this configured in its `appsettings.json`:

```json
  "FeatureManagement": {
    "MyFeature": false
  },
```

And then I thought it would be helpful to have a logger that logs all enabled or disabled features upon startup:

```csharp
    public class FeatureLogger : IFeatureLogger
    {
        private readonly IFeatureManager _featureManager;
        private readonly ILogger _logger;

        public FeatureLogger(IFeatureManager featureManager, ILogger logger)
        {
            _featureManager = featureManager;
            _logger = logger.ForContext<FeatureLogger>();
        }

        public async Task LogFeatureStatus()
        {
            var features = _featureManager.GetFeatureNamesAsync();
            await foreach (var feature in features)
            {
                var isEnabled = await _featureManager.IsEnabledAsync(feature);
                _logger.Information("Feature {Feature} is {Status}", feature, isEnabled ? "enabled" : "disabled");
            }
        }
    }
```

It works as expected as long as I have something in the `FeatureManagement` section in `appsettings.json`:

```console
[10:41:32 INF] Feature MyFeature is disabled
```

But if I decide to remove this `MyFeature` prop from `FeatureManagement` in `appsettings.json`:

```json
  "FeatureManagement": {
  },
```

 then the following is logged:

```console
[10:42:57 INF] Feature applicationName is disabled
[10:42:57 INF] Feature ASPNETCORE_ENVIRONMENT is disabled
[10:42:57 INF] Feature ASPNETCORE_URLS is disabled
[10:42:57 INF] Feature BAMF_DESKTOP_FILE_HINT is disabled
[10:42:57 INF] Feature contentRoot is disabled
[10:42:57 INF] Feature DBUS_SESSION_BUS_ADDRESS is disabled
[10:42:57 INF] Feature DESKTOP_SESSION is disabled
[10:42:57 INF] Feature DISPLAY is disabled
[10:42:57 INF] Feature DOTNET_ROOT is disabled
[10:42:57 INF] Feature ENVIRONMENT is disabled
[10:42:57 INF] Feature GDMSESSION is disabled
[10:42:57 INF] Feature GIO_LAUNCHED_DESKTOP_FILE is disabled
[...]

PR #261, opened 2023-09-13
Adjusts the configuration provider to no longer check top level for feature flags

rossgrambo:
With the release of v3 on the horizon, we can start adding misc. features that were waiting for a major version bump. 

This adjustment was requested in a couple issues
https://github.com/microsoft/FeatureManagement-Dotnet/issues/144
https://github.com/microsoft/FeatureManagement-Dotnet/issues/254

### DIFF
--- a/src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
+++ b/src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
@@ -34,7 +34,7 @@ namespace Microsoft.FeatureManagement
         public ConfigurationFeatureDefinitionProvider(IConfiguration configuration, ILoggerFactory loggerFactory)
         {
             _configuration = configuration ?? throw new ArgumentNullException(nameof(configuration));
-            _logger = loggerFactory.CreateLogger<ConfigurationFeatureDefinitionProvider>();
+            _logger = loggerFactory?.CreateLogger<ConfigurationFeatureDefinitionProvider>() ?? throw new ArgumentNullException(nameof(loggerFactory));
             _definitions = new ConcurrentDictionary<string, FeatureDefinition>();
 
             _changeSubscription = ChangeToken.OnChange(
--- a/src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
+++ b/src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
@@ -31,8 +31,6 @@ namespace Microsoft.FeatureManagement
         private readonly ILogger _logger;
         private int _stale = 0;
 
-        const string ParseValueErrorString = "Invalid setting '{0}' with value '{1}' for feature '{2}'.";
-
         public ConfigurationFeatureDefinitionProvider(IConfiguration configuration, ILoggerFactory loggerFactory)
         {
             _configuration = configuration ?? throw new ArgumentNullException(nameof(configuration));
--- a/src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
+++ b/src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
@@ -213,12 +213,10 @@ namespace Microsoft.FeatureManagement
             {
                 return featureManagementConfigurationSection.GetChildren();
             }
-            else
-            {
-                _logger.LogWarning($"No configuration section named '{FeatureManagementSectionName}' was found.");
 
-                return Enumerable.Empty<IConfigurationSection>();
-            }
+            _logger.LogDebug($"No configuration section named '{FeatureManagementSectionName}' was found.");
+
+            return Enumerable.Empty<IConfigurationSection>();
         }
     }
 }
--- a/src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
+++ b/src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT license.
 //
 using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Primitives;
 using System;
 using System.Collections.Concurrent;
@@ -23,14 +24,19 @@ namespace Microsoft.FeatureManagement
 
         private const string FeatureFiltersSectionName = "EnabledFor";
         private const string RequirementTypeKeyword = "RequirementType";
+        private const string FeatureManagementSectionName = "FeatureManagement";
         private readonly IConfiguration _configuration;
         private readonly ConcurrentDictionary<string, FeatureDefinition> _definitions;
         private IDisposable _changeSubscription;
+        private readonly ILogger _logger;
         private int _stale = 0;
 
-        public ConfigurationFeatureDefinitionProvider(IConfiguration configuration)
+        const string ParseValueErrorString = "Invalid setting '{0}' with value '{1}' for feature '{2}'.";
+
+        public ConfigurationFeatureDefinitionProvider(IConfiguration configuration, ILoggerFactory loggerFactory)
         {
             _configuration = configuration ?? throw new ArgumentNullException(nameof(configuration));
+            _logger = loggerFactory.CreateLogger<ConfigurationFeatureDefinitionProvider>();
             _definitions = new ConcurrentDictionary<string, FeatureDefinition>();
 
             _changeSubscription = ChangeToken.OnChange(
@@ -199,17 +205,19 @@ namespace Microsoft.FeatureManagement
 
         private IEnumerable<IConfigurationSection> GetFeatureDefinitionSections()
         {
-            const string FeatureManagementSectionName = "FeatureManagement";
+            //
+            // Look for feature definitions under the "FeatureManagement" section
+            IConfigurationSection featureManagementConfigurationSection = _configuration.GetSection(FeatureManagementSectionName);
 
-            if (_configuration.GetChildren().Any(s => s.Key.Equals(FeatureManagementSectionName, StringComparison.OrdinalIgnoreCase)))
+            if (featureManagementConfigurationSection.Exists())
             {
-                //
-                // Look for feature definitions under the "FeatureManagement" section
-                return _configuration.GetSection(FeatureManagementSectionName).GetChildren();
+                return featureManagementConfigurationSection.GetChildren();
             }
             else
             {
-                return _configuration.GetChildren();
+                _logger.LogWarning($"No configuration section named '{FeatureManagementSectionName}' was found.");
+
+                return Enumerable.Empty<IConfigurationSection>();
             }
         }
     }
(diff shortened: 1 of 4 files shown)


src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
zhiyuanliang-ms, line 1: Hi, Ross. I have discussed this issue with Jimmy on Tuesday. I intended to submit a PR for this today, but it turns out that it's quite a coincidence that you've already submitted a PR for it. 
What you've implemented is a breaking change for the user, so we should leave it for the release 3.0.
Currently, Jimmy thought we should provide a flag on FeatureManagementOptions for 2.x. I can create a PR for this change.

src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
```diff
@@ -201,15 +201,17 @@ private IEnumerable<IConfigurationSection> GetFeatureDefinitionSections()
...
             {
-                //
-                // Look for feature definitions under the "FeatureManagement" section
-                return _configuration.GetSection(FeatureManagementSectionName).GetChildren();
+                return featureManagementConfigurationSection.GetChildren();
             }
             else
             {
-                return _configuration.GetChildren();
+                return Enumerable.Empty<IConfigurationSection>();
```
zhiyuanliang-ms, line 214: Maybe we can log some warning if there is no FeatureManagement section.

src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
```diff
@@ -201,15 +201,17 @@ private IEnumerable<IConfigurationSection> GetFeatureDefinitionSections()
         {
             const string FeatureManagementSectionName = "FeatureManagement";
```
zhiyuanliang-ms, line 202: ConfigurationFeatureDefitionProvider have these two property
private const string FeatureFiltersSectionName = "EnabledFor";
private const string RequirementTypeKeyword = "RequirementType";
It looks odd for me to have this const string in GetFeatureDefinitionSections().
I think this PR will be a good chance to change it into a const string property for the class. How do you think?

src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
rossgrambo: I figured it was a big enough breaking change to leave it out of v2 entirely. So far, this is the only breaking change of v3 so it's not unreasonable to recommend upgrading if they want. 

Personally, I think it's cleaner to leave things as they are, as the flag isn't one I want around. We could if we think it's important to let users stay on v2.

jimmyca15: > as the flag isn't one I want around

The flag would be removed in v3 even if added in v2. It's existence unblocks stable v2 while we work on v3, which I expect to be in preview for a while.

jimmyca15: Going forward, any v3 changes should target preview branch.

src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
amerjusupovic: This is a good point, but the preview branch actually has a change to use a ConfigurationFields.cs file with constants for each field as a result of the variants PR. That means this should no longer be an issue if we merge this into preview instead, like mentioned in the above comments.

src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
rossgrambo: Changed this to target preview as the v3 fix. v2 will need a separate PR. @zhiyuanliang-ms, do you want to do that one?

src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
rossgrambo: Now targeting preview and using the ConfiguartionFields file.

src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
zhiyuanliang-ms: Yes, I can do that one.

rossgrambo:
Now pointing this PR at main for a new major version.

jimmyca15:
The PR seems messed up. There is a huge change set.

rossgrambo:
My fault. Didn't push up the rebase. Should be good now.

src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
```diff
@@ -199,17 +205,19 @@ We support
...
             {
-                //
-                // Look for feature definitions under the "FeatureManagement" section
-                return _configuration.GetSection(FeatureManagementSectionName).GetChildren();
+                return featureManagementConfigurationSection.GetChildren();
             }
             else
             {
-                return _configuration.GetChildren();
+                _logger.LogWarning($"No configuration section named '{FeatureManagementSectionName}' was found.");
```
jimmyca15, line 218: There's just no features defined. That's fine. We have a separate flag to throw or not if a requested feature is not found.

rossgrambo: For developers, this code gets hit on each `IsEnabledAsync` call and on direct calls to `GetFeatureDefinitionAsync`.

For `IsEnabledAsync`, we do log that the feature wasn't found. I'd be okay to take it out, but I think there is some value in indicating that it never found a FeatureManagement section at all.

But for direct calls to `GetFeatureDefinitionAsync`, this is the only warning. I suppose the resulting object being empty is indication that something is wrong.

What if we kept it but adjusted it to `LogInformation`?

jimmyca15: LogDebug or LogTrace perhaps.

src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
```diff
@@ -199,17 +205,19 @@ We support
...
 
-            if (_configuration.GetChildren().Any(s => s.Key.Equals(FeatureManagementSectionName, StringComparison.OrdinalIgnoreCase)))
+            if (featureManagementConfigurationSection.Exists())
             {
-                //
-                // Look for feature definitions under the "FeatureManagement" section
-                return _configuration.GetSection(FeatureManagementSectionName).GetChildren();
+                return featureManagementConfigurationSection.GetChildren();
             }
             else
```
jimmyca15, line 216: else not needed since if statement returns.

src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
```diff
@@ -199,17 +205,19 @@ We support
...
+            // Look for feature definitions under the "FeatureManagement" section
+            IConfigurationSection featureManagementConfigurationSection = _configuration.GetSection(FeatureManagementSectionName);
 
-            if (_configuration.GetChildren().Any(s => s.Key.Equals(FeatureManagementSectionName, StringComparison.OrdinalIgnoreCase)))
+            if (featureManagementConfigurationSection.Exists())
             {
-                //
-                // Look for feature definitions under the "FeatureManagement" section
-                return _configuration.GetSection(FeatureManagementSectionName).GetChildren();
+                return featureManagementConfigurationSection.GetChildren();
```
jimmyca15, line 214: Can we call GetChildren even if the section doesn't exist? Just thinking that if so, we can shorten this whole thing to

```
return _configuration
    .GetSection(FeatureManagementSectionName)
    .GetChildren();
```

rossgrambo: I just checked, yes that works. However we won't use it here if we keep the log.

src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
rossgrambo: Adjusted to LogDebug

src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
jimmyca15: Okay, resolving.

src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
```diff
@@ -23,14 +24,19 @@ sealed class ConfigurationFeatureDefinitionProvider : IFeatureDefinitionProvider
...
         private const string RequirementTypeKeyword = "RequirementType";
+        private const string FeatureManagementSectionName = "FeatureManagement";
         private readonly IConfiguration _configuration;
         private readonly ConcurrentDictionary<string, FeatureDefinition> _definitions;
         private IDisposable _changeSubscription;
+        private readonly ILogger _logger;
         private int _stale = 0;
 
-        public ConfigurationFeatureDefinitionProvider(IConfiguration configuration)
+        const string ParseValueErrorString = "Invalid setting '{0}' with value '{1}' for feature '{2}'.";
```
jimmyca15, line 34: > ParseValueErrorString

What is this?

rossgrambo: Oops. It's part of the preview branch https://github.com/microsoft/FeatureManagement-Dotnet/blob/preview/src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs#L30 . Must have left it when changing the target here. Removing now

src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
```diff
@@ -23,14 +24,17 @@ sealed class ConfigurationFeatureDefinitionProvider : IFeatureDefinitionProvider
...
         private readonly ConcurrentDictionary<string, FeatureDefinition> _definitions;
         private IDisposable _changeSubscription;
+        private readonly ILogger _logger;
         private int _stale = 0;
 
-        public ConfigurationFeatureDefinitionProvider(IConfiguration configuration)
+        public ConfigurationFeatureDefinitionProvider(IConfiguration configuration, ILoggerFactory loggerFactory)
         {
             _configuration = configuration ?? throw new ArgumentNullException(nameof(configuration));
+            _logger = loggerFactory.CreateLogger<ConfigurationFeatureDefinitionProvider>();
```
jimmyca15, line 37: loggerFactory needs a null check.

jimmyca15 (approved):

jimmyca15 (approved):

jimmyca15 (approved):

jimmyca15 (approved):

jimmyca15 (approved):

PR #262, opened 2023-09-14
Allow different type of filters to share the same Alias

### LINKED ISSUE #98: Unable to create both IFeatureFilter and IContextualFeatureFilter with same FilterAlias.

zhiyuanliang-ms:
Allow filters to share the same Alias. This PR will resolve #98.
One FeatureFilter and multiple ContextualFeatureFilters using different types of context can share the same Alias.

We plan to register built-in filter including percentage, time window and contextual targeting filters by default when AddFeatureManagement() is called.
However, the contextual targeting and targeting filter share the same alias "Microsoft.Targeting". Currently, feature management does not allow filters share the same alias. This PR will also un-block this.

### DIFF
--- a/src/Microsoft.FeatureManagement/FeatureManager.cs
+++ b/src/Microsoft.FeatureManagement/FeatureManager.cs
@@ -357,7 +357,7 @@ namespace Microsoft.FeatureManagement
             }
 
             ContextualFeatureFilterEvaluator filter = _contextualFeatureFilterCache.GetOrAdd(
-                $"{filterName}{Environment.NewLine}{appContextType?.FullName}",
+                $"{filterName}{Environment.NewLine}{appContextType.FullName}",
                 (_) => {
 
                     IFeatureFilterMetadata metadata = GetFeatureFilterMetadata(filterName, appContextType);
--- a/src/Microsoft.FeatureManagement/FeatureManager.cs
+++ b/src/Microsoft.FeatureManagement/FeatureManager.cs
@@ -175,7 +175,7 @@ namespace Microsoft.FeatureManagement
 
                         //
                         // IContextualFeatureFilter
-                        if (useAppContext && !(filter is IFeatureFilter))
+                        if (useAppContext && ContextualFeatureFilterEvaluator.IsContextualFilter(filter, typeof(TContext)))
                         {
                             ContextualFeatureFilterEvaluator contextualFilter = GetContextualFeatureFilter(featureFilterConfiguration.Name, typeof(TContext));
 
--- a/src/Microsoft.FeatureManagement/FeatureManager.cs
+++ b/src/Microsoft.FeatureManagement/FeatureManager.cs
@@ -173,24 +173,28 @@ namespace Microsoft.FeatureManagement
                             Parameters = featureFilterConfiguration.Parameters
                         };
 
-                        if (filter is IFeatureFilter featureFilter)
-                        {
-                            BindSettings(filter, context, filterIndex);
+                        BindSettings(filter, context, filterIndex);
 
-                            if (await featureFilter.EvaluateAsync(context).ConfigureAwait(false) == targetEvaluation) {
+                        //
+                        // IContextualFeatureFilter
+                        if (useAppContext)
+                        {
+                            ContextualFeatureFilterEvaluator contextualFilter = GetContextualFeatureFilter(featureFilterConfiguration.Name, typeof(TContext));
+
+                            if (contextualFilter != null &&
+                                await contextualFilter.EvaluateAsync(context, appContext).ConfigureAwait(false) == targetEvaluation)
+                            {
                                 enabled = targetEvaluation;
 
                                 break;
                             }
                         }
-                        else
+
+                        //
+                        // IFeatureFilter
+                        if (filter is IFeatureFilter featureFilter)
                         {
-                            ContextualFeatureFilterEvaluator contextualFilter = GetContextualFeatureFilter(featureFilterConfiguration.Name, typeof(TContext));
-
-                            BindSettings(filter, context, filterIndex);
-
-                            if (contextualFilter != null &&
-                                await contextualFilter.EvaluateAsync(context, appContext).ConfigureAwait(false) == targetEvaluation)
+                            if (await featureFilter.EvaluateAsync(context).ConfigureAwait(false) == targetEvaluation)
                             {
                                 enabled = targetEvaluation;
 
@@ -357,6 +361,11 @@ namespace Microsoft.FeatureManagement
                 (_) => {
 
                     IFeatureFilterMetadata metadata = GetFeatureFilterMetadata(filterName, appContextType);
+
+                    if (metadata == null)
+                    {
+                        return null;
+                    }
 
                     return new ContextualFeatureFilterEvaluator(metadata, appContextType);
                 }
--- a/src/Microsoft.FeatureManagement/FeatureManager.cs
+++ b/src/Microsoft.FeatureManagement/FeatureManager.cs
@@ -25,8 +25,8 @@ namespace Microsoft.FeatureManagement
         private readonly IEnumerable<IFeatureFilterMetadata> _featureFilters;
         private readonly IEnumerable<ISessionManager> _sessionManagers;
         private readonly ILogger _logger;
-        private readonly ConcurrentDictionary<ValueTuple<string, Type>, IFeatureFilterMetadata> _filterMetadataCache;
-        private readonly ConcurrentDictionary<ValueTuple<string, Type>, ContextualFeatureFilterEvaluator> _contextualFeatureFilterCache;
+        private readonly ConcurrentDictionary<string, IFeatureFilterMetadata> _filterMetadataCache;
+        private readonly ConcurrentDictionary<string, ContextualFeatureFilterEvaluator> _contextualFeatureFilterCache;
         private readonly FeatureManagementOptions _options;
         private readonly IMemoryCache _parametersCache;
         
@@ -48,8 +48,8 @@ namespace Microsoft.FeatureManagement
             _featureFilters = featureFilters ?? throw new ArgumentNullException(nameof(featureFilters));
             _sessionManagers = sessionManagers ?? throw new ArgumentNullException(nameof(sessionManagers));
             _logger = loggerFactory.CreateLogger<FeatureManager>();
-            _filterMetadataCache = new ConcurrentDictionary<ValueTuple<string, Type>, IFeatureFilterMetadata>();
-            _contextualFeatureFilterCache = new ConcurrentDictionary<ValueTuple<string, Type>, ContextualFeatureFilterEvaluator>();
+            _filterMetadataCache = new ConcurrentDictionary<string, IFeatureFilterMetadata>();
+            _contextualFeatureFilterCache = new ConcurrentDictionary<string, ContextualFeatureFilterEvaluator>();
             _options = options?.Value ?? throw new ArgumentNullException(nameof(options));
             _parametersCache = new MemoryCache(new MemoryCacheOptions());
         }
@@ -146,11 +146,11 @@ namespace Microsoft.FeatureManagement
                         if (useAppContext)
                         {
                             filter = GetFeatureFilterMetadata(featureFilterConfiguration.Name, typeof(TContext)) ??
-                                     GetFeatureFilterMetadata(featureFilterConfiguration.Name, null);
+                                     GetFeatureFilterMetadata(featureFilterConfiguration.Name);
                         }
                         else
                         {
-                            filter = GetFeatureFilterMetadata(featureFilterConfiguration.Name, null);
+                            filter = GetFeatureFilterMetadata(featureFilterConfiguration.Name);
                         }
 
                         if (filter == null)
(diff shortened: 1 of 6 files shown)


zhiyuanliang-ms:
Tuple of filter name and context type will be used as the key for the filter metadata cache:
ConcurrentDictionary<ValueTuple<string, Type>, IFeatureFilterMetadata> _filterMetadataCache;
For non-contextual filter, the key will be ("FilterName", null).

No matter whether the feature flag contains any contextual filter, IsEnabledAsync<TContext>(string feature, TContext appContext) can be called to evaluate this feature flag.

One IFeatureFIlter and multiple IContextualFeatureFilters with different types of context can share one alias.

Let's say we have a non-contextual filter called FilterA and two contextual filters FilterB and FilterC which accept TypeB and TypeC contexts respectively. All of three filters share the same alias "SharedFilterName".
We also have a feature flag "MyFeature" which uses the feature filter "SharedFilterName" in its configuration.

If all of three filters are registered:
When we call IsEnabledAsync("MyFeature"), the FilterA will be used to evaluate the feature flag.
When we call IsEnabledAsync("MyFeature", context), if context's type is TypeB, FilterB will be used and if context's type is TypeC, FilterC will be used.

If only FilterA is registered:
When we call IsEnabledAsync("MyFeature", context), FilterA will be used.

rossgrambo:
Just noting this down for myself. For targeting context accessor, nothing changes. Because the filter's code controls what accessor/context it depends on. This means you wouldn't be able to have two filters with aliases "AdvancedTargetingFilter" if you wanted them to use accessors to get their context.

src/Microsoft.FeatureManagement/FeatureManager.cs
```diff
@@ -141,7 +141,17 @@ private async Task<bool> IsEnabledAsync<TContext>(string feature, TContext appCo
                             continue;
                         }
 
-                        IFeatureFilterMetadata filter = GetFeatureFilterMetadata(featureFilterConfiguration.Name);
+                        IFeatureFilterMetadata filter;
+
+                        if (useAppContext)
```
rossgrambo, line 146: If we're now saying a call to `IsEnabledAsync("name", Context);` will use "name" even if it has no context, then I don't think the `useAppContext` bool needs to exist anymore. 

Just use contextual filter if its defined for the context, otherwise use the no contextual filter.

src/Microsoft.FeatureManagement/FeatureManager.cs
```diff
@@ -165,7 +175,7 @@ private async Task<bool> IsEnabledAsync<TContext>(string feature, TContext appCo
 
                         //
                         // IContextualFeatureFilter
-                        if (useAppContext)
+                        if (useAppContext && ContextualFeatureFilterEvaluator.IsContextualFilter(filter, typeof(TContext)))
```
rossgrambo, line 178: Why is the `IsContextualFilter` function written like it is? Couldn't it just be `interfaceType.IsAssignableFrom(classType)` ? I know it's not a part of this PR, but curious.

src/Microsoft.FeatureManagement/FeatureManager.cs
```diff
@@ -267,48 +277,72 @@ private void BindSettings(IFeatureFilterMetadata filter, FeatureFilterEvaluation
...
+
+            if (name == null)
+            {
+                name = filterType.Name.EndsWith(filterSuffix, StringComparison.OrdinalIgnoreCase) ? filterType.Name.Substring(0, filterType.Name.Length - filterSuffix.Length) : filterType.Name;
+            }
+
+            //
+            // Feature filters can have namespaces in their alias
+            // If a feature is configured to use a filter without a namespace such as 'MyFilter', then it can match 'MyOrg.MyProduct.MyFilter' or simply 'MyFilter'
+            // If a feature is configured to use a filter with a namespace such as 'MyOrg.MyProduct.MyFilter' then it can only match 'MyOrg.MyProduct.MyFilter'
```
amerjusupovic, line 294: Out of curiosity, what if there is a filter `MyProduct.MyFilter` as well as `MyOrg.MyProduct.MyFilter`? We treat them as distinct names?

src/Microsoft.FeatureManagement/FeatureManager.cs
zhiyuanliang-ms: There is a staic method IsContextualFilter() of ContextualFeatureFilterEvaluator. 
In my previous commit, my code looked like: 
`if (useAppContext && ! (filter is IFeatureFilter))`
I think it is better to use IsContextualFilter() because it is more readable.

src/Microsoft.FeatureManagement/FeatureManager.cs
zhiyuanliang-ms: Yes. If the configuration of a feature flag contains any filters with a namespace, the feature manager will try to find a registered filter whose alias is the same as the filter name in the feature flag configuration.

src/Microsoft.FeatureManagement/FeatureManager.cs
zhiyuanliang-ms: Context may be null. I think we still need to distinguish the case that users intend to pass null as the context.

src/Microsoft.FeatureManagement/FeatureManager.cs
```diff
@@ -267,48 +277,72 @@ private void BindSettings(IFeatureFilterMetadata filter, FeatureFilterEvaluation
             context.Settings = settings;
         }
 
-        private IFeatureFilterMetadata GetFeatureFilterMetadata(string filterName)
+        private bool IsFilterNameMatched(Type filterType, string filterName)
```
jimmyca15, line 280: Needs to move after the private methods that call it.

src/Microsoft.FeatureManagement/FeatureManager.cs
```diff
@@ -267,48 +277,72 @@ private void BindSettings(IFeatureFilterMetadata filter, FeatureFilterEvaluation
...
+            {
+                //
+                // We take the simple name of a filter, E.g. 'MyFilter' for 'MyOrg.MyProduct.MyFilter'
+                string simpleName = name.Contains('.') ? name.Split('.').Last() : name;
+
+                return string.Equals(simpleName, filterName, StringComparison.OrdinalIgnoreCase);
+            }
+        }
+
+        private IFeatureFilterMetadata GetFeatureFilterMetadata(string filterName, Type appContextType)
```
jimmyca15, line 311:
```suggestion
        private IFeatureFilterMetadata GetFeatureFilterMetadata(string filterName, Type appContextType = null)
```

src/Microsoft.FeatureManagement/FeatureManager.cs
jimmyca15:
```suggestion
        private bool IsMatchingName(Type filterType, string filterName)
```

src/Microsoft.FeatureManagement/FeatureManager.cs
```diff
@@ -48,8 +48,8 @@ private class ConfigurationCacheItem
             _featureFilters = featureFilters ?? throw new ArgumentNullException(nameof(featureFilters));
             _sessionManagers = sessionManagers ?? throw new ArgumentNullException(nameof(sessionManagers));
             _logger = loggerFactory.CreateLogger<FeatureManager>();
-            _filterMetadataCache = new ConcurrentDictionary<string, IFeatureFilterMetadata>();
-            _contextualFeatureFilterCache = new ConcurrentDictionary<string, ContextualFeatureFilterEvaluator>();
+            _filterMetadataCache = new ConcurrentDictionary<ValueTuple<string, Type>, IFeatureFilterMetadata>();
+            _contextualFeatureFilterCache = new ConcurrentDictionary<ValueTuple<string, Type>, ContextualFeatureFilterEvaluator>();
```
jimmyca15, line 52: Instead of caching type, I recommend using just the name of the type like  we did before. Perhaps `Tuple<string, string>``

src/Microsoft.FeatureManagement/FeatureManager.cs
```diff
@@ -165,7 +175,7 @@ private async Task<bool> IsEnabledAsync<TContext>(string feature, TContext appCo
 
                         //
                         // IContextualFeatureFilter
```
jimmyca15, line 177: `if (useAppContext && ContextualFeatureFilterEvaluator.IsContextualFilter(filter, typeof(TContext)))`

At this point we should already know whether the filter is a contextual filter or not by whether filter was populated with or without an app context.

src/Microsoft.FeatureManagement/FeatureManager.cs
```diff
@@ -267,48 +277,72 @@ private void BindSettings(IFeatureFilterMetadata filter, FeatureFilterEvaluation
...
+        {
             IFeatureFilterMetadata filter = _filterMetadataCache.GetOrAdd(
-                filterName,
+                (filterName, appContextType),
                 (_) => {
 
                     IEnumerable<IFeatureFilterMetadata> matchingFilters = _featureFilters.Where(f =>
                     {
-                        Type t = f.GetType();
+                        Type filterType = f.GetType();
```
jimmyca15, line 320: if (!IsFilterNameMatched(filterType, filterName))
{
    return false;
}

src/Microsoft.FeatureManagement/FeatureManager.cs
```diff
@@ -267,48 +277,72 @@ private void BindSettings(IFeatureFilterMetadata filter, FeatureFilterEvaluation
...
-                        // If a feature is configured to use a filter without a namespace such as 'MyFilter', then it can match 'MyOrg.MyProduct.MyFilter' or simply 'MyFilter'
-                        // If a feature is configured to use a filter with a namespace such as 'MyOrg.MyProduct.MyFilter' then it can only match 'MyOrg.MyProduct.MyFilter' 
-                        if (filterName.Contains('.'))
-                        {
-                            //
-                            // The configured filter name is namespaced. It must be an exact match.
-                            return string.Equals(name, filterName, StringComparison.OrdinalIgnoreCase);
+                            return (f is IFeatureFilter) && IsFilterNameMatched(filterType, filterName);
                         }
                         else
```
jimmyca15, line 325: else not needed since the if returns.

src/Microsoft.FeatureManagement/FeatureManager.cs
```diff
@@ -267,48 +277,72 @@ private void BindSettings(IFeatureFilterMetadata filter, FeatureFilterEvaluation
...
-                            //
-                            // We take the simple name of a filter, E.g. 'MyFilter' for 'MyOrg.MyProduct.MyFilter'
-                            string simpleName = name.Contains('.') ? name.Split('.').Last() : name;
+                            IEnumerable<Type> contextualFilterInterfaces = filterType.GetInterfaces().Where(
+                                i => i.IsGenericType &&
+                                     i.GetGenericTypeDefinition().IsAssignableFrom(typeof(IContextualFeatureFilter<>)) &&
+                                     i.GetGenericArguments()[0].IsAssignableFrom(appContextType));
 
-                            return string.Equals(simpleName, filterName, StringComparison.OrdinalIgnoreCase);
+                            return contextualFilterInterfaces.Any() && IsFilterNameMatched(filterType, filterName);
```
jimmyca15, line 332:
```suggestion
                            return ContextualFeatureFilterEvaluator.IsContextualFilter(f, appContextType);
```

src/Microsoft.FeatureManagement/FeatureManager.cs
zhiyuanliang-ms: I convinced myself that using $"{filterName}{Environment.NewLine}{appContextType.FullName}" as the cache key is better.
I plan to follow the previous way.

jimmyca15: But I don't see the change?

src/Microsoft.FeatureManagement/FeatureManager.cs
jimmyca15: I don't see an update to the name

src/Microsoft.FeatureManagement/FeatureManager.cs
```diff
@@ -326,14 +357,17 @@ private ContextualFeatureFilterEvaluator GetContextualFeatureFilter(string filte
             }
 
             ContextualFeatureFilterEvaluator filter = _contextualFeatureFilterCache.GetOrAdd(
-                $"{filterName}{Environment.NewLine}{appContextType.FullName}",
+                $"{filterName}{Environment.NewLine}{appContextType?.FullName}",
```
jimmyca15, line 360:
```suggestion
                $"{filterName}{Environment.NewLine}{appContextType.FullName}",
```

jimmyca15 (approved):

PR #263, opened 2023-09-18
Add a flag for FeatureManagementOption to change the behavior when there is no FeatureManagement section in the configuration

zhiyuanliang-ms:
This PR is for release 2.x to avoid breaking change.


src/Microsoft.FeatureManagement/FeatureManagementOptions.cs
```diff
@@ -22,5 +22,12 @@ public class FeatureManagementOptions
         /// The default value is true.
         /// </summary>
         public bool IgnoreMissingFeatures { get; set; } = true;
+
+        /// <summary>
+        /// Controls the behavior of feature definition provider when "FeatureManagement" section in the configuration is missing.
+        /// If missing configuration section is not ignored, the feature definition provider will fall back to the root of the provided configuration.
+        /// The default value is false.
+        /// </summary>
+        public bool IgnoreMissingConfigurationSection { get; set; }
```
rossgrambo, line 31: I think the naming here could be a little clearer. We're not ignoring, rather we're no longer returning the top level configuration children.

Perhaps something like: "DefaultToTopLevelConfiguration" with a default of true?

zhenlan: Or `RequireFeatureManagementSection`

zhiyuanliang-ms: I vote for RequireFeatureManagementSection.

src/Microsoft.FeatureManagement/ServiceCollectionExtensions.cs
```diff
@@ -48,7 +50,7 @@ public static IFeatureManagementBuilder AddFeatureManagement(this IServiceCollec
                 throw new ArgumentNullException(nameof(configuration));
             }
 
-            services.AddSingleton<IFeatureDefinitionProvider>(new ConfigurationFeatureDefinitionProvider(configuration));
+            services.AddSingleton<IFeatureDefinitionProvider>(sp => new ConfigurationFeatureDefinitionProvider(configuration, sp.GetRequiredService<ILoggerFactory>(), sp.GetRequiredService<IOptions<FeatureManagementOptions>>()));
```
jimmyca15, line 53:
```suggestion
            services.AddSingleton<IFeatureDefinitionProvider>(sp =>
                new ConfigurationFeatureDefinitionProvider(
                    configuration,
                    sp.GetRequiredService<ILoggerFactory>(),
                    sp.GetRequiredService<IOptions<FeatureManagementOptions>>()));
```

src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
```diff
@@ -209,7 +214,16 @@ private IEnumerable<IConfigurationSection> GetFeatureDefinitionSections()
...
             else
             {
-                return _configuration.GetChildren();
+                if (_options.RequireFeatureManagementSection)
+                {
+                    _logger.LogWarning($"No configuration section named '{FeatureManagementSectionName}' was found.");
+
+                    return Enumerable.Empty<IConfigurationSection>();
+                }
+                else
```
jimmyca15, line 223: else isn't needed since the if returns.

src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
```diff
@@ -209,7 +214,16 @@ private IEnumerable<IConfigurationSection> GetFeatureDefinitionSections()
             }
             else
```
jimmyca15, line 215: else isn't needed since the if returns.

src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
```diff
@@ -209,7 +214,16 @@ private IEnumerable<IConfigurationSection> GetFeatureDefinitionSections()
...
             {
-                return _configuration.GetChildren();
+                if (_options.RequireFeatureManagementSection)
+                {
+                    _logger.LogWarning($"No configuration section named '{FeatureManagementSectionName}' was found.");
```
jimmyca15, line 219: If someone is setting that setting I would expect it is because they have a use case where they are running without feature flags at some point. I think a warning logged is just going to be noise.

rossgrambo: It's only logged when RequireFeatureManagementSection is set to true and there is no section. To me, it makes sense to warn because of the flag.

zhiyuanliang-ms: I agree with Ross.
By default, "RequireFeatureManagementSection" flag is false. If this flas is set to true, it indicates that the user intentionally did so. In this case, since the users are using FeatureManagment lib, they should add "FeatureManagement" section in the configuration. If we cannot find this section, there must be something wrong. Maybe, the section name is misspelled, which is a unintentional error. So, I think the warning log is necessary. Besides, it also follows the behavior of other feature management option flags.

rossgrambo:
Closing this in favor of https://github.com/microsoft/FeatureManagement-Dotnet/pull/263 since we're planning a major version bump

Issue #194, opened 2022-08-17
Can Microsoft.TimeWindow filter support recurring time?


sacrajit:
Where as enabling a feature during a certain duration once, can be useful for promotional capabilities, I found scenarios where one may need to turn on a feature during a low or high traffic period of a day or certain days of a week. Having the ability to combine days of week(or All) along with start and end time will be very helpful.

Thanks!

jimmyca15:
@sacrajit

It does not support a recurring time, but if that is something that makes sense we can look into adding a built-in filter to do it. Alternatively, you can implement an `IFeatureFilter` to create a recurring feature flag in your application.

Here's an [example ](https://github.com/microsoft/FeatureManagement-Dotnet/blob/main/examples/FeatureFlagDemo/BrowserFilter.cs) of a custom feature filter implementation.

zhiyuanliang-ms:
Hi, @sacrajit 
We are working on this. Please see the PR: https://github.com/microsoft/FeatureManagement-Dotnet/pull/256.

zhiyuanliang-ms:
@sacrajit #266 merged

In our next stable release 3.3.0, the recurring time window filter will be available.

zhiyuanliang-ms:
@sacrajit 

`Microsoft.FeatureManagement` v3.3.0 has been released.

PR #266, opened 2023-09-22
Support recurrence for TimeWindowFilter

zhiyuanliang-ms:
## Why this PR?

#194  

## Visible changes

Add-on "Recurrence" parameter for TimeWindowFilter using Outlook-style schema:
https://learn.microsoft.com/en-us/graph/outlook-schedule-recurring-events#using-patterns-and-ranges-to-create-recurring-events

The modification we made to the original outlook schema:
1. We remove the "StartDate" property in the RecurrenceRange and we force the "Start" property of the time window to be a valid first occurrence (there are validations for "Start" property, checking whether it follows the recurrence pattern).
2. We remove the "RecurrenceTimeZone" property in the RecurrenceRange and we will use the UTC offset in the "Start" property.
3. We will only support "Daily" and "Weekly" recurrence pattern type: https://github.com/microsoft/FeatureManagement-Dotnet/pull/266#issuecomment-1909088077

Some discussion can be found in #256 

### Other comments

The main logic of how to check whether current time is within any recurring time window:
1. Find the previous occurrence of the recurring time window, let's call it `prevOccurrenceStart`
2. Check whether the current time: `time` is within the time window: `prevOccurrenceStart` ~ `prevOccurrenceStart` + `End` - `Start`


src/Microsoft.FeatureManagement/FeatureFilters/Recurrence/RecurrenceRange.cs
```diff
@@ -0,0 +1,33 @@
...
+    {
+        /// <summary>
+        /// The recurrence range type
+        /// </summary>
+        public string Type { get; set; } = "NoEnd";
+
+        /// <summary>
+        /// The date to stop applying the recurrence pattern
+        /// </summary>
+        public DateTimeOffset? EndDate { get; set; }
```
zhenlan, line 21: Will a null EndDate be enough to represent no end?

src/Microsoft.FeatureManagement/FeatureFilters/Recurrence/RecurrenceEvaluator.cs
```diff
@@ -0,0 +1,1374 @@
...
+
+        //
+        // Recurrence Range Type
+        const string EndDate = "EndDate";
+        const string Numbered = "Numbered";
+        const string NoEnd = "NoEnd";
+
+        const int WeeklyUnitIntervalDuration = 7; // in days
+        const int MonthlyUnitIntervalDuration = 28; // in days
+        const int YearlyUnitIntervalDuration = 365; // in days
```
zhenlan, line 54: We probably should make all these const strings public. You have the `RecurrencePattern` public, but not their values.

src/Microsoft.FeatureManagement/FeatureFilters/TimeWindowFilter.cs
```diff
@@ -56,7 +56,19 @@ public Task<bool> EvaluateAsync(FeatureFilterEvaluationContext context)
...
+            //
+            // Hit the first occurrence of the time window
+            if ((!settings.Start.HasValue || now >= settings.Start.Value) && (!settings.End.HasValue || now < settings.End.Value))
+            {
+                return Task.FromResult(true);
+            }
+
+            if (settings.Recurrence != null)
+            {
+                return Task.FromResult(RecurrenceEvaluator.MatchRecurrence(now, settings));
```
amerjusupovic, line 68: I don't think the error messages mention the feature name anywhere at the moment in `RecurrenceEvaluator`, so maybe we could pass that in here from `context.FeatureName` and add it to the error message to make understanding the exceptions easier.

src/Microsoft.FeatureManagement/FeatureFilters/Recurrence/RecurrenceRange.cs
zhiyuanliang-ms: No, there is "Numbered" recurrence range type, which only requires "NumberOfOccurrences"

src/Microsoft.FeatureManagement/FeatureFilters/Recurrence/RecurrenceRange.cs
```diff
@@ -0,0 +1,33 @@
...
+{
+    /// <summary>
+    /// The recurrence range specifying how long the recurrence pattern repeats
+    /// </summary>
+    public class RecurrenceRange
+    {
+        /// <summary>
+        /// The recurrence range type
+        /// </summary>
+        public string Type { get; set; } = "NoEnd";
```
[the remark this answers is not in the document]
zhiyuanliang-ms: The benefit to use string here is that it can make BindParameters very simple
```
{
    return filterParameters.Get<TimeWindowFilterSettings>() ?? new TimeWindowFilterSettings();
}
```
Currently, the recurrence evaluator will validate the filter configuration settings. This behavior is the same as the targeting evaluator. If the "Type" property becomes enum here, the validation of "Type" will be moved to "BindParameters". I think the validation process should be performed by the recurrence evaluator.

src/Microsoft.FeatureManagement/FeatureFilters/TimeWindowFilter.cs
zhiyuanliang-ms: Sounds good. Maybe we should also change the behavior of the targeting filter. The targeting evaluator will not provide the feature name info in the error message either.

src/Microsoft.FeatureManagement/FeatureFilters/Recurrence/RecurrenceRange.cs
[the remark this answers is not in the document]
amerjusupovic: That makes sense in terms of not separating the validation, I was also trying to see if there's a way to avoid all the individual string comparisons and if statements in the evaluator.

src/Microsoft.FeatureManagement/FeatureFilters/TimeWindowFilter.cs
amerjusupovic: Oh I see, yea not sure if there was a reason for not including it but I'd think the same would apply there.

src/Microsoft.FeatureManagement/FeatureFilters/TimeWindowFilter.cs
```diff
@@ -37,7 +38,7 @@ public object BindParameters(IConfiguration filterParameters)
         }
 
         /// <summary>
-        /// Evaluates whether a feature is enabled based on a configurable time window.
+        /// Evaluates whether a feature is enabled based on configurable fixed time window or recurring time windows.
```
amerjusupovic, line 41:
```suggestion
        /// Evaluates whether a feature is enabled based on a configurable fixed time window or recurring time windows.
```

src/Microsoft.FeatureManagement/FeatureFilters/Recurrence/RecurrencePattern.cs
```diff
@@ -0,0 +1,48 @@
...
+
+        /// <summary>
+        /// The first day of the week.
+        /// </summary>
+        public string FirstDayOfWeek { get; set; } = "Sunday";
+
+        /// <summary>
+        /// Specifies on which instance of the allowed days specified in DaysOfWeek the time window occurs, counted from the first instance in the month
+        /// </summary>
+        public string Index { get; set; } = "First";
```
ivywei0125, line 36: Better to use the const varible from `RecurrenceEvaluator.cs`

src/Microsoft.FeatureManagement/FeatureFilters/Recurrence/RecurrenceEvaluator.cs
```diff
@@ -0,0 +1,1446 @@
...
+            if (!TryValidateSettings(settings, out string paramName, out string reason))
+            {
+                throw new ArgumentException(reason, paramName);
+            }
+
+            if (time < settings.Start.Value)
+            {
+                return false;
+            }
+
```
ivywei0125, line 79: It's kind of confusing why we need to check if `settings.Recurrence` is null. It has already been checked when call `RecurrenceEvaluator.matchRecurrence()` in "TimeWindowFilter.cs"

zhiyuanliang-ms: Removed.

src/Microsoft.FeatureManagement/FeatureFilters/Recurrence/RecurrenceEvaluator.cs
```diff
@@ -0,0 +1,1441 @@
...
+            if (!TryValidateInterval(settings, out paramName, out reason))
+            {
+                return false;
+            }
+
+            string patternType = settings.Recurrence.Pattern.Type;
+
+            if (string.Equals(patternType, Daily, StringComparison.OrdinalIgnoreCase))
+            {
+                if (!TryValidateDailyRecurrencePattern(settings, out paramName, out reason))
```
ivywei0125, line 591:
```
if (!TryValidateDailyRecurrencePattern(settings, out paramName, out reason))
{
    return false;
}
```
The code could be simplified as the following:
```
return TryValidateDailyRecurrencePattern(settings, out paramName, out reason);
```

zhiyuanliang-ms: Resolved.

src/Microsoft.FeatureManagement/FeatureFilters/Recurrence/RecurrenceEvaluator.cs
```diff
@@ -0,0 +1,1421 @@
...
+        /// Check whether the duration is shorter than the minimum gap between recurrence of days of week.
+        /// </summary>
+        /// <param name="duration">The time span of the duration.</param>
+        /// <param name="interval">The recurrence interval.</param>
+        /// <param name="daysOfWeek">The days of the week when the recurrence will occur.</param>
+        /// <param name="firstDayOfWeek">The first day of the week.</param>
+        /// <returns>True if the duration is compliant with days of week, false otherwise.</returns>
+        private static bool IsDurationCompliantWithDaysOfWeek(TimeSpan duration, int interval, IEnumerable<string> daysOfWeek, string firstDayOfWeek)
+        {
+            if (daysOfWeek == null || !daysOfWeek.Any())
```
ivywei0125, line 1254: The check about `daysOfWeek` is also repeated. We already have this check in `TryValidateDaysOfWeek` before.

zhiyuanliang-ms: Removed.

src/Microsoft.FeatureManagement/FeatureFilters/Recurrence/RecurrenceEvaluator.cs
```diff
@@ -0,0 +1,1416 @@
...
+            if (daysOfWeek.Count() == 1)
+            {
+                return true;
+            }
+
+            //
+            // Shift to the first day of the week
+            DateTime date = DateTime.Today;
+
+            int offset = RemainingDaysOfWeek((int)date.DayOfWeek, DayOfWeekNumber(firstDayOfWeek)) % 7;
```
ivywei0125, line 1263: Do we need this `%7`?

src/Microsoft.FeatureManagement/FeatureFilters/Recurrence/RecurrenceEvaluator.cs
```diff
@@ -0,0 +1,1416 @@
...
+
+            TimeSpan intervalDuration = TimeSpan.FromDays(pattern.Interval * MinYearDayNumber);
+
+            //
+            // Time window duration must be shorter than how frequently it occurs
+            if (settings.End.Value - settings.Start.Value > intervalDuration)
+            {
+                paramName = $"{nameof(settings.End)}";
+
+                reason = OutOfRange;
```
ivywei0125, line 879: Do we need more clear error info? like "Time window duration must be shorter than how frequently it occurs"

src/Microsoft.FeatureManagement/FeatureFilters/Recurrence/RecurrenceEvaluator.cs
zhiyuanliang-ms: Removed.

tests/Tests.FeatureManagement/RecurrenceEvaluator.cs
```diff
@@ -0,0 +1,2429 @@
...
+                        },
+                        Range = new RecurrenceRange()
+                        {
+                            Type = "NoEnd"
+                        }
+                    }
+                },
+                true ),
+
+                ( DateTimeOffset.Parse("2023-10-29T00:00:00+08:00"), // 4th Sunday in 2023 Oct
```
ivywei0125, line 1811: "2023-10-29" should be the 5th Sunday in 2023 Oct?
1st Sunday: 2023-10-01
2nd Sunday: 2023-10-08
3rd Sunday: 2023-10-15
4th Sunday: 2023-10-22
5th Sunday: 2023-10-29

src/Microsoft.FeatureManagement/FeatureFilters/Recurrence/RecurrencePattern.cs
```diff
@@ -0,0 +1,48 @@
...
+        public IEnumerable<string> DaysOfWeek { get; set; }
+
+        /// <summary>
+        /// The first day of the week.
+        /// </summary>
+        public string FirstDayOfWeek { get; set; } = "Sunday";
+
+        /// <summary>
+        /// Specifies on which instance of the allowed days specified in DaysOfWeek the time window occurs, counted from the first instance in the month
+        /// </summary>
```
jimmyca15, line 35: What are the valid values of `Index`? Seems like it should be an enum.

src/Microsoft.FeatureManagement/FeatureFilters/Recurrence/RecurrencePattern.cs
```diff
@@ -0,0 +1,48 @@
...
+
+        /// <summary>
+        /// The days of the week on which the time window occurs
+        /// </summary>
+        public IEnumerable<string> DaysOfWeek { get; set; }
+
+        /// <summary>
+        /// The first day of the week.
+        /// </summary>
+        public string FirstDayOfWeek { get; set; } = "Sunday";
```
jimmyca15, line 31: FirstDayOfWeek seems like it should be an enum.

jimmyca15: In settings enums are modeled as string, but a finite set of options is better represented by an enum in .NET

src/Microsoft.FeatureManagement/FeatureFilters/Recurrence/RecurrencePattern.cs
```diff
@@ -0,0 +1,48 @@
...
+{
+    /// <summary>
+    /// The recurrence pattern specifying how often the time window repeats
+    /// </summary>
+    public class RecurrencePattern
+    {
+        /// <summary>
+        /// The recurrence pattern type
+        /// </summary>
+        public string Type { get; set; }
```
jimmyca15, line 16: Enum for `Type` rather than string.

src/Microsoft.FeatureManagement/FeatureFilters/Recurrence/RecurrencePattern.cs
zhiyuanliang-ms: The reason I used string instead of enum here is that
```C#
{
    return filterParameters.Get<TimeWindowFilterSettings>() ?? new TimeWindowFilterSettings();
}
```
If I use enum here, the `BindParameters` will have to parse the string and map it to the enum. In this case, should the validation of the parameters happen in `BindParameters`?


Amer has called out this issue: https://github.com/microsoft/FeatureManagement-Dotnet/pull/266#discussion_r1340432430

src/Microsoft.FeatureManagement/FeatureFilters/Recurrence/RecurrencePattern.cs
zhiyuanliang-ms: https://learn.microsoft.com/en-us/graph/api/resources/recurrencepattern?view=graph-rest-1.0#properties

Index specifies on which instance of the allowed days specified in daysOfWeek the event occurs, counted from the first instance in the month. The possible values are: first, second, third, fourth, last. Default is first. Optional and used if type is relativeMonthly or relativeYearly.

zhiyuanliang-ms:
Null check is unnecessary for the section which has default value.

zhiyuanliang-ms:
Should we add check for the invalid case: month is June and date is 31.

src/Microsoft.FeatureManagement/FeatureFilters/Recurrence/RecurrenceRange.cs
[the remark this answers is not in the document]
zhiyuanliang-ms: This should be enum. It's my fault. I underestimated the powerful capabilities of .NET.
At first, I thought that if I used an enum, the filterParameters.Get<TimeWindowFilterSettings>() wouldn't be able to automatically convert a string in the configuration into the enum. But when I tried it out, I found out that .NET is incredible, it can automatically parse a string into an enum. So, here, we definitly should use enum.

src/Microsoft.FeatureManagement/FeatureFilters/Recurrence/RecurrencePattern.cs
zhiyuanliang-ms: This should be enum. It's my fault. I underestimated the powerful capabilities of .NET.
At first, I thought that if I used an enum, the filterParameters.Get() wouldn't be able to automatically convert a string in the configuration into the enum. But when I tried it out, I found out that .NET is incredible, it can automatically parse a string into an enum. So, here, we definitly should use enum.

zhiyuanliang-ms:
@jimmyca15 @rossgrambo @ivywei0125 This PR is ready for review.

src/Microsoft.FeatureManagement/FeatureFilters/TimeWindowFilter.cs
```diff
@@ -10,6 +10,7 @@ namespace Microsoft.FeatureManagement.FeatureFilters
 {
     /// <summary>
     /// A feature filter that can be used to activate a feature based on a time window.
+    /// The time window filter supports recurrence settings. The time window can occur repeatedly.
```
jimmyca15, line 13:
```suggestion
    /// The time window can be configured to recur periodically.
```

src/Microsoft.FeatureManagement/FeatureFilters/TimeWindowFilter.cs
```diff
@@ -37,7 +38,7 @@ public object BindParameters(IConfiguration filterParameters)
         }
 
         /// <summary>
-        /// Evaluates whether a feature is enabled based on a configurable time window.
+        /// Evaluates whether a feature is enabled based on a configurable fixed time window or recurring time windows.
```
jimmyca15, line 41: I don't think we need to update every summary to mention fixed or recurring. Ideally that's a one time thing and we figure out how to generalize elsewhere.

src/Microsoft.FeatureManagement/FeatureFilters/Recurrence/RecurrencePattern.cs
```diff
@@ -0,0 +1,49 @@
...
+    {
+        /// <summary>
+        /// The recurrence pattern type.
+        /// </summary>
+        public RecurrencePatternType Type { get; set; }
+
+        /// <summary>
+        /// The number of units between occurrences, where units can be in days, weeks, months, or years, depending on the pattern type.
+        /// </summary>
+        public int Interval { get; set; } = 1;
```
jimmyca15, line 22: Why initialize to 1?

src/Microsoft.FeatureManagement/FeatureFilters/Recurrence/RecurrencePattern.cs
```diff
@@ -0,0 +1,49 @@
...
+
+        /// <summary>
+        /// The days of the week on which the time window occurs.
+        /// </summary>
+        public IEnumerable<DayOfWeek> DaysOfWeek { get; set; }
+
+        /// <summary>
+        /// The first day of the week.
+        /// </summary>
+        public DayOfWeek FirstDayOfWeek { get; set; }
```
jimmyca15, line 32: Since this is an existing enum, does the default value work?

src/Microsoft.FeatureManagement/FeatureFilters/Recurrence/RecurrenceEvaluator.cs
```diff
@@ -0,0 +1,1117 @@
...
+
+                    reason = UnrecognizableValue;
+
+                    return false;
+            }
+        }
+
+        private static bool TryValidateDailyRecurrencePattern(TimeWindowFilterSettings settings, out string paramName, out string reason)
+        {
+            Debug.Assert(settings.Recurrence.Pattern.Interval >  0);
```
jimmyca15, line 185: I don't see this validated before this step. Am I missing it?

src/Microsoft.FeatureManagement/FeatureFilters/Recurrence/RecurrenceEvaluator.cs
```diff
@@ -0,0 +1,1117 @@
...
+        {
+            int remainingDays = (int) dayOfWeek - (int) firstDayOfWeek;
+
+            if (remainingDays < 0)
+            {
+                return -remainingDays;
+            }
+            else
+            {
+                return DayNumberOfWeek - remainingDays;
```
jimmyca15, line 1081:
```suggestion
                return DaysPerWeek - remainingDays;
```

src/Microsoft.FeatureManagement/FeatureFilters/Recurrence/RecurrenceEvaluator.cs
```diff
@@ -0,0 +1,1117 @@
...
+
+                        prev = date;
+                    }
+                }
+
+                date += TimeSpan.FromDays(1);
+            }
+
+            //
+            // It may across weeks. Check the next week if the interval is one week.
```
jimmyca15, line 1044: > week if the interval is one week

Can the duration not be 2 weeks, or perhaps 15 days?

zhiyuanliang-ms: No, it can't. The time window duration should at least be shorter than the interval duration.

``` C#
//
// Time window duration must be shorter than how frequently it occurs
if (timeWindowDuration > intervalDuration || 
    !IsDurationCompliantWithDaysOfWeek(timeWindowDuration, pattern.Interval, pattern.DaysOfWeek, pattern.FirstDayOfWeek))
{
    paramName = $"{nameof(settings.End)}";

    reason = TimeWindowDurationOutOfRange;

    return false;
}
```

https://github.com/microsoft/FeatureManagement-Dotnet/blob/zhiyuanliang/recurring-time-window/src/Microsoft.FeatureManagement/FeatureFilters/Recurrence/RecurrenceEvaluator.cs#L225

src/Microsoft.FeatureManagement/FeatureFilters/Recurrence/RecurrencePattern.cs
zhiyuanliang-ms: Yes. The default value of DayOfWeek is Sunday.
``` C#
namespace System
{
    //
    // Summary:
    //     Specifies the day of the week.
    public enum DayOfWeek
    {
        //
        // Summary:
        //     Indicates Sunday.
        Sunday = 0,
        //
        // Summary:
        //     Indicates Monday.
        Monday = 1,
        //
        // Summary:
        //     Indicates Tuesday.
        Tuesday = 2,
        //
        // Summary:
        //     Indicates Wednesday.
        Wednesday = 3,
        //
        // Summary:
        //     Indicates Thursday.
        Thursday = 4,
        //
        // Summary:
        //     Indicates Friday.
        Friday = 5,
        //
        // Summary:
        //     Indicates Saturday.
        Saturday = 6
    }
}
```

src/Microsoft.FeatureManagement/FeatureFilters/Recurrence/RecurrencePattern.cs
zhiyuanliang-ms: Yes, the outlook schema doesn't say it has a default value. The reason I set a default value:
1. It will bring more convenience for users who write the JSON configuration directly. Users won't need to write the interval if they just want the interval to be one. I think it would be better to make the job of writing configuration as easy as possible since this is already a very complex schema. And I also guess the interval = 1 is the most common case to set recurrence.
2. The Index property has a default value. It will be more consistent to give Interval a default value.

src/Microsoft.FeatureManagement/FeatureFilters/Recurrence/RecurrenceEvaluator.cs
zhiyuanliang-ms: Added.

src/Microsoft.FeatureManagement/FeatureFilters/TimeWindowFilter.cs
zhiyuanliang-ms: How about "Evaluates whether a feature is enabled based on the <see cref="TimeWindowFilterSettings"/> specifed in the configuration."?

src/Microsoft.FeatureManagement/FeatureFilters/Recurrence/RecurrenceEvaluator.cs
```diff
@@ -0,0 +1,1117 @@
...
+            return true;
+        }
+
+        private static bool TryValidateAbsoluteMonthlyRecurrencePattern(TimeWindowFilterSettings settings, out string paramName, out string reason)
+        {
+            RecurrencePattern pattern = settings.Recurrence.Pattern;
+
+            Debug.Assert(pattern.Interval > 0);
+
+            TimeSpan intervalDuration = TimeSpan.FromDays(pattern.Interval * MinDayNumberOfMonth);
```
amerjusupovic: How does this work when someone sets the duration to 29 days for example, and their recurrence doesn't include february? Is that a valid scenario?

src/Microsoft.FeatureManagement/FeatureFilters/Recurrence/RecurrenceEvaluator.cs
```diff
@@ -0,0 +1,1117 @@
...
+        /// <param name="index">The index of the day of week in the month.</param>
+        /// <param name="dayOfWeek">The day of week.</param>
+        /// <returns>The data time of the nth day of week in the month.</returns>
+        private static DateTime NthDayOfWeekInTheMonth(DateTime dateTime, WeekIndex index, DayOfWeek dayOfWeek)
+        {
+            var date = new DateTime(dateTime.Year, dateTime.Month, 1);
+
+            //
+            // Find the first provided day of week in the month
+            while (date.DayOfWeek != dayOfWeek)
```
amerjusupovic, line 1098: Just wondering why you decided to loop over the days instead of some kind of subtraction between `dayOfWeek` and `date.DayOfWeek`?

src/Microsoft.FeatureManagement/FeatureFilters/Recurrence/RecurrencePatternType.cs
```diff
@@ -0,0 +1,41 @@
...
+    /// </summary>
+    public enum RecurrencePatternType
+    {
+        /// <summary>
+        /// The pattern where the time window will repeat based on the number of days specified by interval between occurrences.
+        /// </summary>
+        Daily,
+
+        /// <summary>
+        /// The pattern where the time window will repeat on the same day or days of the week, based on the number of weeks between each set of occurrences
```
rossgrambo, line 17:
```suggestion
        /// The pattern where the time window will repeat on the same day or days of the week, based on the number of weeks between each set of occurrences.
```

src/Microsoft.FeatureManagement/FeatureFilters/Recurrence/RecurrenceEvaluator.cs
```diff
@@ -0,0 +1,1117 @@
...
+            //
+            // Time window duration must be shorter than how frequently it occurs
+            if (timeWindowDuration > intervalDuration)
+            {
+                paramName = $"{nameof(settings.End)}";
+
+                reason = TimeWindowDurationOutOfRange;
+
+                return false;
+            }
```
rossgrambo, line 200: Could pull this out into a helper method.

rossgrambo (commented):
I don't see anything wrong with this PR and I love the robustness of the test suite. However- this PR is quite large. 

I'm concerned that if we support things in this way- we'll need to recreate this code in each language. It doesn't sound ideal to me- especially since portal will have to implement this as well, with appropriate erroring. 

Perhaps we think this feature is important enough to justify it, or perhaps we think this schema is reliable (as it's from outlook) and that's enough reasoning to invest in it. 

Personally- I think this feature is valuable enough to justify it, so we can proceed. But I just wanted to call out the additional cost of it.

src/Microsoft.FeatureManagement/FeatureFilters/Recurrence/RecurrenceEvaluator.cs
zhiyuanliang-ms: If the recurrence pattern type is "AbsoluteMonthly" or "RelativeMonthly“ and the "Interval" is 1, the time window duration cannot be longer than 28 days. If the "Interval" is 2, then the time window duration cannot be longer than 48 days. This is the behavior of Outlook. You can check this in the Outlook app by creating a meeting and making it recurring. Please use the old Outlook where you can see the error message. If you use the new Outlook, it only tells you that it fails to create the event.
image

src/Microsoft.FeatureManagement/FeatureFilters/Recurrence/RecurrenceEvaluator.cs
zhiyuanliang-ms: Oh, good suggestion. Perhaps I was not in my clear mind when I was coding at that time. Yes, indeed, subtraction should be used here, or the above DaysOfTheWeek helper function could be utilized.

src/Microsoft.FeatureManagement/FeatureFilters/Recurrence/RecurrenceRange.cs
```diff
@@ -0,0 +1,28 @@
...
+
+        /// <summary>
+        /// The date to stop applying the recurrence pattern.
+        /// </summary>
+        public DateTimeOffset? EndDate { get; set; }
+
+        /// <summary>
+        /// The number of times to repeat the time window.
+        /// </summary>
+        public int NumberOfOccurrences { get; set; } = 1;
```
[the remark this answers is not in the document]
zhiyuanliang-ms: Yes. It makes more sense to let this property be nullable. Decided to remove the default value and change the type to int?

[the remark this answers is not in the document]
zhiyuanliang-ms: After reconsidering it, I still prefer to set a default value here. I agree that it would be more intuitive for this property to be nullable since it is only required when the recurrence range type is "Numbered". But I want to argue for the current choice:

The main reason I choose to set a default value and avoid using type nullable int is:
If the type is nullable int, then if we want to refer it, we will need to write the code like this: 
``` C#
if(Range.NumberOfOccurrences.Value < 1)
```
And also we need to add null check for `Range.NumberOfOccurrences`. I want to simplify the code.


The inconsistency is that "EndDate" property is nullable. 

We do have two nullable properties in RecurrencePattern which are "DayOfMonth" and "Month". The reason why these two properties have no default value is that I cannot come up with a reasonable default value for them. If everyone agree that the DayOfMonth and Month should be 1 by default. I will definitly set a default value for them. Because in this way, we can simplify our validation process by removing the null check for them.

In fact, I just found I forgot to cover the case if the "EndDate" is null when RecurrenceRangeType is "EndDate" in unit tests.

So my current thought is to keep the default value for the "NumberOfOccurrences" and set a default value DateTimeOffset.MaxValue for "EndDate". Do you like this idea?

src/Microsoft.FeatureManagement/FeatureFilters/Recurrence/RecurrenceEvaluator.cs
amerjusupovic: Thanks for the explanation! If that's the outlook behavior then I think it's reasonable to keep things consistent

src/Microsoft.FeatureManagement/FeatureFilters/Recurrence/RecurrenceEvaluator.cs
```diff
@@ -0,0 +1,1115 @@
...
+            {
+                //
+                // E.g. start: 2023.9.1 (the first Friday in 2023.9) and time: 2024.9.2 (the first Friday in 2023.9 is 2024.9.6)
+                // Not a complete year
+                yearGap -= 1;
+            }
+
+            int numberOfInterval = yearGap / interval;
+
+            DateTime previousOccurrenceMonth = start.AddYears(numberOfInterval * interval).DateTime;
```
amerjusupovic, line 967:
```suggestion
            DateTime previousOccurrenceMonth = start.AddYears(yearGap).DateTime;
```

Looks like it's the same value so maybe you can just use `yearGap`

zhiyuanliang-ms: Actually, numberOfInterval * interval may not equal to yearGap. 

For example, if yearGap is 5 and interval is 2
int numberOfInterval = yearGap / interval; // this will be 2

amerjusupovic: Missed that, my fault

jimmyca15:
@zhiyuanliang-ms @rossgrambo @amerjusupovic @zhenlan

zhiyuanliang-ms:
The monthly & yearly recurrence pattern has been removed.

src/Microsoft.FeatureManagement/FeatureFilters/Recurrence/RecurrencePattern.cs
```diff
@@ -0,0 +1,34 @@
...
+    /// </summary>
+    public class RecurrencePattern
+    {
+        /// <summary>
+        /// The recurrence pattern type.
+        /// </summary>
+        public RecurrencePatternType Type { get; set; }
+
+        /// <summary>
+        /// The number of units between occurrences, where units can be in days, weeks, months, or years, depending on the pattern type.
```
rossgrambo, line 20: I suppose months and years should be taken out here for now.

rossgrambo (approved):

src/Microsoft.FeatureManagement/FeatureFilters/Recurrence/RecurrenceRange.cs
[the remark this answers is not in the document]
jimmyca15: If EndDate is `MaxValue` by default, why wouldn't `NumberOfOccurrences` be `int.MaxValue` ?

src/Microsoft.FeatureManagement/FeatureFilters/Recurrence/RecurrencePattern.cs
```diff
@@ -0,0 +1,34 @@
...
+        /// </summary>
+        public RecurrencePatternType Type { get; set; }
+
+        /// <summary>
+        /// The number of units between occurrences, where units can be in days or weeks, depending on the pattern type.
+        /// </summary>
+        public int Interval { get; set; } = 1;
+
+        /// <summary>
+        /// The days of the week on which the time window occurs.
```
jimmyca15, line 25: Perhaps it might help to mention that it's only applicable for weekly occurence.

src/Microsoft.FeatureManagement/FeatureFilters/Recurrence/RecurrenceEvaluator.cs
```diff
@@ -0,0 +1,630 @@
...
+        }
+
+        /// <summary>
+        /// Find the closest previous recurrence occurrence before the provided timestamp according to the "Weekly" recurrence pattern.
+        /// <param name="time">A timestamp.</param>
+        /// <param name="settings">The settings of time window filter.</param>
+        /// <param name="previousOccurrence">The closest previous occurrence.</param>
+        /// <param name="numberOfOccurrences">The number of occurrences between the time and the recurrence start.</param>
+        /// </summary>
+        private static void FindWeeklyPreviousOccurrence(DateTimeOffset time, TimeWindowFilterSettings settings, out DateTimeOffset previousOccurrence, out int numberOfOccurrences)
```
jimmyca15, line 446: I'd like to chat about this algorithm, I'm wondering if we can do something a bit easier to follow.

In my mind a proposition is

```

1. first find day 0 (d0), it's the day representing the start day on the week of `Start`.
2. Find start day of most recent occurring week floor((targetDate - d0) / (interval * 7)) * (interval * 7)
3. if that's over 7 days ago, then previous occurence is the day with greatest offset of last occurring week
4. If gotten this far, then the current week has an occurence:
  i. if day offset > min offset then use maximum offset that is less than current to check if in window.
  ii.  If day offset < min offset,  then previous occurence is the day with greatest offset of last occurring week
```

but we can discuss

src/Microsoft.FeatureManagement/FeatureFilters/TimeWindowFilter.cs
```diff
@@ -56,7 +57,19 @@ public Task<bool> EvaluateAsync(FeatureFilterEvaluationContext context)
...
-            return Task.FromResult((!settings.Start.HasValue || now >= settings.Start.Value) && (!settings.End.HasValue || now < settings.End.Value));
+            //
+            // Hit the first occurrence of the time window
+            if ((!settings.Start.HasValue || now >= settings.Start.Value) && (!settings.End.HasValue || now < settings.End.Value))
+            {
+                return Task.FromResult(true);
+            }
+
+            if (settings.Recurrence != null)
+            {
```
jimmyca15, line 68: So the recurrence evaluation gets called on every request. I'd like for us to consider the implications of that. For example, is any caching needed?

zhiyuanliang-ms: The validation of recurrence settings also gets called on every request. I can move the validation to `BindParameters`.

To implement a cache for recurrence evaluator, we should at least make some change to the FeatureFilterEvaluationContext. The reason is that: One feature flag can have multiple time window filters. We need the filter index information to distinguish different time window filters of the same feature flag so that we can retrieve the correct cache key "featurename_timewindowfilter_index_recurrence_cache"

zhiyuanliang-ms:
Update: 
Now TimeWindowFilter can use cache while evaluating recurrence. The test result of performance improvement can be found [here](https://github.com/zhiyuanliang-ms/FeatureManagement-Dotnet/tree/recurring-time-window-perf-test/benchmark/BenchmarkDotNet.Artifacts/results)

src/Microsoft.FeatureManagement/FeatureFilters/Recurrence/RecurrenceEvaluator.cs
```diff
@@ -0,0 +1,695 @@
...
+        }
+
+        /// <summary>
+        /// Perform validation of time window settings.
+        /// <param name="settings">The settings of time window filter.</param>
+        /// <param name="paramName">The name of the invalid setting, if any.</param>
+        /// <param name="reason">The reason that the setting is invalid.</param>
+        /// <returns>True if the provided settings are valid. False if the provided settings are invalid.</returns>
+        /// </summary>
+        public static bool TryValidateSettings(TimeWindowFilterSettings settings, out string paramName, out string reason)
```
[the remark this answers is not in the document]
zhiyuanliang-ms: I follow the TargetingEvaluator.
https://github.com/microsoft/FeatureManagement-Dotnet/blob/preview/src/Microsoft.FeatureManagement/Targeting/TargetingEvaluator.cs#L239

Since the validation for recurrence settings is quite complex and there are a lot of methods for it. I agree to move them to another file.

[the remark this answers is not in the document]
zhiyuanliang-ms: @rossgrambo Moved.

jimmyca15:
@zhiyuanliang-ms let's go ahead and include associated readme update in this PR so it goes in one shot.

src/Microsoft.FeatureManagement/FeatureFilters/Recurrence/RecurrenceEvaluator.cs
```diff
@@ -0,0 +1,330 @@
...
+            // netstandard2.0 does not support '/' operator for TimeSpan. After we stop supporting netstandard2.0, we can remove .TotalSeconds.
+            int numberOfInterval = (int)Math.Floor((time - firstDayOfStartWeek).TotalSeconds / TimeSpan.FromDays(interval * DaysPerWeek).TotalSeconds);
+
+            DateTimeOffset firstDayOfMostRecentOccurringWeek = firstDayOfStartWeek.AddDays(numberOfInterval * (interval * DaysPerWeek));
+
+            List<DayOfWeek> sortedDaysOfWeek = SortDaysOfWeek(pattern.DaysOfWeek, pattern.FirstDayOfWeek);
+
+            //
+            // Subtract the days before the start in the first week.
+            numberOfOccurrences = numberOfInterval * sortedDaysOfWeek.Count - sortedDaysOfWeek.IndexOf(start.DayOfWeek);
```
jimmyca15, line 243: For weekly, does number of occurences represent how many weeks pass? Or how many individual windows pass?

For example, if I have Weekly with Mon, Wed, Fri and number of occurrences is two, does that mean I'll have two separate weeks with Mon, Wed, Fri or a single week with only two days?

zhiyuanliang-ms: Number of occurrences represent how many individual windows **start time** have passed.

> For example, if I have Weekly with Mon, Wed, Fri and number of occurrences is two, does that mean I'll have two separate weeks with Mon, Wed, Fri or a single week with only two days?

a single week with only two days 

It could also be other case, for example: today is 2024.3.20 (Wed), 
Weekly Pattern is [Mon, Tue, Thu] and Interval is 2
The start is 2024.3.7 (Thu).

For today (2024.3.20 (Wed)), the number of occurrences is 3 (but it actualy went across 3 weeks)

README.md
```diff
@@ -565,13 +565,117 @@ This filter provides the capability to enable a feature based on a time window.
...
+                "End": "Mon, 01 Jul 2019 00:00:00 GMT"
             }
         }
     ]
 }
 ```
 
+The time window can be configured to recur periodically. This can be useful for the scenarios where one may need to turn on a feature during a low or high traffic period of a day or certain days of a week. To expand the individual time window to recurring time windows, the recurrence rule should be specified in the `Recurrence` parameter.
+
+**Note:** `Start` and `End` must be both specified to enable `Recurrence`.
```
rossgrambo, line 577: Why does `End` need to be specified?

src/Microsoft.FeatureManagement/FeatureFilters/TimeWindowFilter.cs
```diff
@@ -56,7 +68,74 @@ public Task<bool> EvaluateAsync(FeatureFilterEvaluationContext context)
...
+                //
+                // The reference of the object will be used for hash key.
+                // If there is no pre-bounded settings attached to the context, there will be no cached filter settings and each call will have a unique settings object.
+                // In this case, the cache for recurrence settings won't work.
+                if (context.Settings != null)
+                {
+                    DateTimeOffset cachedTime = _recurrenceCache.GetOrAdd(
```
jimmyca15, line 86: this cache grows unbound.

jimmyca15: We may want to use a memory cache to support expiration.

Also, I expect the cache key should be the feature name. Otherwise, tweaks to the settings for a given feature would result in a new cache item and a leftover cache item that's useless.

src/Microsoft.FeatureManagement/FeatureFilters/Recurrence/RecurrenceEvaluator.cs
```diff
@@ -0,0 +1,330 @@
...
+
+        /// <summary>
+        /// Try to find the closest previous recurrence occurrence (if any) before the provided timestamp and the next occurrence.
+        /// <param name="time">A timestamp.</param>
+        /// <param name="settings">The settings of time window filter.</param>
+        /// <param name="prevOccurrence">The closest previous occurrence. If there is no previous occurrence, it will be set to <see cref="DateTimeOffset.MinValue"/>.</param>
+        /// <param name="nextOccurrence">The next occurrence.</param>
+        /// <returns>True if the closest previous occurrence is within the recurrence range or the time is before the first occurrence, false otherwise.</returns>
+        /// </summary>
+        public static bool TryFindPrevAndNextOccurrences(DateTimeOffset time, TimeWindowFilterSettings settings, out DateTimeOffset prevOccurrence, out DateTimeOffset nextOccurrence)
```
jimmyca15, line 45: This method is really hard to understand. Given the current name, one would expect it to only be true if prevOccurrence and nextOccurrence are both found.

src/Microsoft.FeatureManagement/FeatureFilters/TimeWindowFilter.cs
```diff
@@ -56,7 +68,74 @@ public Task<bool> EvaluateAsync(FeatureFilterEvaluationContext context)
...
+                            {
+                                return prevOccurrence;
+                            }
+
+                            //
+                            // There is no previous occurrence within the reccurrence range.
+                            return DateTimeOffset.MaxValue;
+                        });
+
+                    if (now < cachedTime)
```
jimmyca15, line 100: If there's no previous occurrence, you set the cached value to MaxValue right above. Wouldn't that mean always returning false from that point forward?

src/Microsoft.FeatureManagement/FeatureFilters/TimeWindowFilter.cs
```diff
@@ -56,7 +68,74 @@ public Task<bool> EvaluateAsync(FeatureFilterEvaluationContext context)
...
+                    //
+                    // There is no previous occurrence within the reccurrence range.
+                    _recurrenceCache.AddOrUpdate(
+                        settings,
+                        (_) => throw new KeyNotFoundException(),
+                        (_, _) => DateTimeOffset.MaxValue);
+
+                    return Task.FromResult(false);
+                }
+
```
jimmyca15, line 134: I'd rather see the if reverted.

```
if (context.Settings == null)
{
    return Task.FromResult(RecurrenceEvaluator.MatchRecurrence(now, settings));
}

//
// Here evaluate with cached settings

//
// No longer need final `return Task.FromResult(false);` since it's handled as last line of whole method.

```

src/Microsoft.FeatureManagement/FeatureFilters/Recurrence/RecurrenceEvaluator.cs
jimmyca15: Or

```
GetSurroundingOccurrences(
    DateTimeOffset time,
    out DateTimeOffset? previous,
    out DateTimeOffset? next
)
```

README.md
```diff
@@ -565,13 +565,117 @@ This filter provides the capability to enable a feature based on a time window.
...
+
+There are three possible recurrence range type: `NoEnd`, `EndDate` and `Numbered`.
+
+- `NoEnd`
+
+    The `NoEnd` range causes the recurrence to occur indefinitely.
+
+    | Property | Relevance | Description |
+    |----------|-----------|-------------|
+    | **Type** | Required | Must be set to `EndDate`. |
```
jimmyca15, line 655: Should it say, "Must be set to `NoEnd`" ?

README.md
```diff
@@ -565,13 +565,117 @@ This filter provides the capability to enable a feature based on a time window.
...
+    The `Numbered` range causes the time window to occur a fixed number of times (based on the pattern).
+
+    | Property | Relevance | Description |
+    |----------|-----------|-------------|
+    | **Type** | Required | Must be set to `Numbered`. |
+    | **NumberOfOccurrences** | Required | 	Specifies the number of occurrences. |
+
+To create a recurrence rule, you must specify both `Pattern` and `Range`. Any pattern type can work with any range type.
+
+**Advanced:** The time zone offset of the `Start` will apply to the recurrence settings.
```
jimmyca15, line 677:
```suggestion
**Advanced:** The time zone offset of the `Start` property will apply to the recurrence settings.
```

README.md
```diff
@@ -565,13 +565,117 @@ This filter provides the capability to enable a feature based on a time window.
...
+- `EndDate`
+
+    The `EndDate` range causes the time window to occur on all days that fit the applicable pattern until the end date.
+
+    | Property | Relevance | Description |
+    |----------|-----------|-------------|
+    | **Type** | Required | Must be set to `EndDate`. |
+    | **EndDate** | Required | 	Specifies the date time to stop applying the pattern. Note that as long as the start time of the last occurrence falls before the end date, the end time of that occurrence is allowed to extend beyond it. |
+
+- `Numbered`
```
jimmyca15, line 666: I think an example for EndDate and Numbered would be helpful, like what you did with Weekly recurrence pattern.

zhiyuanliang-ms: Added.

The following example will repeat the time window every day until the last occurrence happens on April 1st, 2024.

    ``` javascript
    "Start": "Fri, 22 Mar 2024 18:00:00 GMT",
    "End": "Fri, 22 Mar 2024 20:00:00 GMT",
    "Recurrence":{
        "Pattern": {
            "Type": "Daily",
            "Interval": 1
        },
        "Range": {
            "Type": "EndDate",
            "EndDate": "Mon, 1 Apr 2024 20:00:00 GMT"
        }
    }
    ```
    
The following example will repeat the time window on Monday and Tuesday until the there are 3 occurrences, which respectively happens on April 1st(Mon), April 2nd(Tue) and April 8th(Mon).

    ``` javascript
    "Start": "Mon, 1 Apr 2024 18:00:00 GMT",
    "End": "Mon, 1 Apr 2024 20:00:00 GMT",
    "Recurrence":{
        "Pattern": {
            "Type": "Weekly",
            "Interval": 1
            "DaysOfWeek": ["Monday", "Tuesday"],
        },
        "Range": {
            "Type": "Numbered",
            "NumberOfOccurrences": 3
        }
    }
    ```

src/Microsoft.FeatureManagement/FeatureFilters/TimeWindowFilter.cs
zhiyuanliang-ms: I covered this scenario in the test suite (Line 1581 in Tests.FeatureManagement/RecurrenceEvaluation.cs)


If there's no previous occurrence (haven't reached start yet), `TryFindPrevAndNextOccurrences` will return true and prevOccurrence will be set to MinValue.

``` C#
public static bool TryFindPrevAndNextOccurrences(DateTimeOffset time, TimeWindowFilterSettings settings, out DateTimeOffset prevOccurrence, out DateTimeOffset nextOccurrence)
{
    prevOccurrence = DateTimeOffset.MinValue;

    nextOccurrence = DateTimeOffset.MaxValue;

    if (time < settings.Start.Value)
    {
        //
        // The time is before the first occurrence.
        nextOccurrence = settings.Start.Value;

        return true;
    }
```
But this implementation seems wired. I will take the suggestion to modify the `TryFindPrevAndNextOccurrences` method to `void CalculateSurroundingOccurrences`

src/Microsoft.FeatureManagement/FeatureFilters/TimeWindowFilter.cs
zhiyuanliang-ms: We have `AddMemoryCache` call within the `AddFeatureManagement`. While using DI, feature manager will use the shared IMemoryCache. Should TimeWindowFilter follow this pattern? Or it should have its own memory cache?

zhiyuanliang-ms: > I expect the cache key should be the feature name.

I remember we discussed this before. A feature flag can have multiple timewindow filters with recurrence settings. 

> tweaks to the settings for a given feature would result in a new cache item

This is expected, isn't it? If we use feature name as cache key, when recurrence setting of a feature is changed, how we know the old cache with the key of the feature name should be discarded?

> a leftover cache item that's useless.

This is a problem. I agree that we should use Memory cache and set an expiration time.

src/Microsoft.FeatureManagement/FeatureFilters/Recurrence/RecurrenceEvaluator.cs
zhiyuanliang-ms: How about rename it to `TryCalculateSurroundingOccurrences(
    DateTimeOffset time,
    TimeWindowFilterSetting settings,
    out DateTimeOffset previous,
    out DateTimeOffset next)
    
This method will be true if either prev or next occurrence is found.

zhiyuanliang-ms: Updated.

``` C#
/// <summary>
/// Calculate the closest previous recurrence occurrence (if any) before the provided timestamp and the next occurrence (if any) after the provided timestamp.
/// <param name="time">A timestamp.</param>
/// <param name="settings">The settings of time window filter.</param>
/// <param name="prevOccurrence">The closest previous occurrence. Note that prev occurrence can be null even if the time is past the start date, because the recurrence range may have surpassed its end.</param>
/// <param name="nextOccurrence">The next occurrence. Note that next occurrence can be null even if the prev occurrence is not null, because the recurrence range may have reached its end.</param>
/// </summary>
public static void CalculateSurroundingOccurrences(DateTimeOffset time, TimeWindowFilterSettings settings, out DateTimeOffset? prevOccurrence, out DateTimeOffset? nextOccurrence)
{
    prevOccurrence = null;

    nextOccurrence = null;

    if (time < settings.Start.Value)
    {
        nextOccurrence = settings.Start.Value;

        return;
    }

    if (TryFindPreviousOccurrence(time, settings, out DateTimeOffset prev, out int numberOfOccurrences))
    {
        prevOccurrence = prev;

        RecurrencePattern pattern = settings.Recurrence.Pattern;

        if (pattern.Type == RecurrencePatternType.Daily)
        {
            nextOccurrence = prev.AddDays(pattern.Interval);
        }

        if (pattern.Type == RecurrencePatternType.Weekly)
        {
            nextOccurrence = CalculateWeeklyNextOccurrence(prev, settings);
        }

        RecurrenceRange range = settings.Recurrence.Range;

        if (range.Type == RecurrenceRangeType.EndDate)
        {
            if (nextOccurrence > range.EndDate)
            {
                nextOccurrence = null;
            }
        }

        if (range.Type == RecurrenceRangeType.Numbered)
        {
            if (numberOfOccurrences >= range.NumberOfOccurrences)
            {
                nextOccurrence = null;
            }
        }
    }
}
```

src/Microsoft.FeatureManagement/FeatureFilters/TimeWindowFilter.cs
jimmyca15: the key should be the flag name, and the cached value should have something to indicate that it's still relevant.

src/Microsoft.FeatureManagement/FeatureFilters/Recurrence/RecurrenceEvaluator.cs
```diff
@@ -0,0 +1,323 @@
...
+        const int DaysPerWeek = 7;
+
+        /// <summary>
+        /// Checks if a provided timestamp is within any recurring time window specified by the Recurrence section in the time window filter settings.
+        /// If the time window filter has an invalid recurrence setting, an exception will be thrown.
+        /// <param name="time">A timestamp.</param>
+        /// <param name="settings">The settings of time window filter.</param>
+        /// <returns>True if the timestamp is within any recurring time window, false otherwise.</returns>
+        /// </summary>
+        public static bool MatchRecurrence(DateTimeOffset time, TimeWindowFilterSettings settings)
```
jimmyca15, line 22:
```suggestion
        public static bool IsMatch(DateTimeOffset time, TimeWindowFilterSettings settings)
```

src/Microsoft.FeatureManagement/FeatureFilters/TimeWindowFilter.cs
```diff
@@ -56,7 +69,83 @@ public Task<bool> EvaluateAsync(FeatureFilterEvaluationContext context)
...
+                //
+                // The reference of the object will be used for hash key.
+                // If there is no pre-bounded settings attached to the context, there will be no cached filter settings and each call will have a unique settings object.
+                // In this case, the cache for recurrence settings won't work.
+                if (context.Settings == null)
+                {
+                    return Task.FromResult(RecurrenceEvaluator.MatchRecurrence(now, settings));
+                }
+
+                DateTimeOffset cachedTime = _recurrenceCache.GetOrAdd(
```
jimmyca15, line 90: We should be able to have something cleaner to follow where we don't invoke CalculateSurroundingOccurrences twice. Something like:

```
DateTimeOffset cachedtime;

if (!_recurrenceCache.TryGetValue(_settings, out cachedTime))
{
    _recurrenceCache.AddOrUpdate(_settings, GetCacheValue));

    cachedTime = _recurrenceCache[_settings];
}
```

src/Microsoft.FeatureManagement/FeatureFilters/TimeWindowFilter.cs
```diff
@@ -56,7 +69,83 @@ public Task<bool> EvaluateAsync(FeatureFilterEvaluationContext context)
...
+                        // There is no previous occurrence within the reccurrence range.
+                        return DateTimeOffset.MaxValue;
+                    });
+
+                if (now < cachedTime)
+                {
+                    return Task.FromResult(false);
+                }
+
+                if (now < cachedTime + (settings.End.Value - settings.Start.Value))
```
jimmyca15, line 116: how does that work if cachedtime is datetimeoffset.maxvalue?

jimmyca15, line 116: Seems we should impose a limit on how long this can be. I'm not sure if we do that already.

zhiyuanliang-ms: Now we will check whether the time window duration is shorter than how often it will recur. https://github.com/microsoft/FeatureManagement-Dotnet/pull/266#discussion_r1462221896

zhiyuanliang-ms: How about add an additional check for the time window duration, e.g. the duration can not longer than 10 years?

zhiyuanliang-ms: Do you think whether it is necessary to handle the potential System.ArgumentOutOfRangeException here? This operation may overflow the max datetime (i.e. 9999-12-31-23:59:59). But I don't believe this will happen in production.

src/Microsoft.FeatureManagement/FeatureFilters/TimeWindowFilter.cs
zhiyuanliang-ms: Updated as follow:
`RecurrenceEvaluator.CalculateClosestStart`
``` C#
/// <summary>
/// Calculate the start time of the closest active time window.
/// <param name="time">A timestamp.</param>
/// <param name="settings">The settings of time window filter.</param>
/// <returns>The start time of the closest active time window or null if the recurrence range surpasses its end.</returns>
/// </summary>
public static DateTimeOffset? CalculateClosestStart(DateTimeOffset time, TimeWindowFilterSettings settings)
{
    CalculateSurroundingOccurrences(time, settings, out DateTimeOffset? prevOccurrence, out DateTimeOffset? nextOccurrence);

    if (time < settings.Start.Value)
    {
        return nextOccurrence.Value;
    }

    if (prevOccurrence != null)
    {
        bool isWithinPreviousTimeWindow =
                time < prevOccurrence.Value + (settings.End.Value - settings.Start.Value);

        if (isWithinPreviousTimeWindow)
        {
            return prevOccurrence.Value;
        }

        if (nextOccurrence != null)
        {
            return nextOccurrence.Value;
        }
    }

    return null;
}
```

Recurrence evaluation in TimeWIndowFilter
``` C#
DateTimeOffset? closestStart = _recurrenceCache.GetOrAdd(settings, RecurrenceEvaluator.CalculateClosestStart(now, settings));

if (closestStart == null || now < closestStart.Value)
{
    return Task.FromResult(false);
}

if (now < closestStart.Value + (settings.End.Value - settings.Start.Value))
{
    return Task.FromResult(true);
}

closestStart = RecurrenceEvaluator.CalculateClosestStart(now, settings);

_recurrenceCache.AddOrUpdate(
    settings,
    (_) => throw new KeyNotFoundException(),
    (_, _) => closestStart);

if (closestStart == null || now < closestStart.Value)
{
    return Task.FromResult(false);
}

return Task.FromResult(now < closestStart.Value + (settings.End.Value - settings.Start.Value));
```

src/Microsoft.FeatureManagement/FeatureFilters/TimeWindowFilter.cs
zhiyuanliang-ms: > A feature flag can have multiple timewindow filters with different recurrence settings. If we want to use feature name as cache key, we will need additional information about the filter index.

Current `FeatureFIlterEvaluationContext` doesn't include such information.

zhiyuanliang-ms: Discussed offline. Will still use settings reference as cache key and use IMemoryCache instead of ConcurrentDictionary.

src/Microsoft.FeatureManagement/FeatureFilters/TimeWindowFilter.cs
jimmyca15: I think what we have now is fine then.

src/Microsoft.FeatureManagement/FeatureFilters/TimeWindowFilter.cs
```diff
@@ -56,7 +78,74 @@ public Task<bool> EvaluateAsync(FeatureFilterEvaluationContext context)
...
+            // Hit the first occurrence of the time window
+            if ((!settings.Start.HasValue || now >= settings.Start.Value) && (!settings.End.HasValue || now < settings.End.Value))
+            {
+                return Task.FromResult(true);
+            }
+
+            if (settings.Recurrence != null)
+            {
+                //
+                // The reference of the object will be used for hash key.
```
jimmyca15, line 91:
```suggestion
                // The reference of the object will be used for cache key.
```

src/Microsoft.FeatureManagement/FeatureFilters/TimeWindowFilter.cs
```diff
@@ -56,7 +78,74 @@ public Task<bool> EvaluateAsync(FeatureFilterEvaluationContext context)
...
+                // If there is no pre-bounded settings attached to the context, there will be no cached filter settings and each call will have a unique settings object.
+                // In this case, the cache for recurrence settings won't work.
+                if (context.Settings == null || Cache == null)
+                {
```
jimmyca15, line 95:
```suggestion
                if (Cache == null || context.Settings == null)
                {
```

src/Microsoft.FeatureManagement/FeatureFilters/TimeWindowFilter.cs
```diff
@@ -56,7 +78,74 @@ public Task<bool> EvaluateAsync(FeatureFilterEvaluationContext context)
...
+
+                Cache.Set(
+                    settings,
+                    closestStart,
+                    new MemoryCacheEntryOptions
+                    {
+                        SlidingExpiration = ParametersCacheSlidingExpiration,
+                        AbsoluteExpirationRelativeToNow = ParametersCacheAbsoluteExpirationRelativeToNow,
+                        Size = 1
+                    });
```
jimmyca15, line 138: closestStart = CalculateClosestStart(settings);

---

private static DateTimeOffset? CalculateClosestStart(TimeWindowFilterSettings)
{
    DateTimeOffset? closestStart = RecurrenceEvaluator.CalculateClosestStart(clock.UtcNow, settings);

    Cache.Set(
        settings,
        closestStart,
        new MemoryCacheEntryOptions
        {
            SlidingExpiration = ParametersCacheSlidingExpiration,
            AbsoluteExpirationRelativeToNow = ParametersCacheAbsoluteExpirationRelativeToNow,
            Size = 1
        });

    return closestStart;
}

src/Microsoft.FeatureManagement/FeatureManagementBuilder.cs
```diff
@@ -52,6 +53,38 @@ public FeatureManagementBuilder(IServiceCollection services)
             return this;
         }
 
+        public IFeatureManagementBuilder AddFeatureFilter<T>(Func<IServiceProvider, object> implementationFactory) where T : IFeatureFilterMetadata
```
jimmyca15, line 56: I would opt to keep this internal.

src/Microsoft.FeatureManagement/FeatureManagementBuilder.cs
```diff
@@ -67,5 +68,37 @@ public FeatureManagementBuilder(IServiceCollection services)
 
             return this;
         }
+
+        internal IFeatureManagementBuilder AddFeatureFilter<T>(Func<IServiceProvider, object> implementationFactory) where T : IFeatureFilterMetadata
```
jimmyca15, line 72: I realized the type is internal. This method should be public then.

jimmyca15 (approved):

tests/Tests.FeatureManagement/RecurrenceEvaluation.cs
```diff
@@ -0,0 +1,1740 @@
...
+                    }
+                },
+                null)
+            };
+
+            ConsumeEvalutationTestData(testData);
+        }
+
+        [Fact]
+        public async void RecurrenceEvaluationThroughCacheTest()
```
ivywei0125, line 1617: Shall we check if there's an entry in cache?

zhiyuanliang-ms: I just sent a PR #445 to test it.

PR #267, opened 2023-09-26
Updates dependencies. Removes ASP.NET support for netstandard2.0 & tests for netcoreapp2.1

rossgrambo:


rossgrambo:
I removed netcoreapp2.1 from tests, as FeatureManagement.AspNetCore was no longer able to build with on netcoreapp2.1 after removing netstandard2.0.

I did not remove netstandard2.0 from the normal FeatureManagement package. We could either:
1. Remove it there as well
2. Leave it and accept that tests won't run against it
3. Split the tests between ASP.NET tests and FeatureManagement tests and run on both.

src/Microsoft.FeatureManagement.AspNetCore/Microsoft.FeatureManagement.AspNetCore.csproj
```diff
@@ -11,7 +11,7 @@
   <Import Project="..\..\build\Versioning.props" />
 
   <PropertyGroup>
-    <TargetFrameworks>netstandard2.0;netcoreapp3.1;net5.0;net6.0</TargetFrameworks>
+    <TargetFrameworks>netcoreapp3.1;net5.0;net6.0</TargetFrameworks>
```
zhenlan, line 14: Please drop net5.0 too. It's a short-term supported version and is already out of support.

zhenlan:
> I removed netcoreapp2.1 from tests, as FeatureManagement.AspNetCore was no longer able to build with on netcoreapp2.1 after removing netstandard2.0.

Good. We should have dropped netcoreapp2.1 long ago.

> I did not remove netstandard2.0 from the normal FeatureManagement package. We could either:
> 
> 1. Remove it there as well
> 2. Leave it and accept that tests won't run against it
> 3. Split the tests between ASP.NET tests and FeatureManagement tests and run on both.

We need to keep `Microsoft.FeatureManagement` as a netstandard2.0 package. What's the problem with testing?

rossgrambo:
> What's the problem with testing?

The testing project has targets as well. If `target == netstandard2.0` then the testing project can't build, as it depends on the ASP.NET package (which can't built in netstandard2.0 without the deprecated packages).

If I remove `netstandard2.0` from testing, then we aren't running unit tests for that framework.

src/Microsoft.FeatureManagement.AspNetCore/Microsoft.FeatureManagement.AspNetCore.csproj
jimmyca15: netcoreapp3.1 is also out of support.

rossgrambo: Do we want to just fully update this to the current in-support frameworks?

jimmyca15:
> > What's the problem with testing?
> 
> The testing project has targets as well. If `target == netstandard2.0` then the testing project can't build, as it depends on the ASP.NET package (which can't built in netstandard2.0 without the deprecated packages).
> 
> If I remove `netstandard2.0` from testing, then we aren't running unit tests for that framework.

we need to separate the test project into two projects then.

src/Microsoft.FeatureManagement.AspNetCore/Microsoft.FeatureManagement.AspNetCore.csproj
jimmyca15: Yes.

rossgrambo:
Could we instead upgrade to netstandard2.1?

src/Microsoft.FeatureManagement.AspNetCore/Microsoft.FeatureManagement.AspNetCore.csproj
```diff
@@ -11,7 +11,7 @@
   <Import Project="..\..\build\Versioning.props" />
```
jimmyca15, line 12: @rossgrambo

/threadstart

Upgrading to netstandard2.1 is a big change. That means if anyone depends on us is netstandard2.0 the whole change needs to upgrade.

rossgrambo: Ah okay that makes sense. Should we instead be targeting both? As we should also be supporting netstandard2.1 for those customers?

jimmyca15: netstandard2.1 can reference netstandard2.1 packages.

zhenlan:
> > > What's the problem with testing?
> > 
> > 
> > The testing project has targets as well. If `target == netstandard2.0` then the testing project can't build, as it depends on the ASP.NET package (which can't built in netstandard2.0 without the deprecated packages).
> > If I remove `netstandard2.0` from testing, then we aren't running unit tests for that framework.
> 
> we need to separate the test project into two projects then.

Right, we should have separate test projects for MS.FM and MS.FM.ANC.

The test projects are applications. They are not libraries (or designed to be consumed by anyone else). We don't need to build test projects targeting netstandard. That said, you shouldn't have a problem once you separate the test project.

rossgrambo:
Updated this PR to target only supported frameworks and separated the FeatureManagement and FeatureManagement.AspNetCore test packages.

examples/ConsoleApp/ConsoleApp.csproj
```diff
@@ -2,7 +2,7 @@
 
   <PropertyGroup>
     <OutputType>Exe</OutputType>
-    <TargetFramework>net6.0</TargetFramework>
+    <TargetFramework>net6.0;</TargetFramework>
```
jimmyca15, line 5: what's the extra simcolon?

jimmyca15 (approved):

jimmyca15:
PR build is failing. The container we use to build doesn't have .net 7 sdk. Seems like we need a new container.

examples/ConsoleApp/ConsoleApp.csproj
rossgrambo: It's pointless. Just added it so they all matched. Can take it out if you prefer 😄

jimmyca15: Yes, please remove.

jimmyca15 (approved):

Issue #269, opened 2023-10-03
Update docs with latest recommendations for defining and consuming feature flags


brental:
In [this comment](https://github.com/microsoft/FeatureManagement-Dotnet/issues/268#issuecomment-1743344584) it is stated that the recommended approach for defining and consuming feature flags is no longer to use an enum. Instead, it is recommended to use constants and that the docs need to be updated to reflect this. So, this issue is to get the docs updated with the new recommended approach.

jimmyca15:
Updated. Rather than us making a recommendation we have just replaced the references in documentation to string literals.

Thanks for bringing this to our attention.

brental:
Thanks @jimmyca15 @zhiyuanliang-ms, just using strings without a recommendation makes sense. 

However, I did find that there is still one place in the docs where nameof is still used which is in the section on the tag helper. That should probably be updated to just use a string as well?

zhiyuanliang-ms:
@brental Thank you for catching it. That's a mistake.

PR #275, opened 2023-10-09
Update the recommended way to reference feature flag name in README

zhiyuanliang-ms:
Issue #269 

Change the recommended approach to reference the feature flag names from enum to const strings.


README.md
```diff
@@ -144,12 +144,12 @@ In the above example, `FeatureW` specifies a `RequirementType` of `All`, meaning
 To make it easier to reference these feature flags in code, we recommend to define feature flag variables like below.
```
jimmyca15, line 144: > we recommend to define feature flag variables like below

I'm thinking, is it worth a recommendation? This is fairly basic definition of strings. I would imagine just removing this section would be fine.

The only issue I can think with that is that we reference `MyFeatureFlags.FeatureX` and others below. Perhaps though, it's clear enough that those references should be the names of the flags.

jimmyca15:
References of `nameof()` should be removed.

README.md
zhiyuanliang-ms: If we want to remove the recommendation, maybe we can just use string "FeatureX" instead of the reference.
If I am a user, I can understand that this is just the name of the feature flag which is a string variable. 
But there is a customer who raised an issue for this. He expected us to update the recommendation.

README.md
```diff
@@ -141,17 +141,7 @@ In the above example, `FeatureW` specifies a `RequirementType` of `All`, meaning
...
-``` C#
-// Define feature flags using const strings
-public static class MyFeatureFlags
-{
-    public const string FeatureT = "FeatureT";
-    public const string FeatureU = "FeatureU";
-    public const string FeatureV = "FeatureV";
-}
-```
+In the following code examples, MyFeatureFlags.FeatureX which is a const string property of the MyFeatureFlags class, is used for referencing the name of feature flag "FeatureX".
```
jimmyca15, line 144: What do you think about just removing MyFeatureFlags altogether and using string literals?

zhiyuanliang-ms: Sure. That's my original thought https://github.com/microsoft/FeatureManagement-Dotnet/pull/275#discussion_r1351194449

jimmyca15: Oh, sorry. Missed it.

jimmyca15 (approved):

PR #276, opened 2023-10-11
Add option to allow scoped feature manager

zhiyuanliang-ms:


rossgrambo:
Is this PR on hold as we might just expose the classes publicly? Or do we plan to still merge this as well.

zhiyuanliang-ms:
Last Thursday, we discussed about it, so far, we prefer to have AddScopedFeatureManagement instead of expose Feature Manager public. But we didn't have the chance to discuss the design of the cache during the meeting last week.

zhiyuanliang-ms:
Close this PR because of discussion of re-design the constructor of Feature Manager

PR #277, opened 2023-10-12
Recreates examples with latest packages

rossgrambo:


examples/ConsoleApp/AccountServiceContext.cs
```diff
@@ -1,12 +1,7 @@
 ﻿// Copyright (c) Microsoft Corporation.
 // Licensed under the MIT license.
 //
-using Consoto.Banking.AccountService.FeatureFilters;
-
-namespace Consoto.Banking.AccountService
```
jimmyca15, line 6: why'd the namespace go away?

rossgrambo: The ConsoleApp and other templates now defaults to https://aka.ms/new-console-template

Namespace automatically matches the project name. I left the namespace, "class Program", "Main" scaffolding in some of the Examples, but wanted to use the new console template in at least a few places to better match the new experience.

jimmyca15 (approved):

examples/FeatureFlagDemo/FeatureNotEnabledDisabledHandler.cs
```diff
@@ -17,7 +17,7 @@ public Task HandleDisabledFeatures(IEnumerable<string> features, ActionExecuting
         {
             var result = new ViewResult()
             {
-                ViewName = "Views/Shared/FeatureNotEnabled.cshtml",
+                ViewName = "Pages/Shared/FeatureNotEnabled.cshtml",
```
zhiyuanliang-ms, line 20: There is no "Pages" folder.

PR #284, opened 2023-10-13
Adds example for Application Insights Publisher

rossgrambo:
image


examples/EvaluationDataToApplicationInsights/appsettings.Development.json
zhiyuanliang-ms, line 1: Is this file necessary? I didn't see it in our other example projects.

examples/EvaluationDataToApplicationInsights/Program.cs
zhiyuanliang-ms, line 1: We didn't use top-level statements in other example projects. Should we keep the code style consistent?

examples/EvaluationDataToApplicationInsights/Pages/Shared/_Layout.cshtml
```diff
@@ -0,0 +1,95 @@
...
+                </button>
+                <div class="navbar-collapse collapse d-sm-inline-flex justify-content-between">
+                    <ul class="navbar-nav flex-grow-1">
+                        <li class="nav-item">
+                            <a class="nav-link text-dark" asp-area="" asp-page="/Index">Home</a>
+                        </li>
+                        <li class="nav-item">
+                            <a class="nav-link text-dark" asp-area="" asp-page="/Checkout">Checkout</a>
+                        </li>
+                        <feature name="FeatureA">
```
zhiyuanliang-ms, line 61: I think it would be better to move this into Index.cshtml
Something like:
<feature name="FeatureA">
    <p>Since you are seeing this, that means the "FeatureA" is enabled for the current user because he/she falls within the default rollout percentage of the targeting filter.</p>
</feature>
In this way, it would be easier for user to understand how this example project works.

examples/EvaluationDataToApplicationInsights/appsettings.Development.json
rossgrambo: > Is this file necessary?

No, but I think it's better to use the template and change as little as possible.

examples/EvaluationDataToApplicationInsights/Pages/Shared/_Layout.cshtml
rossgrambo: We can do that. Currently they're in _Layout so they show on the top for all pages (Index, Checkout, and Error). Moving it to Index means you'll only see it on that page. Adding pages for each feature feels like too much for this example.

The main flow of this example is to Checkout. So I definitely want it present on that page. Would you prefer I add your suggestion ^ to the checkout page, add your example to only index, or leave it in the shared header?

examples/EvaluationDataToApplicationInsights/appsettings.json
```diff
@@ -0,0 +1,53 @@
+{
+  "Logging": {
+    "LogLevel": {
+      "Default": "Information",
+      "Microsoft.AspNetCore": "Warning"
+    }
+  },
+  "AllowedHosts": "*",
+  "ApplicationInsights": {
+    "ConnectionString": "<connection-string>"
```
jimmyca15, line 10: Can you put a readme of what a user needs to do to get this sample working?

rossgrambo: Yeah, good idea

examples/EvaluationDataToApplicationInsights/Pages/Index.cshtml
```diff
@@ -0,0 +1,10 @@
+﻿@page
+@model IndexModel
+@{
+    ViewData["Title"] = "Home page";
+}
+
+<div class="text-center">
+    <h1 class="display-4">Welcome <script>document.write(getCookie("username"));</script></h1>
```
jimmyca15, line 8: Given this is a razor page, I'd recommend putting the username into the view data and using razor syntax to inject it into the DOM.

examples/EvaluationDataToApplicationInsights/Pages/Shared/_Layout.cshtml
```diff
@@ -0,0 +1,95 @@
...
+                                <a class="nav-link text-dark" asp-area="" asp-page="/Home">Feature A</a>
+                            </li>
+                        </feature>
+                        <feature name="FeatureB">
+                            <li class="nav-item">
+                                <a class="nav-link text-dark" asp-area="" asp-page="/Home">Feature B</a>
+                            </li>
+                        </feature>
+                        <li style="display: flex;">
+                            <button onclick="document.cookie = ' username=' + Math.floor(Math.random() * 100000);deleteCookie('ai_user');deleteCookie('ai_session');window.location=window.location.origin;">
```
jimmyca15, line 72: This refreshes the page right? Seems like a good candidate to do this server side rather than custom java script.

examples/EvaluationDataToApplicationInsights/appsettings.json
```diff
@@ -0,0 +1,53 @@
...
+              "Groups": [],
+              "DefaultRolloutPercentage": 50
+            }
+          }
+        }
+      ]
+    },
+    "FeatureB": {
+      "TelemetryEnabled": true,
+      "Label": "A Label",
```
jimmyca15, line 32: Shouldn't this be in TelemetryMetadata?

examples/EvaluationDataToApplicationInsights/Pages/Checkout.cshtml
```diff
@@ -0,0 +1,12 @@
+﻿@page
+@model CheckoutModel
+@{
+    ViewData["Title"] = "Checkout";
+}
+<h1>@ViewData["Title"]</h1>
+
+<p>Click Below To Check Out!</p>
+
+<button id="checkout" name="checkout" onclick="appInsights.trackEvent({ name: 'checkout', properties: { 'success' : 'yes' } }); appInsights.trackMetric({ name: 'checkoutAmount', average: Math.floor(Math.random() * 100) }); alert('Checked Out!');">
```
jimmyca15, line 10: Also worth calling out the events/metrics you have wired up in the README. This are of great interest when running the example. Perhaps even how to query them.

jimmyca15:
I believe you mentioned something about a specific usage of cookies being important with this publisher. Is that in this example?

examples/EvaluationDataToApplicationInsights/appsettings.json
rossgrambo: Added in https://github.com/microsoft/FeatureManagement-Dotnet/pull/284/commits/9a4370bbb83f522d57b789a1d0df549d66c880bc?short_path=dc904d4#diff-dc904d4dce5d3420435595d0fb7154a78d5cd82d78e02cc0e1d262266869af20

examples/EvaluationDataToApplicationInsights/appsettings.json
```diff
@@ -0,0 +1,53 @@
...
+  "FeatureManagement": {
+    "FeatureA": {
+      "TelemetryEnabled": true,
+      "EnabledFor": [
+        {
+          "Name": "Targeting",
+          "Parameters": {
+            "Audience": {
+              "Users": [
+                "ross"
```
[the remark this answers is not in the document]
rossgrambo: Editing the username cookie in the browser. Perhaps I should just exclude individual targeting from the example.

examples/EvaluationDataToApplicationInsights/Pages/Checkout.cshtml
rossgrambo: Good call. Added a snippet about it.

examples/EvaluationDataToApplicationInsights/appsettings.json
rossgrambo: Yes. Nothing was being emitted for these fields after the TelemetryMetadata change. Fixed.

examples/EvaluationDataToApplicationInsights/Pages/Shared/_Layout.cshtml
rossgrambo: Yes. This is kind of a hard refresh to simulate a new user & a new browser. It generates a new username, clears the App Insights cookies, and then refreshes.

examples/EvaluationDataToApplicationInsights/Pages/Shared/_Layout.cshtml
zhiyuanliang-ms: Sounds good!

examples/EvaluationDataToApplicationInsights/appsettings.json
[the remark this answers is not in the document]
rossgrambo: > Also, doesn't seem like having a hardcoded user already included may be desirable.

I wanted to see what happened for the not logged-in case. We could adjust it, but I think it's nice to show events flowing for guest users.

rossgrambo:
Here's the snippet about cookie magic. It's not a lot: https://github.com/microsoft/FeatureManagement-Dotnet/pull/284/files#diff-dc904d4dce5d3420435595d0fb7154a78d5cd82d78e02cc0e1d262266869af20R46

examples/EvaluationDataToApplicationInsights/Pages/Shared/_Layout.cshtml
rossgrambo: Moved this to server side

examples/EvaluationDataToApplicationInsights/Program.cs
```diff
@@ -0,0 +1,46 @@
...
+var builder = WebApplication.CreateBuilder(args);
+
+// Add services to the container.
+builder.Services.AddRazorPages();
+builder.Services.AddHttpContextAccessor();
+builder.Services.AddApplicationInsightsTelemetry();
+builder.Services.AddSingleton<IHttpContextAccessor, HttpContextAccessor>();
+builder.Services.AddSingleton<ITargetingContextAccessor, HttpContextTargetingContextAccessor>();
+builder.Services.AddFeatureManagement()
+    .AddFeatureFilter<TargetingFilter>()
```
zhiyuanliang-ms, line 20: Should we use WithTargeting<>() since this is the lastest example?

Issue #248, opened 2023-07-18
IgnoreMissingFeatureFilters should not be false by default


Rutix:
Hi,

I was wondering why `IgnoreMissingFeatureFilters` is by default `false`? This is super dangerous when you have multiple services and can lead to significant production problems. We have the following situation:

* We use Azure App Configuration Feature Manager to manage the whole load of feature toggles.
* All services connect to that Azure App Configuration
* We have configured in all our services that the feature toggles are refreshed
* In most services we had the following configured: `services.AddFeatureManagement();`
* Someone on production saw the "Microsoft.TimeWindow" feature filter and thought: "Hey that's handy!" and configured it. This caused all services to fail and throw exceptions which caused production outages.

Apart from that we should have tested this in a development environment, it is my opinion that "forgotten configuration" should never break an application. Especially since this can be a full sneaky mode and be caused by someone else.

I found #13 where this is discussed but I think the perspective of the default isn't discussed very well there.

Rutix:
> > I was wondering why IgnoreMissingFeatureFilters is by default false?
> 
> It is default to false to prevent unintentionally missing feature filter registration. This way when applications are being built-out no misconfiguration goes unnoticed.
> 
> The problem you are mentioning is valid, but it does happen because of touching live configuration in production. And as you mentioned that configuration update wasn't verified in any development environment. It's difficult to design for a scenario like that where so many things can go wrong. But, we offer the IgnoreMissingFeatureFilters flag if that is a workflow that is desired.
> 
> It was a pros and cons decision and in the end the decision was to make it easier to avoid misconfiguration in development.

I understand where you are coming from but you are basically saying that the risk in development trumps the risk in production. I wholeheartedly disagree with that. If you forget in development then you will find out quickly and at worst might have to make a change to fix it for yourself. In production (and especially with a lot of teams and big organizations) someone might decide to use the TimeWindow to roll out on a certain date in production (while it has been tested in development with the feature toggle on true) and that can cause downtime.

It's basically an argument of debugging/tracking why something doesn't work vs a potential production shutdown because of the defaults. I hope you can agree that the second one is a way bigger risk. So I hope you and your team want to reconsider the default setting of this one.

zhiyuanliang-ms:
@Rutix After discussion, we still want to remain the default value of IgnoreMissingFeatureFilter as false. But we plan to register the built-in filter like TimeWindow and Targeting filter by default after calling AddFeatureManagement().

[...]

PR #287, opened 2023-10-18
Register built-in filter by default

zhiyuanliang-ms:
Mitigate issue #248


jimmyca15:
I don't expect any issues, but can you confirm no issues occur if someone manually adds these with

`featureManagementBuilder.AddFeatureFilter` or `services.AddSingleton<TargetingFilter>`

src/Microsoft.FeatureManagement/ServiceCollectionExtensions.cs
```diff
@@ -58,6 +58,12 @@ public static IFeatureManagementBuilder AddFeatureManagement(this IServiceCollec
 
             services.TryAddScoped<IVariantFeatureManagerSnapshot>(sp => sp.GetRequiredService<FeatureManagerSnapshot>());
 
+            services.AddSingleton<PercentageFilter>();
```
jimmyca15, line 61: Probably a good idea to call our `AddFeatureFilter` overload. It only adds the filter if it hasn't already been added.

zhenlan, line 61: We probably shouldn't add PercentageFilter. It was created to showcase how to write a feature filter. We may want to drop it one day.

jimmyca15: Discussed with @zhenlan offline. My concern is that if we have the filter in the library, it makes sense to add it by default. If we want to exclude it by default, then most likely we should just remove it from the library as well.

Given that we do have current usage of this filter, it doesn't make sense to remove it at this point. We agreed that we should add the percentage filter by default.

zhiyuanliang-ms:
@jimmyca15
Currently, we have ContextualTargetingFilter and TargetingFilter which both have alias "Microsoft.Targeting". They cannot be registered at the same time.
We need to allow different types of filter to share the same alias first (PR #262)

src/Microsoft.FeatureManagement/ServiceCollectionExtensions.cs
zhiyuanliang-ms: Oh, sorry. This is my fault. I found that `services.AddSingleton<PercentageFilter>();` is completly wrong. Because the feature filter will be injected as IFeatureFilterMetadata.

jimmyca15: Test didn't catch that?

zhiyuanliang-ms: All previous testcases called AddFeatureFilter() by themselves. I have not written any new testcases.

jimmyca15: Sounds like it would be worth to have a test that would've caught the issue.

src/Microsoft.FeatureManagement/ServiceCollectionExtensions.cs
```diff
@@ -33,7 +34,17 @@ public static IFeatureManagementBuilder AddFeatureManagement(this IServiceCollec
 
             services.AddScoped<IFeatureManagerSnapshot, FeatureManagerSnapshot>();
 
-            return new FeatureManagementBuilder(services);
+            services.TryAddSingleton<ITargetingContextAccessor, EmptyTargetingContextAccessor>();
```
jimmyca15, line 37: Shouldn't be here.

src/Microsoft.FeatureManagement/ServiceCollectionExtensions.cs
```diff
@@ -33,7 +34,17 @@ public static IFeatureManagementBuilder AddFeatureManagement(this IServiceCollec
...
+
+            var builder = new FeatureManagementBuilder(services);
+
+            builder.AddFeatureFilter<PercentageFilter>();
+
+            builder.AddFeatureFilter<TimeWindowFilter>();
+
+            builder.AddFeatureFilter<TargetingFilter>();
```
jimmyca15, line 45: We discussed only adding ContextualTargetingFilter by default since TargetingFilter requires ITargetingContextAccessor implementation.

jimmyca15:
```
services
    .AddFeatureManagement()
    .WithTargeting<MyTargetingContextAccessor>(); 
```

zhiyuanliang-ms: Yes. This commit was before the meeting on last Thursday. I did not update it.

zhiyuanliang-ms:
Testcases are not updated. After PR #262 is approved, more testcases will be added and AddFeatureFilter<> for built-in filters in the testcases will be removed.

src/Microsoft.FeatureManagement/IFeatureManagementBuilder.cs
```diff
@@ -30,5 +31,12 @@ public interface IFeatureManagementBuilder
         /// <typeparam name="T">An implementation of <see cref="ISessionManager"/></typeparam>
         /// <returns>The feature management builder.</returns>
         IFeatureManagementBuilder AddSessionManager<T>() where T : ISessionManager;
+
+        /// <summary>
+        /// Adds an <see cref="ITargetingContextAccessor"/> to be used for targeting.
+        /// </summary>
+        /// <typeparam name="T">An implementation of <see cref="ITargetingContextAccessor"/></typeparam>
+        /// <returns>The feature management builder.</returns>
+        IFeatureManagementBuilder WithTargeting<T>() where T : ITargetingContextAccessor;
```
jimmyca15, line 40: Adding a method on an interface is a breaking change. This should be an extension method. I believe we already have `IFeatureManagementBuilderExtensions`

jimmyca15 (approved):

PR #288, opened 2023-10-18
prevent the usage of colon in feature flag name

zhiyuanliang-ms:


jimmyca15:
Should be against main branch.

jimmyca15 (approved):

PR #290, opened 2023-10-24
Merges variants and isenabled paths. Adds variant reason field.

rossgrambo:
## Logic Merge

This PR merges the logic of `IsEnabledAsync` and `GetVariantAsync` into a single `EvaluateFeature`. 

`EvaluateFeature` takes and returns an `EvaluationEvent`, populating the fields during evaluation.

 Why?
1. Deduplicate logic
2. Due to the `Reason` field, a persistent way to return a value (like variant) and additional data (reason) was needed throughout the evaluation.

## Variant Assignment Reason

`VariantAssignmentReason` is the proposed field to satisfy the requirements of "Reason". It is a simple string reflecting why a variant was assigned. The planned metrics consumer only cares about Variants. This means we can add this field now, while reserving the right to add an "EnabledReason" later, that may contain additional information on filters.

The string values for this field are:
1. `DefaultWhenDisabled`
1. `DefaultWhenEnabled`
1. `User`
1. `Group`
1. `Percentile`


src/Microsoft.FeatureManagement/FeatureManager.cs
```diff
@@ -71,92 +71,148 @@ private class ConfigurationCacheItem
...
 
-                if (featureDefinition.Variants != null && featureDefinition.Variants.Any() && featureDefinition.Allocation != null)
+        private async ValueTask<Variant> GetVariantEvaluation(string feature, TargetingContext context, bool useContext, CancellationToken cancellationToken)
+        {
+            EvaluationEvent evaluationEvent = new EvaluationEvent
+            {
+                FeatureDefinition = await GetFeatureDefinition(feature).ConfigureAwait(false)
+            };
+
+            await EvaluateFeature(evaluationEvent, context, useContext, cancellationToken);
```
amerjusupovic, line 136: It looks like `EvaluateFeature` returns an `EvaluationEvent` but the return value isn't used anywhere it's called.

src/Microsoft.FeatureManagement/FeatureManager.cs
```diff
@@ -71,92 +71,148 @@ private class ConfigurationCacheItem
...
+
+                //
+                // Determine Variant
+                VariantDefinition variantDefinition;
+
+                if (evaluationEvent.FeatureDefinition.Allocation == null || (!evaluationEvent.FeatureDefinition.Variants?.Any() ?? false))
+                {
+                    variantDefinition = null;
+
+                    evaluationEvent.VariantReason = "No Allocation or Variants";
```
amerjusupovic, line 157: Not sure if this is just unclear to me, but the message sounds like the feature is missing both allocation and variants

rossgrambo: Agreed, and it's supposed to be either. Any suggestions for a better one? Perhaps "Missing Allocation or Variants"...?

src/Microsoft.FeatureManagement/FeatureManager.cs
rossgrambo: Yeah that's accurate. I wanted to indicate that it's the resulting object (all of the results are on the object), but it's the same object that's passed in, so these could all be void.

src/Microsoft.FeatureManagement/FeatureManager.cs
amerjusupovic: Yea, I like "Missing Allocation or Variants" more.

src/Microsoft.FeatureManagement/Telemetry/EvaluationEvent.cs
```diff
@@ -24,5 +24,10 @@ public class EvaluationEvent
         /// The variant given after evaluation.
         /// </summary>
         public Variant Variant { get; set; }
+
+        /// <summary>
+        /// The reason the variant was given.
+        /// </summary>
+        public string VariantReason { get; set; }
```
jimmyca15, line 31: I think enum would be more fitting here.

jimmyca15: `public AllocationReason AllocationReason`

```
public enum AllocationReason
{
    None,
    DisabledDefault,
    EnabledDefault,
    User,
    Group,
    Percentile
}
```

src/Microsoft.FeatureManagement/FeatureManager.cs
```diff
@@ -71,92 +71,148 @@ private class ConfigurationCacheItem
...
         }
 
         public ValueTask<bool> IsEnabledAsync<TContext>(string feature, TContext appContext, CancellationToken cancellationToken)
         {
-            return IsEnabledWithVariantsAsync(feature, appContext, useAppContext: true, cancellationToken);
+            return IsEnabledEvaluation(feature, appContext, useAppContext: true, cancellationToken);
         }
 
-        private async ValueTask<bool> IsEnabledWithVariantsAsync<TContext>(string feature, TContext appContext, bool useAppContext, CancellationToken cancellationToken)
+        private async ValueTask<bool> IsEnabledEvaluation<TContext>(string feature, TContext appContext, bool useAppContext, CancellationToken cancellationToken)
```
zhiyuanliang-ms, line 92: According to https://learn.microsoft.com/en-us/dotnet/standard/design-guidelines/names-of-type-members#names-of-methods
The method name should be verb. Is the name "IsEnabledEvaluation" a noun?

I suggest "EvaluateFeature".

src/Microsoft.FeatureManagement/FeatureManager.cs
```diff
@@ -241,77 +241,108 @@ public ValueTask<Variant> GetVariantAsync(string feature, TargetingContext conte
...
-                isFeatureEnabled = await IsEnabledAsync(featureDefinition, appContext, useAppContext, cancellationToken).ConfigureAwait(false);
+                FeatureDefinition = await GetFeatureDefinition(feature).ConfigureAwait(false)
+            };
+
+            await EvaluateFeature(evaluationEvent, appContext, useAppContext, cancellationToken);
+
+            return evaluationEvent.IsEnabled;
+        }
+
+        private async Task<EvaluationEvent> EvaluateFeature<TContext>(EvaluationEvent evaluationEvent, TContext context, bool useContext, CancellationToken cancellationToken)
```
zhiyuanliang-ms, line 271: evaluationEvent will be returned eventually. Would it be better to use `ref EvaluationEvent evaluationEvent`

src/Microsoft.FeatureManagement/FeatureManager.cs
zhiyuanliang-ms: Since it is the same object that's passed in, we should use `ref` parameter. Unfortunately, this is an async method, we cannot use `ref` parameters.
I suggest:
```C#
private async Task<EvaluationEvent> EvaluateFeature<TContext>(string feature, TContext context, bool useContext, CancellationToken cancellationToken)
{
   EvaluationEvent evaluationEvent = new EvaluationEvent
   {
       FeatureDefinition = await GetFeatureDefinition(feature).ConfigureAwait(false)
   };

   ...
}
```

src/Microsoft.FeatureManagement/FeatureManager.cs
jimmyca15: We may want to avoid this change given IFeatureManager's main method is `IsEnabled`. That's not a verb. I don't want to stray to far from that with the internal methods.

jimmyca15: You raise a good point, perhaps we could have named `IsEnabled` something like `GetXyz` in the beginning.

zhiyuanliang-ms:
ApplicationInsightsTelemetryPublisher should add "AssignmentReason" field.

src/Microsoft.FeatureManagement/FeatureManager.cs
```diff
@@ -541,95 +530,103 @@ private async ValueTask<TargetingContext> ResolveTargetingContextAsync(Cancellat
...
-                variantDefinition = await AssignVariantAsync(featureDefinition, context, cancellationToken).ConfigureAwait(false);
+                variantDefinition = await AssignVariantAsync(evaluationEvent, context, cancellationToken).ConfigureAwait(false);
             }
 
             if (variantDefinition == null)
             {
-                variantDefinition = featureDefinition.Variants.FirstOrDefault((variant) => variant.Name == featureDefinition.Allocation.DefaultWhenEnabled);
+                variantDefinition = evaluationEvent.FeatureDefinition.Variants.FirstOrDefault((variant) => variant.Name == evaluationEvent.FeatureDefinition.Allocation.DefaultWhenEnabled);
+
+                evaluationEvent.AssignmentReason = AssignmentReason.EnabledDefault;
```
zhiyuanliang-ms, line 546: If there is no default variant for the enabled feature, should the reason be None or EnabledDefault?

src/Microsoft.FeatureManagement/FeatureManager.cs
```diff
@@ -241,77 +241,108 @@ public ValueTask<Variant> GetVariantAsync(string feature, TargetingContext conte
...
+                    evaluationEvent.AssignmentReason = AssignmentReason.None;
+                }
+                else
+                {
+                    if (!evaluationEvent.IsEnabled)
                     {
-                        variantDefinition = featureDefinition.Variants.FirstOrDefault((variant) => variant.Name == featureDefinition.Allocation.DefaultWhenDisabled);
+                        variantDefinition = evaluationEvent.FeatureDefinition.Variants.FirstOrDefault((variant) => variant.Name == evaluationEvent.FeatureDefinition.Allocation.DefaultWhenDisabled);
+
+                        evaluationEvent.AssignmentReason = AssignmentReason.DisabledDefault;
```
zhiyuanliang-ms, line 295: If there is no default variant for the disabled feature, should the reason be None or DisabledDefault?

src/Microsoft.FeatureManagement/FeatureManager.cs
```diff
@@ -241,77 +241,108 @@ public ValueTask<Variant> GetVariantAsync(string feature, TargetingContext conte
...
+                            evaluationEvent.IsEnabled = false;
                         }
                     }
                 }
             }
 
             foreach (ISessionManager sessionManager in _sessionManagers)
             {
-                await sessionManager.SetAsync(feature, isFeatureEnabled).ConfigureAwait(false);
+                await sessionManager.SetAsync(evaluationEvent.FeatureDefinition.Name, evaluationEvent.IsEnabled).ConfigureAwait(false);
```
zhiyuanliang-ms, line 337: Need null check for evaluationEvent.FeatureDefinition here

src/Microsoft.FeatureManagement/FeatureManager.cs
```diff
@@ -241,77 +241,108 @@ public ValueTask<Variant> GetVariantAsync(string feature, TargetingContext conte
...
             }
 
-            if (featureDefinition.TelemetryEnabled)
+            if (evaluationEvent.FeatureDefinition.TelemetryEnabled)
```
zhiyuanliang-ms, line 340: Need null check for evaluationEvent.FeatureDefinition here

src/Microsoft.FeatureManagement/FeatureManager.cs
amerjusupovic: I think the only additional information `EnabledDefault` gives is that the feature was enabled, and you can already check if the feature was enabled in the telemetry. I also think `None` makes more sense since there's no variant assigned.

src/Microsoft.FeatureManagement/FeatureManager.cs
```diff
@@ -241,77 +251,100 @@ public ValueTask<Variant> GetVariantAsync(string feature, TargetingContext conte
...
+                }
+                else
                 {
-                    if (!isFeatureEnabled)
+                    if (!evaluationEvent.IsEnabled)
                     {
-                        variantDefinition = featureDefinition.Variants.FirstOrDefault((variant) => variant.Name == featureDefinition.Allocation.DefaultWhenDisabled);
+                        variantDefinition = evaluationEvent.FeatureDefinition.Variants.FirstOrDefault((variant) => variant.Name == evaluationEvent.FeatureDefinition.Allocation.DefaultWhenDisabled);
+
+                        if (variantDefinition != null)
```
jimmyca15, line 288: And if `variantDefinition` is null, then AssignmentReason seems to be the default value. Which looks to be AssignmentReason.None. But on line 280, the assignment reason is explicitly set to none. I expect we should stick to one strategy, either explicitly set assignment reason to none, or let it be none due to being the default.

src/Microsoft.FeatureManagement/FeatureManager.cs
```diff
@@ -541,101 +532,98 @@ private async ValueTask<TargetingContext> ResolveTargetingContextAsync(Cancellat
...
+                            return new ValueTask<VariantDefinition>(
+                                evaluationEvent.FeatureDefinition
+                                    .Variants
+                                    .FirstOrDefault((variant) => variant.Name == percentile.Variant));
+                        }
                     }
                 }
             }
 
+            if (variant == null)
```
jimmyca15, line 622: I would suggest to let the caller decide to set the assignment reason to none.

jimmyca15: Actually, I don't see us ever setting `variant` to anything other than null when we first declare it. Am I missing something?

src/Microsoft.FeatureManagement/FeatureManager.cs
```diff
@@ -541,101 +532,98 @@ private async ValueTask<TargetingContext> ResolveTargetingContextAsync(Cancellat
...
-                variantDefinition = featureDefinition.Variants.FirstOrDefault((variant) => variant.Name == featureDefinition.Allocation.DefaultWhenEnabled);
-            }
-
-            return variantDefinition;
-        }
-
-        private ValueTask<VariantDefinition> AssignVariantAsync(FeatureDefinition featureDefinition, TargetingContext targetingContext, CancellationToken cancellationToken)
+        private ValueTask<VariantDefinition> AssignVariantAsync(EvaluationEvent evaluationEvent, TargetingContext targetingContext, CancellationToken cancellationToken)
         {
             VariantDefinition variant = null;
```
jimmyca15, line 538: if targetingContext is null, don't call this method.

src/Microsoft.FeatureManagement.Telemetry.ApplicationInsights/ApplicationInsightsTelemetryPublisher.cs
```diff
@@ -41,6 +41,11 @@ public ValueTask PublishEvent(EvaluationEvent evaluationEvent, CancellationToken
                 properties["Variant"] = evaluationEvent.Variant.Name;
             }
 
+            if (evaluationEvent.AssignmentReason != AssignmentReason.None)
+            {
+                properties["AssignmentReason"] = evaluationEvent.AssignmentReason.ToString();
```
jimmyca15, line 46: Can you create a private method

```
private static string ToString(AssignemntReason reason);
```

So we don't rely on the enum name to determine our telemetry output. We wouldn't want code refactoring to introduce changes into our communication protocol.

src/Microsoft.FeatureManagement/FeatureManager.cs
zhiyuanliang-ms: Sounds good. I vote for explicity setting the assignment reason to none.

src/Microsoft.FeatureManagement/Telemetry/AssignmentReason.cs
```diff
@@ -0,0 +1,41 @@
...
+    /// </summary>
+    public enum AssignmentReason
+    {
+        /// <summary>
+        /// No variant is assigned.
+        /// </summary>
+        None,
+
+        /// <summary>
+        /// Variant is assigned by default after processing the user/group/percentile allocation, when the feature flag is disabled.
```
zhenlan, line 17: When the feature flag is disabled, we don't process "the user/group/percentile allocation".

src/Microsoft.FeatureManagement/Telemetry/AssignmentReason.cs
```diff
@@ -0,0 +1,41 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+//
+namespace Microsoft.FeatureManagement.Telemetry
+{
+    /// <summary>
+    /// The reason the variant was assigned during the evaluation of a feature.
+    /// </summary>
+    public enum AssignmentReason
```
zhenlan, line 9: Not sure if it's just me feeling this confusing. No where in this enum the variant is mentioned. For example, `AssignmentReason.None` sounds like it's saying "something is assigned with no reason". Don't ask me to guess what `AssignmentReason.DisabledDefault` means :) The list is great, but can we come up with something that is more intuitive?

zhiyuanliang-ms: We can change the enum name to `VariantAssignmentReason`. Is that enough?

AssignmentReason.None is the default value of the enum. When there is no variant assigned, the reason will be `None`. Maybe, we can change it to `NotAssigned`?

zhiyuanliang-ms: `DisabledDefault` is in our feature flag schema. Only when you config the `DisabledDefault` section, you can recieve this reason field in the telemetry.

src/Microsoft.FeatureManagement.Telemetry.ApplicationInsights/ApplicationInsightsTelemetryPublisher.cs
```diff
@@ -41,6 +41,11 @@ public ValueTask PublishEvent(EvaluationEvent evaluationEvent, CancellationToken
                 properties["Variant"] = evaluationEvent.Variant.Name;
             }
 
+            if (evaluationEvent.AssignmentReason != AssignmentReason.None)
```
zhiyuanliang-ms, line 44: I want to call out that the ApplicationInsightsTelemetryPublisher will not send AssignmentReason telemetry if the AssignmentReason is None (which will only occur when the variant is null).

src/Microsoft.FeatureManagement.Telemetry.ApplicationInsights/ApplicationInsightsTelemetryPublisher.cs
```diff
@@ -66,5 +71,26 @@ private void ValidateEvent(EvaluationEvent evaluationEvent)
                 throw new ArgumentNullException(nameof(evaluationEvent.FeatureDefinition));
             }
         }
+
+        private static string ToString(AssignmentReason reason)
+        {
+            const string None = "None";
```
zhiyuanliang-ms, line 77: Since we have implemented a ToString method, shall we add more detailed variant assignment reason? For example, AssignmentReason.None can be mapped to "No variant can be allocated".

src/Microsoft.FeatureManagement.Telemetry.ApplicationInsights/ApplicationInsightsTelemetryPublisher.cs
zhenlan: Then it doesn't sound right. We should send the telemetry. This is a case people will want to know that the feature flag is misconfigured.

jimmyca15: > This is a case people will want to know that the feature flag is misconfigured.

I'm not sure what you intend to be discovered here. What would you expect "AssignmentReason" to be on a basic on/off feature flag? We already don't send variant in this case. I'm not sure what meaning "AssignmentReason" would have if no variant has been assigned.

src/Microsoft.FeatureManagement.Telemetry.ApplicationInsights/ApplicationInsightsTelemetryPublisher.cs
jimmyca15: These are metric dimensions, I don't think we should be verbose.

src/Microsoft.FeatureManagement.Telemetry.ApplicationInsights/ApplicationInsightsTelemetryPublisher.cs
zhenlan: Sync'ed up with Jimmy offline and Jimmy help clarify the scenario. I was confused by the comment for `None`: _No variant is assigned_. `None` actually indicates the variant evaluation never occurs so the assignment is irrelevant. The assignment reason more like "NA" in this case. We could update the comment to make it clearer.

src/Microsoft.FeatureManagement/FeatureManager.cs
zhiyuanliang-ms: Update: 
`None` is only used when there are no variants. As soon as a variant is defined then some sort of assignment happens. Even if we return a null variant it's because we got to a point where we decided that the user should get `DefaultWhenDisabled`, or `DefaultWhenEnabled`, etc...

src/Microsoft.FeatureManagement.Telemetry.ApplicationInsights/ApplicationInsightsTelemetryPublisher.cs
```diff
@@ -30,17 +30,22 @@ public ValueTask PublishEvent(EvaluationEvent evaluationEvent, CancellationToken
...
             };
 
             if (evaluationEvent.Variant != null)
             {
                 properties["Variant"] = evaluationEvent.Variant.Name;
             }
 
+            if (evaluationEvent.VariantAssignmentReason != VariantAssignmentReason.None)
+            {
+                properties["VariantAssignmentReason"] = ToString(evaluationEvent.VariantAssignmentReason);
```
jimmyca15, line 46: Always send Variant if assignment reason is not `None`. Even if variant is null.

src/Microsoft.FeatureManagement.Telemetry.ApplicationInsights/ApplicationInsightsTelemetryPublisher.cs
```diff
@@ -66,5 +71,24 @@ private void ValidateEvent(EvaluationEvent evaluationEvent)
                 throw new ArgumentNullException(nameof(evaluationEvent.FeatureDefinition));
             }
         }
+
+        private static string ToString(VariantAssignmentReason reason)
+        {
```
jimmyca15, line 76: Can you add `Debug.Assert(reason != VariantAssignmentReason.None);` at the beginning of the method. It helps add clarity to the reader that this is accounted for and is not missing.

src/Microsoft.FeatureManagement/FeatureManager.cs
```diff
@@ -241,77 +251,99 @@ public ValueTask<Variant> GetVariantAsync(string feature, TargetingContext conte
...
                         }
                     }
                 }
-            }
 
-            foreach (ISessionManager sessionManager in _sessionManagers)
-            {
-                await sessionManager.SetAsync(feature, isFeatureEnabled).ConfigureAwait(false);
-            }
+                Debug.Assert(evaluationEvent.Variant != null ? evaluationEvent.VariantAssignmentReason != VariantAssignmentReason.None : true);
```
jimmyca15, line 333: I've never seen an assert like this before. Seems a bit odd.

src/Microsoft.FeatureManagement/FeatureManager.cs
```diff
@@ -241,77 +251,99 @@ public ValueTask<Variant> GetVariantAsync(string feature, TargetingContext conte
...
+                //
+                // Determine IsEnabled
+                evaluationEvent.Enabled = await IsEnabledAsync(evaluationEvent.FeatureDefinition, context, useContext, cancellationToken).ConfigureAwait(false);
+
+                //
+                // Determine Variant
+                VariantDefinition variantDefinition = null;
 
-                if (featureDefinition.Variants != null && featureDefinition.Variants.Any() && featureDefinition.Allocation != null)
+                if (evaluationEvent.FeatureDefinition.Allocation != null && (evaluationEvent.FeatureDefinition.Variants?.Any() ?? false))
```
jimmyca15, line 276: If there is at least one variant then I believe we should run assignment. In the case that there is at least one variant, but allocation is null then I expect the variant returned should be `null` but the assignment reason should be `DefaultWhenDisabled`. Allocation being null in this case is the same as Allocation being not null, but `Allocation.DefaultWhenDisabled` being null.

jimmyca15: Can invert the logic below to check for 

```
if (!evaluationEvent.Enabled ||
    evaluationEvent.Allocation == null)```

File: src/Microsoft.FeatureManagement/FeatureManager.cs (hunk 20 above)
Comment (zhiyuanliang-ms) [reply src/Microsoft.FeatureManagement/FeatureManager.cs:333]: Removed.

File: src/Microsoft.FeatureManagement/FeatureManager.cs (hunk 21 above)
Comment (zhiyuanliang-ms) [reply src/Microsoft.FeatureManagement/FeatureManager.cs:276]: To make it clear, I put the assignment process here:
image

> In the case that there is at least one variant, but allocation is null then I expect the variant returned should be null but the assignment reason should be DefaultWhenDisabled. Allocation being null in this case is the same as Allocation being not null, but Allocation.DefaultWhenDisabled being null.

I am a bit confused, why should the reason be `DefaultWhenDisabled`. I think the reason should be `DefaultWhenEnabled` or `DefaultWhenDisabled` depending on whether the feature flag is on or off.

Comment (jimmyca15) [reply src/Microsoft.FeatureManagement/FeatureManager.cs]: Discussed offline, I should have said it is either DefualtWhenDisabled or DefaultWhenEnabled

File: src/Microsoft.FeatureManagement/FeatureManager.cs [hunk 22]
```diff
@@ -241,77 +251,110 @@ public ValueTask<Variant> GetVariantAsync(string feature, TargetingContext conte
...
+                        {
+                            if (evaluationEvent.FeatureDefinition.Allocation.DefaultWhenEnabled != null)
+                            {
+                                variantDefinition = evaluationEvent.FeatureDefinition
+                                    .Variants
+                                    .FirstOrDefault(variant =>
+                                        variant.Name == evaluationEvent.FeatureDefinition.Allocation.DefaultWhenEnabled);
+                            }
+
+                            evaluationEvent.VariantAssignmentReason = VariantAssignmentReason.DefaultWhenEnabled;
```
jimmyca15, line 325: Not necessarily, it could be possible that there was a user match, but the variant that was pointed to doesn't exist. We shouldn't overwrite VariantAssignmentReason in that case.

jimmyca15: We can check assignment reason rather than `variantDefinition != null`

jimmyca15 (approved):

src/Microsoft.FeatureManagement/Telemetry/VariantAssignmentReason.cs
```diff
@@ -0,0 +1,41 @@
...
+        /// </summary>
+        None,
+
+        /// <summary>
+        /// Variant is assigned by default when the feature flag is disabled.
+        /// </summary>
+        DefaultWhenDisabled,
+
+        /// <summary>
+        /// Variant is assigned by default after processing the user/group/percentile allocation, when the feature flag is enabled.
```
zhenlan, line 22: Following the pattern of your other descriptions, will this be clearer?

The default variant is assigned because of no applicable user/group/percentile allocation when a feature flag is enabled.

src/Microsoft.FeatureManagement/Telemetry/VariantAssignmentReason.cs
```diff
@@ -0,0 +1,41 @@
...
+        /// </summary>
+        DefaultWhenEnabled,
+
+        /// <summary>
+        /// Variant is assigned because of the user allocation.
```
zhenlan, line 27: Do we want to be more complete?

The variant is assigned because of the user allocation when a feature flag is enabled.

zhenlan (approved):

Issue #291, opened 2023-10-25
Change the feature flag name reference to const string


zhiyuanliang-ms:
Currently, our testcases are using `Enum.GetName(typeof(Features), Features.TargetingTestFeature)` to reference a feature flag.

zhiyuanliang-ms:
FeatureFlagDemo should also be updated.

juniwang:
Could you explain what's wrong with name reference? 

Also, I'm aslo wondering why not just use `nameof`? For example, `nameof(Features.TargetingTestFeature)`, which sounds pretty simple and straightforward.

zhiyuanliang-ms:
@juniwang The context is here: #268  #269 
Jimmy and I discussed this offline. We prefered to use const string. There was also a PR to update the README: https://github.com/microsoft/FeatureManagement-Dotnet/pull/275

PR #294, opened 2023-10-27
use const string to referece feature name

zhiyuanliang-ms:
Issue #291


jimmyca15 (approved):

tests/Tests.FeatureManagement/FeatureFlags.cs
```diff
@@ -0,0 +1,19 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+//
+namespace Tests.FeatureManagement
+{
+    static class FeatureFlags
```
jimmyca15, line 6: Why change Features to FeatureFlags?

zhiyuanliang-ms: I can change it back to Features. In README, we use "FeatureFlag".

Issue #258, opened 2023-09-11
Feature Manager DI Scoping


samgibsonmoj:
I have a net7 Blazor Server application and have integrated the Feature Management package into the application, registering various filters.

One filter, the 'TenantFilter' is responsible for enabling/disabling various areas of the system based on current users tenancy. This filter makes use of my 'ICurrentUserService', which is a scoped service that gives contextual information based on the user (e.g. tenant id, user id, user name).

I've noted that the Feature Manager (and other dependencies) are registered as a Singleton... There's no way around this scoping, as these types aren't made available outside of the library.

I'm unable to get any of the users contextual information via the HttpContextAccessor, [as the HttpContext isn't guaranteed to be available via this interface](https://learn.microsoft.com/en-us/aspnet/core/blazor/security/?view=aspnetcore-3.1&tabs=visual-studio#avoid-ihttpcontextaccessorhttpcontext-in-razor-components).

While I understand there may be no enhancements planned for Blazor Server, I was curious on your take on this, and any potential solutions/workarounds. Currently, I've had to fork this repository, and register the services as Scoped.

Thanks.

rossgrambo:
Hello @samgibsonmoj 

It sounds like a workaround is:
> The recommended approach for passing request state to the Blazor app is through root component parameters during the app's initial rendering.

I'm curious if you're able to pass this state when you construct your HTTPContextAccessor (create a new Accessor that accepts it as a parameter) so it could instead use that state instead of the normal HttpContext. Might not make sense with the lifecycle of Blazor apps, as I haven't used Blazor yet. 

Otherwise I think your fork is the correct solution for now. I'll check with others about adding an injection option for Blazor.

samgibsonmoj:
Thanks @rossgrambo, I appreciate the fast response. 

> I'm curious if you're able to pass this state when you construct your HTTPContextAccessor (create a new Accessor that accepts it as a parameter) so it could instead use that state instead of the normal HttpContext. Might not make sense with the lifecycle of Blazor apps, as I haven't used Blazor yet.

I'm not 100% sure this would work for what I'm looking to achieve. From what I can see, passing these values via the root component would make them available to other child components, but not within the Filter itself. Please let me know if I have misunderstood.

For some further clarification, the filter looks like this:

```
// TenantFilter.cs

[FilterAlias("App.Tenant")]
public class TenantFilter : IFeatureFilter
{
	private readonly ICurrentUserService _userService;

[...]

Issue #126, opened 2021-05-26
Other di containers


da1rren:
Hey Guys,

We're attempting to use this in a .net framework application that uses autofac.  While attempting to register the various dependencies:

* ConfigurationFeatureDefinitionProvider
* FeatureManager
* EmptySessionManager
* FeatureManagerSnapshot

 I've discovered that they are all sealed, is there any guidance or is it even possible to use this without resorting to reflection hacks?

jeffcollaboro:
... or any other number of containers, Simple Injector etc, These classes are the public interface into feature flags they shouldn't be internal, we need to instantiate them to use this.

danielmackay:
We're having this problem also.  It would be great to see some guidance of how to use the `FeatureManager` with other IoC containers.

zhiyuanliang-ms:
Update:
The PR #302 of exposing FeatureManager to public has been merged into main branch. Now the FeatureManagement should be compatible with other DI framework.

zhiyuanliang-ms:
Fixed in release 3.1.0

zhiyuanliang-ms:
Fixed in release 3.1.0

zhiyuanliang-ms:
I created a repo to demonstrate how to use Microsoft.FeatureManagement in third-party Di containers: https://github.com/zhiyuanliang-ms/FeatureManagement-Dotnet-DependencyInjection
:)

Issue #15, opened 2019-09-29
Unable to inject Scoped/Transient services into FeatureFilter


artyom-p:
In.NET Core 3.0 DI scope validation is enabled by default.
Application throws an exception that it is not possible to resolve service that is Scoped/Transient .

`Cannot consume scoped service 'Common.Services.IRequestDataProvider' from singleton 'Microsoft.FeatureManagement.IFeatureFilter'`

jimmyca15:
Is this different from existing behavior? I wouldn't expect a scoped service to be able to be used in the feature filters which are singletons. This is why things like `IHttpContextAccessor` services exist which give a way to access the scoped HttpContext.

jimmyca15:
Closing since there has been no response. Please re-open if the my response is inadequate.

cbrianball:
I ran into this issue yesterday. I managed to code around it, but it was less than ideal.

I ran into two primary issues:

1. I inject the current user via an interface I've created.
2. I was unable to inject my EF Core database context (and even if I could, there could be potential issues as EF Core DB Context is not meant to be run on multiple threads).

For the first issue, a web project could inject `IHttpContextAccessor`, but my code is in a shared library that is used by both a web project and non-web projects. I worked around this by creating a type that functions similarly to `HttpContextAccessor`, but for my interface that describes the current user instead.

I could have passed the user object into the feature manager method as a context, but then the calling code would need to know that the configured feature filters would need to act on the current user, and I don't believe that's a good idea.

For the second issue, I ended up injecting `IServiceProvider` into my feature filters that need to access the database. This is a hack, in my opinion, and I hope to find another solution in the near future.

Both of these could have been avoided if the registered feature filters were not singletons (either scoped or transient would have sufficed).

The code-base seems pretty straight forward, but I don't want to make any bad assumptions. If you can give me some general direction (or at least tell me what you are trying to avoid), I can work on a PR to allow feature filters to not necessarily be singleton.

szalapski:
I am hitting this too.  I wish it could depend on a scoped ISessionManager.

zhiyuanliang-ms:
> I ran into this issue yesterday. I managed to code around it, but it was less than ideal.
> 
> I ran into two primary issues:
> 
> 1. I inject the current user via an interface I've created.
> 2. I was unable to inject my EF Core database context (and even if I could, there could be potential issues as EF Core DB Context is not meant to be run on multiple threads).
> 
[...]

PR #298, opened 2023-11-01
Expose feature manager to public for preview branch

zhiyuanliang-ms:
Our users called out the requirement to expose the FeatureManager class to public. (Issue #258 #126 #15 )
Exposing FeatureManager to public can allow uses to have the flexibility to register the FeatureManager as scoped or transient. 
In .NET, singleton service cannot consume scoped service.
Besides, it allows otherthird-party DI containers like Autofac Unity, etc, to use FeatureManagement.


jimmyca15:
Can you provide some details in the subscription of the goal. Perhaps link to an issue.

src/Microsoft.FeatureManagement/ServiceCollectionExtensions.cs
```diff
@@ -27,31 +28,32 @@ public static IFeatureManagementBuilder AddFeatureManagement(this IServiceCollec
...
             services.TryAddSingleton<IFeatureDefinitionProvider, ConfigurationFeatureDefinitionProvider>();
 
             services.AddSingleton(sp => new FeatureManager(
                 sp.GetRequiredService<IFeatureDefinitionProvider>(),
-                sp.GetRequiredService<IEnumerable<IFeatureFilterMetadata>>(),
-                sp.GetRequiredService<IEnumerable<ISessionManager>>(),
-                sp.GetRequiredService<ILoggerFactory>(),
-                sp.GetRequiredService<IOptions<FeatureManagementOptions>>(),
-                sp.GetRequiredService<IOptions<TargetingEvaluationOptions>>())
+                sp.GetRequiredService<IOptions<FeatureManagementOptions>>()?.Value)
```
jimmyca15, line 39:
```suggestion
                sp.GetRequiredService<IOptions<FeatureManagementOptions>>().Value)
```

Can't be null since you're using GetRequiredService

src/Microsoft.FeatureManagement/ServiceCollectionExtensions.cs
```diff
@@ -27,31 +28,32 @@ public static IFeatureManagementBuilder AddFeatureManagement(this IServiceCollec
...
             {
-                Configuration = sp.GetService<IConfiguration>(),
-                TargetingContextAccessor = sp.GetService<ITargetingContextAccessor>(),
+                FeatureFilters = sp.GetService<IEnumerable<IFeatureFilterMetadata>>(),
```
jimmyca15, line 41: Can you use GetRequiredService? This should ensure we always get an IEnumerable.

src/Microsoft.FeatureManagement/ServiceCollectionExtensions.cs
```diff
@@ -27,31 +28,32 @@ public static IFeatureManagementBuilder AddFeatureManagement(this IServiceCollec
...
+                SessionManagers = sp.GetService<IEnumerable<ISessionManager>>(),
```
jimmyca15, line 42: `SessionManagers = sp.GetService<IEnumerable<ISessionManager>>(),`

FeatureManager throws if SessionManagers is set to null, but I thought `sp.GetService<IEnumerable<ISessionManager>>()` would return null if no-one registered any session managers?

src/Microsoft.FeatureManagement/FeatureManager.cs
```diff
@@ -44,46 +43,127 @@ private class ConfigurationCacheItem
...
+
+        /// <summary>
+        /// The logger for the feature manager
+        /// </summary>
+        public ILogger Logger { get; init; }
+
+        /// <summary>
+        /// The collection of telemetry publishers
+        /// </summary>
         public IEnumerable<ITelemetryPublisher> TelemetryPublishers { get; init; }
```
jimmyca15, line 103: How come this one is nullable?

src/Microsoft.FeatureManagement/FeatureManagementBuilderExtensions.cs
```diff
@@ -42,6 +42,7 @@ private static IFeatureManagementBuilder AddTelemetryPublisher(this IFeatureMana
             return builder;
         }
 
+        /// <summary>
```
jimmyca15, line 45: Needs a separate PR to fix this.

src/Microsoft.FeatureManagement/FeatureManager.cs
```diff
@@ -22,20 +22,19 @@ namespace Microsoft.FeatureManagement
     /// <summary>
     /// Used to evaluate whether a feature is enabled or disabled.
```
jimmyca15, line 23: It also can be used to retrieve variants now.

src/Microsoft.FeatureManagement/FeatureManager.cs
```diff
@@ -44,46 +43,127 @@ private class ConfigurationCacheItem
             public object Settings { get; set; }
         }
 
+        /// <summary>
+        /// Create a feature manager
+        /// </summary>
+        /// <param name="featureDefinitionProvider">The provider of feature flag definitions.</param>
+        /// <param name="options">Options controlling the behavior of the feature manager.</param>
```
jimmyca15, line 50: A throws statement would be valuable to detail that IFeatureDefinitionProvider and the options are required.

src/Microsoft.FeatureManagement/FeatureManager.cs
```diff
@@ -44,46 +43,127 @@ private class ConfigurationCacheItem
...
             _contextualFeatureFilterCache = new ConcurrentDictionary<string, ContextualFeatureFilterEvaluator>();
-            _options = options?.Value ?? throw new ArgumentNullException(nameof(options));
-            _parametersCache = new MemoryCache(new MemoryCacheOptions());
+            _featureFilters = Enumerable.Empty<IFeatureFilterMetadata>();
+            _sessionManagers = Enumerable.Empty<ISessionManager>();
+            _assignerOptions = new TargetingEvaluationOptions();
+        }
+
+        /// <summary>
+        /// The collection of feature filter metadata
```
jimmyca15, line 65: It's a sentence. Should end with a period.

src/Microsoft.FeatureManagement/FeatureManager.cs
```diff
@@ -44,46 +43,127 @@ private class ConfigurationCacheItem
...
+        /// <summary>
+        /// Checks whether a given feature is enabled.
+        /// </summary>
+        /// <param name="feature">The name of the feature to check.</param>
+        /// <param name="appContext">A context providing information that can be used to evaluate whether a feature should be on or off.</param>
+        /// <param name="cancellationToken">The cancellation token to cancel the operation.</param>
+        /// <returns>True if the feature is enabled, otherwise false.</returns>
         public ValueTask<bool> IsEnabledAsync<TContext>(string feature, TContext appContext, CancellationToken cancellationToken)
         {
             return IsEnabledWithVariantsAsync(feature, appContext, useAppContext: true, cancellationToken);
```
jimmyca15, line 169: The public methods should be before all the private methods. Looks like a private method got snuck in with the publics.

src/Microsoft.FeatureManagement/ServiceCollectionExtensions.cs
zhiyuanliang-ms: No, it wouldn't return null if there is not any registered ISessionManager. An empty IEnumerable will be returned.
Anyway, I agree with that we should change it into GetRequiredService<>

src/Microsoft.FeatureManagement/FeatureManager.cs
zhiyuanliang-ms: https://github.com/microsoft/FeatureManagement-Dotnet/blob/preview/src/Microsoft.FeatureManagement/FeatureManager.cs#L674

We will log warning if the feature flag asks to publish telemetry but there is no registered telemetry publisher.
``` C#
if (TelemetryPublishers == null || !TelemetryPublishers.Any())
{
    _logger.LogWarning("The feature declaration enabled telemetry but no telemetry publisher was registered.");
 }
```

jimmyca15: We can simplify to just `!TelemetryPublishers.Any` like our other `IEnumerable` usage.

src/Microsoft.FeatureManagement/FeatureManager.cs
```diff
@@ -44,51 +43,208 @@ private class ConfigurationCacheItem
...
+            _featureFilters = Enumerable.Empty<IFeatureFilterMetadata>();
+            _sessionManagers = Enumerable.Empty<ISessionManager>();
+            _telemetryPublishers = Enumerable.Empty<ITelemetryPublisher>();
+            _assignerOptions = new TargetingEvaluationOptions();
         }
 
-        public IEnumerable<ITelemetryPublisher> TelemetryPublishers { get; init; }
+        /// <summary>
+        /// The collection of feature filter metadata.
+        /// <exception cref="ArgumentNullException">Thrown if it is set to null.</exception>
```
jimmyca15, line 69: Isn't exception supposed to be outside of the summary element? Can you check the generated docs?

src/Microsoft.FeatureManagement/FeatureManager.cs
```diff
@@ -44,51 +43,208 @@ private class ConfigurationCacheItem
...
+        public IMemoryCache Cache { get; init; }
+
+        /// <summary>
+        /// The logger for the feature manager.
+        /// </summary>
+        public ILogger Logger { get; init; }
+
+        /// <summary>
+        /// The collection of telemetry publishers.
+        /// /// <exception cref="ArgumentNullException">Thrown if it is set to null.</exception>
```
jimmyca15, line 107: Has some extra '/'. Also, similar comment as above.

src/Microsoft.FeatureManagement/ServiceCollectionExtensions.cs
```diff
@@ -27,31 +29,33 @@ public static IFeatureManagementBuilder AddFeatureManagement(this IServiceCollec
...
-                sp.GetRequiredService<ILoggerFactory>(),
-                sp.GetRequiredService<IOptions<FeatureManagementOptions>>(),
-                sp.GetRequiredService<IOptions<TargetingEvaluationOptions>>())
+                sp.GetRequiredService<IOptions<FeatureManagementOptions>>().Value)
             {
-                Configuration = sp.GetService<IConfiguration>(),
-                TargetingContextAccessor = sp.GetService<ITargetingContextAccessor>(),
+                FeatureFilters = sp.GetRequiredService<IEnumerable<IFeatureFilterMetadata>>(),
+                SessionManagers = sp.GetRequiredService<IEnumerable<ISessionManager>>(),
                 TelemetryPublishers = sp.GetService<IOptions<FeatureManagementOptions>>()?.Value.TelemetryPublisherFactories?
```
jimmyca15, line 44:
```suggestion
                TelemetryPublishers = sp.GetRequiredService<IOptions<FeatureManagementOptions>>().Value?.TelemetryPublisherFactories?
```

src/Microsoft.FeatureManagement/ServiceCollectionExtensions.cs
```diff
@@ -27,31 +29,33 @@ public static IFeatureManagementBuilder AddFeatureManagement(this IServiceCollec
...
                     .Select(factory => factory(sp))
-                    .ToList()
+                    .ToList() ??
+                    Enumerable.Empty<ITelemetryPublisher>(),
+                Cache = sp.GetService<IMemoryCache>(),
```
jimmyca15, line 48: Can get it as a RequiredService given we add it previously.

src/Microsoft.FeatureManagement/ServiceCollectionExtensions.cs
```diff
@@ -27,31 +29,33 @@ public static IFeatureManagementBuilder AddFeatureManagement(this IServiceCollec
...
+                Logger = sp.GetService<ILoggerFactory>().CreateLogger<FeatureManager>(),
```
jimmyca15, line 49:
```suggestion
                Logger = sp.GetRequiredService<ILoggerFactory>().CreateLogger<FeatureManager>(),
```

We ensure logging is there before this. So we can require it.

jimmyca15 (approved):

PR #302, opened 2023-11-07
Expose feature manager to public 

zhiyuanliang-ms:
Our users called out the requirement to expose the FeatureManager class to public. (Issue https://github.com/microsoft/FeatureManagement-Dotnet/issues/258 https://github.com/microsoft/FeatureManagement-Dotnet/issues/126 https://github.com/microsoft/FeatureManagement-Dotnet/issues/15 )
Exposing FeatureManager to public can allow uses to have the flexibility to register the FeatureManager as scoped or transient.
In .NET, singleton service cannot consume scoped service.
Besides, it allows otherthird-party DI containers like Autofac Unity, etc, to use FeatureManagement.

Targeting on the main branch


src/Microsoft.FeatureManagement/ServiceCollectionExtensions.cs
```diff
@@ -16,23 +19,35 @@ namespace Microsoft.FeatureManagement
...
+                FeatureFilters = sp.GetRequiredService<IEnumerable<IFeatureFilterMetadata>>(),
+                SessionManagers = sp.GetRequiredService<IEnumerable<ISessionManager>>(),
+                Cache = sp.GetRequiredService<IMemoryCache>(),
+                Logger = sp.GetRequiredService<ILoggerFactory>().CreateLogger<FeatureManager>()
+            });
+
+            services.TryAddSingleton<IFeatureManager>(sp => sp.GetRequiredService<FeatureManager>());
 
-            services.AddSingleton<ISessionManager, EmptySessionManager>();
+            services.AddScoped<FeatureManagerSnapshot>();
```
jimmyca15, line 48: What's the reasoning to go away from usage like `services.AddScoped<IFeatureManagerSnapshot, FeatureManagerSnapshot>();`

jimmyca15: I would expect `services.TryAddScoped<IFeatureManagerSnapshot, FeatureManagerSnapshot>()` should accomplish our needs.

jimmyca15 (approved):

src/Microsoft.FeatureManagement/ServiceCollectionExtensions.cs
jimmyca15: Same with registration of IFeatureManager.

zhiyuanliang-ms: In the preview branch, we have IVariantFeatureManager and IFeatureManager sharing the same implementation FeatureManager, we use this pattern to register them. In the main branch, I agree to change it back to our previous way to register IFeatureManager.

zhiyuanliang-ms: There is no overload of TryAddSingleton<TService, TImplementation>(ImplementationFactory). We need to use factory method to register FeatureManager implementation now. I think the current way to register IFeatureManager is better.

PR #303, opened 2023-11-07
Implement AddScopedFeatureManagement() method

zhiyuanliang-ms:


src/Microsoft.FeatureManagement/FeatureManagementBuilderExtensions.cs
```diff
@@ -18,7 +20,16 @@ public static class FeatureManagementBuilderExtensions
         /// <returns>A <see cref="IFeatureManagementBuilder"/> that can be used to customize feature management functionality.</returns>
         public static IFeatureManagementBuilder WithTargeting<T>(this IFeatureManagementBuilder builder) where T : ITargetingContextAccessor
         {
-            builder.Services.TryAddSingleton(typeof(ITargetingContextAccessor), typeof(T));
+            //
+            // Register the targeting context accessor with the same lifetime as the feature manager
+            if (!builder.Services.Any(descriptor => descriptor.ServiceType == typeof(FeatureManager) && descriptor.Lifetime == ServiceLifetime.Scoped))
```
jimmyca15, line 25: This condition seems backwards. Addsingleton if feature manager is registered as scoped?

jimmyca15: Oh it's negated.

jimmyca15:
```suggestion
            if (builder.Services.Any(descriptor => descriptor.ServiceType == typeof(FeatureManager) && descriptor.Lifetime == ServiceLifetime.Scoped))
            {
                // register as scoped
            }
```

I feel it will be clearer to remove the negation and instead override to scoped if you do detect a scoped feature manager.

src/Microsoft.FeatureManagement/FeatureManagementBuilder.cs
```diff
@@ -37,15 +37,33 @@ public FeatureManagementBuilder(IServiceCollection services)
 
             if (!Services.Any(descriptor => descriptor.ServiceType == serviceType && descriptor.ImplementationType == implementationType))
             {
-                Services.AddSingleton(serviceType, implementationType);
+                //
+                // Register the feature filter with the same lifetime as the feature manager
+                if (!Services.Any(descriptor => descriptor.ServiceType == typeof(FeatureManager) && descriptor.Lifetime == ServiceLifetime.Scoped))
```
jimmyca15, line 42: ideally, check IFeatureManager, not FeatureManager.

src/Microsoft.FeatureManagement/ServiceCollectionExtensions.cs
```diff
@@ -64,5 +88,75 @@ public static IFeatureManagementBuilder AddFeatureManagement(this IServiceCollec
...
+
+            services.AddLogging();
+
+            services.AddMemoryCache();
+
+            //
+            // Add required services
+            services.TryAddSingleton<IFeatureDefinitionProvider, ConfigurationFeatureDefinitionProvider>();
+
+            services.AddScoped(sp => new FeatureManager(
```
jimmyca15, line 115: I don't see a reason to add `FeatureManager` into DI. Previously we only registered IFeatureManager.

zhiyuanliang-ms: In preview branch, we add FeatureManager and use it for IFeatureManager and IVariantFeatureManager.
https://github.com/microsoft/FeatureManagement-Dotnet/blob/preview/src/Microsoft.FeatureManagement/ServiceCollectionExtensions.cs#L34
I follow that way.

src/Microsoft.FeatureManagement/FeatureManager.cs
```diff
@@ -16,54 +15,107 @@ namespace Microsoft.FeatureManagement
     /// <summary>
     /// Used to evaluate whether a feature is enabled or disabled.
     /// </summary>
-    class FeatureManager : IFeatureManager, IDisposable
+    public sealed class FeatureManager : IFeatureManager
```
jimmyca15, line 18: Does this PR need to be merged/reviewed after the PR to make feature manager public is in?

zhiyuanliang-ms: Yes.

jimmyca15 (approved):

PR #305, opened 2023-11-07
Fix solution issues

jimmyca15:
* missing project inclusion in solution file
* Project declared nullable support but wasn't written with it in mind
* pack.ps1 script didn't reference newest project.


zhiyuanliang-ms:
Should we remove the nullable declaration in RazorPages example project?

examples/TargetingConsoleApp/TargetingConsoleApp.csproj
```diff
@@ -4,7 +4,6 @@
     <OutputType>Exe</OutputType>
     <TargetFramework>net6.0</TargetFramework>
     <ImplicitUsings>enable</ImplicitUsings>
-    <Nullable>enable</Nullable>
```
zhiyuanliang-ms, line 7: Is there any reason that we keep the <Nullable>enables</Nullable> in the RazorPages example?

jimmyca15: Because that project was made with nullable in mind. It uses it correctly. The other projects don't use nullable in the code, and it seems it was added into csproj after the fact causing compiler warnings.

jimmyca15:
@zhiyuanliang-ms any further concerns on this PR?

zhiyuanliang-ms (approved):

zhiyuanliang-ms:
@jimmyca15 I have no further concerns about this PR.

PR #312, opened 2023-11-14
Add option for no feature management section found

### LINKED ISSUE #308: builder.Services.AddFeatureManagement() with configuration section does not work since v3.0.0
I know that there was a breaking change in v3.0.0 created by https://github.com/microsoft/FeatureManagement-Dotnet/pull/261, but I feel like that should not affect the use case of targeting the configuration section to use in the appsettings.json file.

For example, with  this configuration file:

```json
{
  "Logging": {
    "LogLevel": {
      "Default": "Information",
      "Microsoft.AspNetCore": "Warning"
    }
  },
  "FeatureFlags": {
    "PushFeatureEnabled": true,
    "PutFeatureEnabled": true
  }
}
```

and this program.cs line:

```csharp
builder.Services.AddFeatureManagement(builder.Configuration.GetSection("FeatureFlags"));

var app = builder.Build();
```

the feature manager does not load our feature flags like it did in 2.6.0. If we're not allowed to set our own configuration section, it should no be a possible parameter or at least raise a deprecated warning

zhiyuanliang-ms:
Resolve #308


src/Microsoft.FeatureManagement/ServiceCollectionExtensions.cs
```diff
@@ -60,7 +61,15 @@ public static IFeatureManagementBuilder AddFeatureManagement(this IServiceCollec
                 throw new ArgumentNullException(nameof(configuration));
             }
 
-            services.AddSingleton<IFeatureDefinitionProvider>(sp => new ConfigurationFeatureDefinitionProvider(configuration, sp.GetRequiredService<ILoggerFactory>()));
+            services.Configure<FeatureManagementOptions>(options =>
+            {
+                options.RequireFeatureManagementSection = false;
+            });
+
+            services.AddSingleton<IFeatureDefinitionProvider>(sp => new ConfigurationFeatureDefinitionProvider(
```
jimmyca15, line 69: I don't see a need for FeatureManagementOptions to be updated. In fact it's dangerous. If IFeatureDefinitionProvider that is created with the inline configuration is not the one that ends up being used (multiple addfeaturemanagmeent calls), then the options won't match the added service.

I think it needs to be

```
            services.TryAddSingleton<IFeatureDefinitionProvider>(sp =>
                    configuration, 
                    sp.GetRequiredService<ILoggerFactory>())
                {
                    UseTopLevelConfiguration = true
                });
```

src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
```diff
@@ -42,6 +43,19 @@ public ConfigurationFeatureDefinitionProvider(IConfiguration configuration, ILog
...
+        /// <summary>
+        /// The option that controls the behavior when "FeatureManagement" section in the configuration is missing.
+        /// </summary>
+        public bool UseTopLevelConfiguration
+        {
+            get => _useTopLevelConfiguration;
+
+            init
+            {
+                _useTopLevelConfiguration = value;
```
jimmyca15, line 55: This is unnecessary since no validation is performed. We should be able to just have


```
public bool UseTopLevelConfiguration { get; init; }
```

src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
```diff
@@ -217,6 +231,13 @@ private IEnumerable<IConfigurationSection> GetFeatureDefinitionSections()
                 return featureManagementConfigurationSection.GetChildren();
             }
 
+            //
+            // There is no "FeatureManagement" section in the configuration
```
jimmyca15, line 235: we shouldn't be performing the check for the featuremanagement section if this flag is set.

src/Microsoft.FeatureManagement/FeatureManagementOptions.cs
```diff
@@ -22,5 +22,12 @@ public class FeatureManagementOptions
         /// The default value is true.
         /// </summary>
         public bool IgnoreMissingFeatures { get; set; } = true;
+
+        /// <summary>
+        /// Controls the behavior of feature definition provider when "FeatureManagement" section in the configuration is missing.
+        /// If missing configuration section is not ignored, the feature definition provider will fall back to the root of the provided configuration.
+        /// The default value is false.
+        /// </summary>
+        public bool RequireFeatureManagementSection { get; set; } = true;
```
jimmyca15, line 31: Where is this used?

zhiyuanliang-ms: Sorry, forget to remove. Removed.

jimmyca15 (approved):

PR #319, opened 2023-11-22
Support Microsoft Feature Flag schema feature flag schema v1.1.0 by default

zhiyuanliang-ms:
We want to augment the FeatureManagement-Dotnet to support the [Microsoft Feature Flag schema v1.1.0](https://github.com/Azure/AppConfiguration/blob/main/docs/FeatureManagement/FeatureFlag.v1.1.0.schema.json)

``` Json
{
    "FeatureManagement": {
        "FeatureFlags": [
            // Microsoft Feature Flag objects
        ]
    }
}
```
The Microsoft Feature Flag definitions will be listed in the **_"FeatureFlags"_** array.

It is not allowed to use different schemas at the same time. If we can find the “FeatureFlags” array in the “FeatureManagement” section, we will only parse the Microsoft feature flag definitions listed in the "FeatureFlags" array. The old schema (if any) will be ignored.

### DIFF
--- a/src/Microsoft.FeatureManagement.AspNetCore/FeatureGateAttribute.cs
+++ b/src/Microsoft.FeatureManagement.AspNetCore/FeatureGateAttribute.cs
@@ -133,7 +133,7 @@ namespace Microsoft.FeatureManagement.Mvc
             IFeatureManagerSnapshot fm = context.HttpContext.RequestServices.GetRequiredService<IFeatureManagerSnapshot>();
 
             //
-            // FeatureFlagEnabled state is determined by either 'any' or 'all' features being enabled.
+            // Enabled state is determined by either 'any' or 'all' features being enabled.
             bool enabled = RequirementType == RequirementType.All ?
                              await Features.All(async feature => await fm.IsEnabledAsync(feature).ConfigureAwait(false)) :
                              await Features.Any(async feature => await fm.IsEnabledAsync(feature).ConfigureAwait(false));
--- a/src/Microsoft.FeatureManagement.AspNetCore/FeatureGateAttribute.cs
+++ b/src/Microsoft.FeatureManagement.AspNetCore/FeatureGateAttribute.cs
@@ -105,7 +105,7 @@ namespace Microsoft.FeatureManagement.Mvc
             IFeatureManagerSnapshot fm = context.HttpContext.RequestServices.GetRequiredService<IFeatureManagerSnapshot>();
 
             //
-            // Enabled state is determined by either 'any' or 'all' features being enabled.
+            // FeatureFlagEnabled state is determined by either 'any' or 'all' features being enabled.
             bool enabled = RequirementType == RequirementType.All ?
                              await Features.All(async feature => await fm.IsEnabledAsync(feature).ConfigureAwait(false)) :
                              await Features.Any(async feature => await fm.IsEnabledAsync(feature).ConfigureAwait(false));
@@ -133,7 +133,7 @@ namespace Microsoft.FeatureManagement.Mvc
             IFeatureManagerSnapshot fm = context.HttpContext.RequestServices.GetRequiredService<IFeatureManagerSnapshot>();
 
             //
-            // Enabled state is determined by either 'any' or 'all' features being enabled.
+            // FeatureFlagEnabled state is determined by either 'any' or 'all' features being enabled.
             bool enabled = RequirementType == RequirementType.All ?
                              await Features.All(async feature => await fm.IsEnabledAsync(feature).ConfigureAwait(false)) :
                              await Features.Any(async feature => await fm.IsEnabledAsync(feature).ConfigureAwait(false));
--- a/src/Microsoft.FeatureManagement.AspNetCore/FeatureGateAttribute.cs
+++ b/src/Microsoft.FeatureManagement.AspNetCore/FeatureGateAttribute.cs
@@ -105,7 +105,7 @@ namespace Microsoft.FeatureManagement.Mvc
             IFeatureManagerSnapshot fm = context.HttpContext.RequestServices.GetRequiredService<IFeatureManagerSnapshot>();
 
             //
-            // FeatureFlagEnabled state is determined by either 'any' or 'all' features being enabled.
+            // Enabled state is determined by either 'any' or 'all' features being enabled.
             bool enabled = RequirementType == RequirementType.All ?
                              await Features.All(async feature => await fm.IsEnabledAsync(feature).ConfigureAwait(false)) :
                              await Features.Any(async feature => await fm.IsEnabledAsync(feature).ConfigureAwait(false));
--- a/src/Microsoft.FeatureManagement/AzureAppConfigurationFeatureFlagConfigurationField.cs
+++ b/src/Microsoft.FeatureManagement/AzureAppConfigurationFeatureFlagConfigurationField.cs
@@ -1,0 +1,19 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+//
+
+namespace Microsoft.FeatureManagement
+{
+    //
+    // Azure App Configuration feature flag schema: https://github.com/Azure/AppConfiguration/blob/main/docs/FeatureManagement/FeatureFlag.v1.1.0.schema.json
+    internal static class AzureAppConfigurationFeatureFlagConfigurationFields
+    {
+        public const string FeatureFlagsSectionName = "FeatureFlags";
+
+        public const string Id = "id";
+        public const string Enabled = "enabled";
+        public const string Conditions = "conditions";
+        public const string ClientFilters = "client_filters";
+        public const string RequirementType = "requirement_type";
+    }
+}
--- a/src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
+++ b/src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
@@ -154,6 +154,16 @@ namespace Microsoft.FeatureManagement
 
         private FeatureDefinition ReadFeatureDefinition(IConfigurationSection configurationSection)
         {
+            if (_featureFlagArraySchemaEnabled)
+            {
+                return ParseFeatureDefinitionInFeatureFlagsArray(configurationSection);
+            }
+
+            return ParseFeatureDefinition(configurationSection);
+        }
+
+        private FeatureDefinition ParseFeatureDefinition(IConfigurationSection configurationSection)
+        {
             /*
               
             We support
@@ -185,6 +195,77 @@ namespace Microsoft.FeatureManagement
                 enabledFor: [ "myFeatureFilter1", "myFeatureFilter2" ],
             },
 
+            */
+
+            string featureName = configurationSection[ConfigurationFields.IdKeyword];
+
+            List<FeatureFilterConfiguration> enabledFor = new List<FeatureFilterConfiguration>();
+
+            RequirementType requirementType = RequirementType.Any;
+
+            string val = configurationSection.Value; // configuration[$"{featureName}"];
+
+            if (string.IsNullOrEmpty(val))
+            {
+                val = configurationSection[ConfigurationFields.FeatureFiltersSectionName];
+            }
+
+            if (!string.IsNullOrEmpty(val) && bool.TryParse(val, out bool result) && result)
+            {
+                //
+                //myAlwaysEnabledFeature: true
+                // OR
+                //myAlwaysEnabledFeature: {
+                //  enabledFor: true
+                //}
+                enabledFor.Add(new FeatureFilterConfiguration
+                {
+                    Name = "AlwaysOn"
+                });
+            }
+            else
+            {
+                string rawRequirementType = configurationSection[ConfigurationFields.RequirementType];
+
+                //
+                // If requirement type is specified, parse it and set the requirementType variable
+                if (!string.IsNullOrEmpty(rawRequirementType) && !Enum.TryParse(rawRequirementType, ignoreCase: true, out requirementType))
+                {
+                    throw new FeatureManagementException(
+                        FeatureManagementError.InvalidConfigurationSetting,
+                        $"Invalid requirement type '{rawRequirementType}' for feature '{featureName}'.");
+                }
+
+                IEnumerable<IConfigurationSection> filterSections = configurationSection.GetSection(ConfigurationFields.FeatureFiltersSectionName).GetChildren();
+
+                foreach (IConfigurationSection section in filterSections)
+                {
+                    //
+                    // Arrays in json such as "myKey": [ "some", "values" ]
+                    // Are accessed through the configuration system by using the array index as the property name, e.g. "myKey": { "0": "some", "1": "values" }
+                    if (int.TryParse(section.Key, out int _) && !string.IsNullOrEmpty(section[ConfigurationFields.NameKeyword]))
+                    {
+                        enabledFor.Add(new FeatureFilterConfiguration()
+                        {
+                            Name = section[ConfigurationFields.NameKeyword],
+                            Parameters = new ConfigurationWrapper(section.GetSection(ConfigurationFields.FeatureFilterConfigurationParameters))
+                        });
+                    }
+                }
+            }
+
+            return new FeatureDefinition()
+            {
+                Name = featureName,
+                EnabledFor = enabledFor,
+                RequirementType = requirementType
+            };
+        }
+
+        private FeatureDefinition ParseFeatureDefinitionInFeatureFlagsArray(IConfigurationSection configurationSection)
+        {
+            /*
+            
             If App Config server side schema is enabled, we support
 
             FeatureFlags: [
@@ -211,111 +292,15 @@ namespace Microsoft.FeatureManagement
 
             */
 
-            string featureName;
+            string featureName = configurationSection[ConfigurationFields.IdKeyword];
+
+            List<FeatureFilterConfiguration> enabledFor = new List<FeatureFilterConfiguration>();
 
             RequirementType requirementType = RequirementType.Any;
 
-            var enabledFor = new List<FeatureFilterConfiguration>();
+            IConfigurationSection conditions = configurationSection.GetSection(ConfigurationFields.ConditionsKeyword);
 
-            string rawRequirementType = string.Empty;
-
-            if (_featureFlagArraySchemaEnabled)
-            {
-                featureName = configurationSection[ConfigurationFields.IdKeyword];
-
-                IConfigurationSection conditions = configurationSection.GetSection(ConfigurationFields.ConditionsKeyword);
-
-                rawRequirementType = conditions[ConfigurationFields.LowercaseRequirementType];
-
-                string rawEnabled = configurationSection[ConfigurationFields.EnabledKeyword];
-
-                bool enabled = false;
-
-                if (!string.IsNullOrEmpty(rawEnabled) && !bool.TryParse(rawEnabled, out enabled))
-                {
-                    throw new FeatureManagementException(
-                            FeatureManagementError.InvalidConfigurationSetting,
-                            $"Invalid enabled '{rawEnabled}' for feature '{featureName}'.");
-                }
-
-                if (enabled)
-                {
-                    var test = conditions.GetSection(ConfigurationFields.ClientFiltersSectionName);
-
-                    IEnumerable<IConfigurationSection> filterSections = conditions.GetSection(ConfigurationFields.ClientFiltersSectionName).GetChildren();
-
-                    if (filterSections.Any())
-                    {
-                        foreach (IConfigurationSection section in filterSections)
-                        {
-                            //
-                            // Arrays in json such as "myKey": [ "some", "values" ]
-                            // Are accessed through the configuration system by using the array index as the property name, e.g. "myKey": { "0": "some", "1": "values" }
-                            if (int.TryParse(section.Key, out int _) && !string.IsNullOrEmpty(section[ConfigurationFields.LowercaseFeatureManagementSectionName]))
-                            {
-                                enabledFor.Add(new FeatureFilterConfiguration()
-                                {
-                                    Name = section[ConfigurationFields.LowercaseFeatureManagementSectionName],
-                                    Parameters = new ConfigurationWrapper(section.GetSection(ConfigurationFields.LowercaseFeatureManagementSectionParameters))
-                                });
-                            }
-                        }
-                    }
-                    else
-                    {
-                        enabledFor.Add(new FeatureFilterConfiguration
-                        {
-                            Name = "AlwaysOn"
-                        });
-                    }
-                }
-            }
-            else
-            {
-                featureName = configurationSection.Key;
-
-                string val = configurationSection.Value; // configuration[$"{featureName}"];
-
-                if (string.IsNullOrEmpty(val))
-                {
-                    val = configurationSection[ConfigurationFields.FeatureFiltersSectionName];
-                }
-
-                if (!string.IsNullOrEmpty(val) && bool.TryParse(val, out bool result) && result)
-                {
-                    //
-                    //myAlwaysEnabledFeature: true
-                    // OR
-                    //myAlwaysEnabledFeature: {
-                    //  enabledFor: true
-                    //}
-                    enabledFor.Add(new FeatureFilterConfiguration
-                    {
-                        Name = "AlwaysOn"
-                    });
-                }
-                else
-                {
-                    rawRequirementType = configurationSection[ConfigurationFields.RequirementType];
-
-                    IEnumerable<IConfigurationSection> filterSections = configurationSection.GetSection(ConfigurationFields.FeatureFiltersSectionName).GetChildren();
-
-                    foreach (IConfigurationSection section in filterSections)
-                    {
-                        //
-                        // Arrays in json such as "myKey": [ "some", "values" ]
-                        // Are accessed through the configuration system by using the array index as the property name, e.g. "myKey": { "0": "some", "1": "values" }
-                        if (int.TryParse(section.Key, out int _) && !string.IsNullOrEmpty(section[ConfigurationFields.NameKeyword]))
-                        {
-                            enabledFor.Add(new FeatureFilterConfiguration()
-                            {
-                                Name = section[ConfigurationFields.NameKeyword],
-                                Parameters = new ConfigurationWrapper(section.GetSection(ConfigurationFields.FeatureFilterConfigurationParameters))
-                            });
-                        }
-                    }
-                }
-            }
+            string rawRequirementType = conditions[ConfigurationFields.LowercaseRequirementType];
 
             //
             // If requirement type is specified, parse it and set the requirementType variable
@@ -324,6 +309,47 @@ namespace Microsoft.FeatureManagement
                 throw new FeatureManagementException(
                     FeatureManagementError.InvalidConfigurationSetting,
                     $"Invalid requirement type '{rawRequirementType}' for feature '{featureName}'.");
+            }
+
+            string rawEnabled = configurationSection[ConfigurationFields.EnabledKeyword];
+
+            bool enabled = false;
+
+            if (!string.IsNullOrEmpty(rawEnabled) && !bool.TryParse(rawEnabled, out enabled))
+            {
+                throw new FeatureManagementException(
+                        FeatureManagementError.InvalidConfigurationSetting,
+                        $"Invalid enabled '{rawEnabled}' for feature '{featureName}'.");
+            }
+
+            if (enabled)
+            {
+                IEnumerable<IConfigurationSection> filterSections = conditions.GetSection(ConfigurationFields.ClientFiltersSectionName).GetChildren();
+
+                if (filterSections.Any())
+                {
+                    foreach (IConfigurationSection section in filterSections)
+                    {
+                        //
+                        // Arrays in json such as "myKey": [ "some", "values" ]
+                        // Are accessed through the configuration system by using the array index as the property name, e.g. "myKey": { "0": "some", "1": "values" }
+                        if (int.TryParse(section.Key, out int _) && !string.IsNullOrEmpty(section[ConfigurationFields.LowercaseFeatureManagementSectionName]))
+                        {
+                            enabledFor.Add(new FeatureFilterConfiguration()
+                            {
+                                Name = section[ConfigurationFields.LowercaseFeatureManagementSectionName],
+                                Parameters = new ConfigurationWrapper(section.GetSection(ConfigurationFields.LowercaseFeatureManagementSectionParameters))
+                            });
+                        }
+                    }
+                }
+                else
+                {
+                    enabledFor.Add(new FeatureFilterConfiguration
+                    {
+                        Name = "AlwaysOn"
+                    });
+                }
             }
 
             return new FeatureDefinition()
--- a/src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
+++ b/src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
@@ -96,7 +96,7 @@ namespace Microsoft.FeatureManagement
         /// </summary>
         /// <returns>An enumerator which provides asynchronous iteration over feature definitions.</returns>
         //
-        // The async featureName word is necessary for creating IAsyncEnumerable.
+        // The async key word is necessary for creating IAsyncEnumerable.
         // The need to disable this warning occurs when implementaing async stream synchronously. 
 #pragma warning disable CS1998 // Async method lacks 'await' operators and will run synchronously
         public async IAsyncEnumerable<FeatureDefinition> GetAllFeatureDefinitionsAsync()
--- a/src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
+++ b/src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
@@ -22,13 +22,11 @@ namespace Microsoft.FeatureManagement
         // IFeatureDefinitionProviderCacheable interface is only used to mark this provider as cacheable. This allows our test suite's
         // provider to be marked for caching as well.
 
-        private const string FeatureFiltersSectionName = "EnabledFor";
-        private const string RequirementTypeKeyword = "RequirementType";
-        private const string FeatureManagementSectionName = "FeatureManagement";
         private readonly IConfiguration _configuration;
         private readonly ConcurrentDictionary<string, FeatureDefinition> _definitions;
         private IDisposable _changeSubscription;
         private int _stale = 0;
+        private bool _serverSideSchemaEnabled;
 
         /// <summary>
         /// Creates a configuration feature definition provider.
@@ -98,7 +96,7 @@ namespace Microsoft.FeatureManagement
         /// </summary>
         /// <returns>An enumerator which provides asynchronous iteration over feature definitions.</returns>
         //
-        // The async key word is necessary for creating IAsyncEnumerable.
+        // The async featureName word is necessary for creating IAsyncEnumerable.
         // The need to disable this warning occurs when implementaing async stream synchronously. 
 #pragma warning disable CS1998 // Async method lacks 'await' operators and will run synchronously
         public async IAsyncEnumerable<FeatureDefinition> GetAllFeatureDefinitionsAsync()
@@ -113,16 +111,38 @@ namespace Microsoft.FeatureManagement
             // Iterate over all features registered in the system at initial invocation time
             foreach (IConfigurationSection featureSection in GetFeatureDefinitionSections())
             {
+                string featureName = featureSection.Key;
+
+                if (_serverSideSchemaEnabled)
+                {
+                    featureName = featureSection[ConfigurationFields.ServerSideIdKeyword];
+
+                    if (string.IsNullOrEmpty(featureName))
+                    {
+                        continue;
+                    }
+                }
+
                 //
                 // Underlying IConfigurationSection data is dynamic so latest feature definitions are returned
-                yield return  _definitions.GetOrAdd(featureSection.Key, (_) => ReadFeatureDefinition(featureSection));
+                yield return  _definitions.GetOrAdd(featureName, (_) => ReadFeatureDefinition(featureSection));
             }
         }
 
         private FeatureDefinition ReadFeatureDefinition(string featureName)
         {
             IConfigurationSection configuration = GetFeatureDefinitionSections()
-                                                    .FirstOrDefault(section => section.Key.Equals(featureName, StringComparison.OrdinalIgnoreCase));
+                .FirstOrDefault(section => 
+                {
+                    if (_serverSideSchemaEnabled)
+                    {
+                        string id = section[ConfigurationFields.ServerSideIdKeyword];
+
+                        return !string.IsNullOrEmpty(id) && id.Equals(featureName, StringComparison.OrdinalIgnoreCase);
+                    }
+
+                    return section.Key.Equals(featureName, StringComparison.OrdinalIgnoreCase);
+                });
 
             if (configuration == null)
             {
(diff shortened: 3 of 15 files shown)


src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
```diff
@@ -22,13 +22,11 @@ public sealed class ConfigurationFeatureDefinitionProvider : IFeatureDefinitionP
...
         // provider to be marked for caching as well.
 
-        private const string FeatureFiltersSectionName = "EnabledFor";
-        private const string RequirementTypeKeyword = "RequirementType";
-        private const string FeatureManagementSectionName = "FeatureManagement";
         private readonly IConfiguration _configuration;
         private readonly ConcurrentDictionary<string, FeatureDefinition> _definitions;
         private IDisposable _changeSubscription;
         private int _stale = 0;
+        private bool _serverSideSchemaEnabled;
```
RichardChen820, line 29: Is `_isFeatureFlagArraySchema` better? Anyway，we'd avoid using the “serverSide” word, since it is not meaningful for FeatureManagement library

src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
```diff
@@ -165,92 +185,199 @@ We support
...
-                return featureManagementConfigurationSection.GetChildren();
+                //
+                // There is no "FeatureManagement" section in the configuration
+                if (RootConfigurationFallbackEnabled)
+                {
+                    _serverSideSchemaEnabled = featureFlagsConfigurationSection.Exists();
+
+                    if (_serverSideSchemaEnabled)
+                    {
+                        return featureFlagsConfigurationSection.GetChildren();
```
RichardChen820, line 351: Once `FeatureFlags:[]` section exists, we no longer treat other root configurations as feature flags, is this intended?

src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
```diff
@@ -165,92 +185,199 @@ We support
...
 
         private IEnumerable<IConfigurationSection> GetFeatureDefinitionSections()
         {
-            //
-            // Look for feature definitions under the "FeatureManagement" section
-            IConfigurationSection featureManagementConfigurationSection = _configuration.GetSection(FeatureManagementSectionName);
+            IConfigurationSection featureManagementConfigurationSection = _configuration.GetSection(ConfigurationFields.FeatureManagementSectionName);
 
-            if (featureManagementConfigurationSection.Exists())
+            IConfigurationSection featureFlagsConfigurationSection = featureManagementConfigurationSection.GetSection(ConfigurationFields.FeatureFlagsSectionName);
```
RichardChen820, line 339: If `featureFlagsConfigurationSection` does not exists, this line won't throw an exception?

src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
zhiyuanliang-ms: Yes.

src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
zhiyuanliang-ms: No. The GetSection method will never return null. If there is no matching sub-section, an empty IConfigurationSection will be returned.

src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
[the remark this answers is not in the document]
jimmyca15: There is no such thing as server side schema. This library exists completely independently from Azure App Configuration. All notion of server side should be eliminated.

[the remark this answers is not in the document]
RichardChen820: The current PR title makes sense to me, this PR is exactly for that purpose, support App Config Schema. But I think we should not name anything in the code with "app configuration"

src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
```diff
@@ -193,64 +225,178 @@ We support
...
-            if (featureManagementConfigurationSection.Exists())
+            string rawEnabled = configurationSection[ConfigurationFields.EnabledKeyword];
+
+            bool enabled = false;
+
+            if (!string.IsNullOrEmpty(rawEnabled) && !bool.TryParse(rawEnabled, out enabled))
             {
-                return featureManagementConfigurationSection.GetChildren();
+                throw new FeatureManagementException(
+                        FeatureManagementError.InvalidConfigurationSetting,
```
RichardChen820, line 321: nit: wrong indentation

src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
```diff
@@ -193,64 +225,178 @@ We support
...
+                        $"Invalid enabled '{rawEnabled}' for feature '{featureName}'.");
```
RichardChen820, line 322: Invalid value '{rawEnabled}' for `enabled` field of feature '{featureName}'.

src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
```diff
@@ -193,64 +225,178 @@ We support
...
+                IEnumerable<IConfigurationSection> filterSections = conditions.GetSection(ConfigurationFields.ClientFiltersSectionName).GetChildren();
+
+                if (filterSections.Any())
+                {
+                    foreach (IConfigurationSection section in filterSections)
+                    {
+                        //
+                        // Arrays in json such as "myKey": [ "some", "values" ]
+                        // Are accessed through the configuration system by using the array index as the property name, e.g. "myKey": { "0": "some", "1": "values" }
+                        if (int.TryParse(section.Key, out int _) && !string.IsNullOrEmpty(section[ConfigurationFields.LowercaseFeatureManagementSectionName]))
```
RichardChen820, line 336: Should we throw an exception if the child of `client_filters` is not an array?

src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
```diff
@@ -193,64 +225,178 @@ We support
...
+                        {
+                            enabledFor.Add(new FeatureFilterConfiguration()
+                            {
+                                Name = section[ConfigurationFields.LowercaseFeatureManagementSectionName],
```
RichardChen820, line 340: Should we handle the situation that the name and parameters are missing?

src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
RichardChen820: I'm trying to imagine such a user case below in Kubernetes (since App Config Kubernetes Provider will be the only place that provides the server side schema).

A customer has a configuration file that includes FeatureManagement section (old schema) in his app. He uses Kubernetes Provider to get some feature flags from Azure App Configuration and expect the feature flags from App Config can overwrite the feature flags in his local file.

I'm not sure whether this is a valid user case we should support. But apparently, we can't handle this case with current code we have.

src/Microsoft.FeatureManagement/ConfigurationFields.cs
```diff
@@ -0,0 +1,30 @@
...
+        public const string FeatureFiltersSectionName = "EnabledFor";
+        public const string FeatureFilterConfigurationParameters = "Parameters";
+
+        // Other keywords
+        public const string NameKeyword = "Name";
+        public const string FeatureManagementSectionName = "FeatureManagement";
+        public const string FeatureFlagsSectionName = "FeatureFlags";
+
+        // Azure App Configuration feature flag schema keywords
+        public const string IdKeyword = "id";
```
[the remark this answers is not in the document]
zhiyuanliang-ms: Id, Enabled, Conditions and ClientFilters are renamed.
 
I want to remain the naming of NameKeyword because it is consistent with the preview branch. https://github.com/microsoft/FeatureManagement-Dotnet/blob/preview/src/Microsoft.FeatureManagement/ConfigurationFields.cs#L38

src/Microsoft.FeatureManagement/ConfigurationFields.cs
```diff
@@ -0,0 +1,30 @@
...
+        public const string EnabledKeyword = "enabled";
+        public const string ConditionsKeyword = "conditions";
+        public const string ClientFiltersSectionName = "client_filters";
+        public const string LowercaseFeatureManagementSectionName = "name";
```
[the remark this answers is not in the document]
zhiyuanliang-ms: Sorry, this is a mistake. My intended naming way is to add "Lowercase-" prefix to the previous variable name. (e.g. LowercaseNameKeyword, LowercaseFeatureFilterConfigurationParameters)

src/Microsoft.FeatureManagement/ConfigurationFields.cs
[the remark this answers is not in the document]
RichardChen820: It's for internal use only, why we can't update it?

src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
```diff
@@ -193,64 +225,178 @@ We support
...
+
+                    if (_AzureAppConfigurationFeatureFlagSchemaEnabled)
+                    {
+                        return featureFlagsConfigurationSection.GetChildren();
+                    }
+                    
+                    return _configuration.GetChildren();
+                }
+
+                Logger?.LogDebug($"No configuration section named '{ConfigurationFields.FeatureManagementSectionName}' was found.");
```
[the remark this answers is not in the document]
zhiyuanliang-ms: If code enters this line, the RootConfigurationFallBackEnabled should be false. In this case, "FeatureManagement" section is required no matter whether there is "FeatureFlags" section.

src/Microsoft.FeatureManagement/ConfigurationFields.cs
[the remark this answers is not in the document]
zhiyuanliang-ms: Changed to "Name" and "LowercaseName".

src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
```diff
@@ -113,16 +111,38 @@ public async IAsyncEnumerable<FeatureDefinition> GetAllFeatureDefinitionsAsync()
...
             IConfigurationSection configuration = GetFeatureDefinitionSections()
-                                                    .FirstOrDefault(section => section.Key.Equals(featureName, StringComparison.OrdinalIgnoreCase));
+                .FirstOrDefault(section => 
+                {
+                    if (_AzureAppConfigurationFeatureFlagSchemaEnabled)
+                    {
+                        string id = section[ConfigurationFields.IdKeyword];
+
+                        return !string.IsNullOrEmpty(id) && id.Equals(featureName, StringComparison.OrdinalIgnoreCase);
+                    }
```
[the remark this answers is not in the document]
zhiyuanliang-ms: I propose to add a private method:

``` C#
private string GetFeatureFlagSectionName(IConfigurationSection section)
{
    if (_azureAppConfigurationFeatureFlagSchemaEnabled)
    {
        return section[ConfigurationFields.FeatureFlagId];
    }
    else
    {
        return section.Key;
    }
}
```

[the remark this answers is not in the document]
RichardChen820: I'm okay with it

src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
zhiyuanliang-ms: I think we should just ignore it since we cannot parse it.

src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
zhiyuanliang-ms: If the name is missing, it will not enter the if branch. The parameters is ok to be missing.

src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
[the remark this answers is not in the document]
zhenlan: "App Configuration" is the branding. It's like the FM library adopts a schema that is created by App Configuration.

src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
zhenlan: Schema validation should be based on the schema
https://github.com/Azure/AppConfiguration/blob/main/docs/FeatureManagement/FeatureFlag.v1.1.0.schema.json

src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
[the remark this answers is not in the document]
RichardChen820: So you are suggesting naming correlated variables with "AppConfiguration" in the code?

jimmyca15:
PR description should reference the schema we are aiming to support: 

https://github.com/Azure/AppConfiguration/blob/main/docs/FeatureManagement/FeatureFlag.v1.1.0.schema.json

src/Microsoft.FeatureManagement.AspNetCore/FeatureGateAttribute.cs
```diff
@@ -133,7 +133,7 @@ public async Task OnPageHandlerExecutionAsync(PageHandlerExecutingContext contex
             IFeatureManagerSnapshot fm = context.HttpContext.RequestServices.GetRequiredService<IFeatureManagerSnapshot>();
 
             //
-            // Enabled state is determined by either 'any' or 'all' features being enabled.
+            // FeatureFlagEnabled state is determined by either 'any' or 'all' features being enabled.
```
jimmyca15, line 136: I don't understand this change

src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
```diff
@@ -193,64 +210,190 @@ We support
...
+                Name = featureName,
+                EnabledFor = enabledFor,
+                RequirementType = requirementType
+            };
+        }
 
-            return Enumerable.Empty<IConfigurationSection>();
+        private IEnumerable<IConfigurationSection> GetFeatureDefinitionSections()
+        {
+            IConfigurationSection featureManagementConfigurationSection = _configuration.GetSection(ConfigurationFields.FeatureManagementSectionName);
```
jimmyca15, line 350: A bit different than what I was expecting.

I was expecting:
First decide if we need to use "FeatureManagement" section or root configuration.

Then decide if we use original schema, or App Configuration schema based off whether the previously identified section has a feature flags subsection.

src/Microsoft.FeatureManagement/ConfigurationFields.cs
```diff
@@ -0,0 +1,30 @@
...
+        public const string NameKeyword = "Name";
+        public const string FeatureManagementSectionName = "FeatureManagement";
+        public const string FeatureFlagsSectionName = "FeatureFlags";
+
+        // Azure App Configuration feature flag schema keywords
+        public const string FeatureFlagId = "id";
+        public const string FeatureFlagEnabled = "enabled";
+        public const string FeatureFlagConditions = "conditions";
+        public const string FeatureFlagClientFilters = "client_filters";
+        public const string LowercaseNameKeyword = "name";
```
jimmyca15, line 26: Why do we need lower case name? .NET configuration system is case insensitive.

src/Microsoft.FeatureManagement.AspNetCore/FeatureGateAttribute.cs
zhiyuanliang-ms: Sorry, my bad. This is a mistake caused by rename. I click the "Include comments" and "Include strings" by mistake.

src/Microsoft.FeatureManagement/ConfigurationFields.cs
zhiyuanliang-ms: （°o°；）Really? I never noticed that. Thanks for pointing it out!

jimmyca15: Got pulled away before I could write another comment. It would be better to have App Configuration fields in a separate file.

src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
```diff
@@ -193,64 +210,182 @@ We support
...
+                {
+                    Logger?.LogDebug($"No configuration section named '{ConfigurationFields.FeatureManagementSectionName}' was found.");
+
+                    return Enumerable.Empty<IConfigurationSection>();
+                }
+
+                featureManagementConfigurationSection = _configuration as IConfigurationSection;
+            }
+
+            IConfigurationSection featureFlagsConfigurationSection = featureManagementConfigurationSection.GetSection(AzureAppConfigurationFeatureFlagConfigurationFields.FeatureFlagsSectionName);
```
jimmyca15, line 364: This is assigned to with an `as` implying this can be null. Looks like a possible null-ref.

src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
```diff
@@ -193,64 +210,182 @@ We support
...
+
             //
-            // There is no "FeatureManagement" section in the configuration
-            if (RootConfigurationFallbackEnabled)
+            // "FeatureFlag" section should be an array if Azure App Configuration feature flag schema is enabled
+            _azureAppConfigurationFeatureFlagSchemaEnabled = featureFlagsConfigurationSection.Exists() &&
```
jimmyca15, line 368: I left another comment about this. This is not designed well to be assigned to multiple times. This class is designed as thread safe. Meaning this value can change while another thread is somewhere having previously evaluated it.

Either a bigger redesign is needed, or we need to assign to this only once.

src/Microsoft.FeatureManagement/AzureAppConfigurationFeatureFlagConfigurationField.cs
```diff
@@ -0,0 +1,19 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+//
+
+namespace Microsoft.FeatureManagement
+{
+    //
+    // Azure App Configuration feature flag schema: https://github.com/Azure/AppConfiguration/blob/main/docs/FeatureManagement/FeatureFlag.v1.1.0.schema.json
+    internal static class AzureAppConfigurationFeatureFlagConfigurationFields
```
jimmyca15, line 9:
```suggestion
    internal static class AzureAppConfigurationFeatureFlagFields
```

src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
```diff
@@ -193,64 +210,182 @@ We support
...
 
                 //
                 // If requirement type is specified, parse it and set the requirementType variable
                 if (!string.IsNullOrEmpty(rawRequirementType) && !Enum.TryParse(rawRequirementType, ignoreCase: true, out requirementType))
                 {
                     throw new FeatureManagementException(
-                        FeatureManagementError.InvalidConfigurationSetting, 
-                        $"Invalid requirement type '{rawRequirementType}' for feature '{configurationSection.Key}'.");
+                        FeatureManagementError.InvalidConfigurationSetting,
+                        $"Invalid value '{rawRequirementType}' for '{ConfigurationFields.RequirementType}' field of feature '{featureName}'.");
```
jimmyca15, line 221:
```suggestion
                        $"Invalid value '{rawRequirementType}' for field '{ConfigurationFields.RequirementType}' of feature '{featureName}'.");
```

src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
```diff
@@ -193,64 +210,182 @@ We support
...
+            
+            If Azure App Configuration feature flag schema is enabled, we support
+
+            FeatureFlags: [
+              {
+                id: "myFeature",
+                enabled: true,
+                conditions: {
+                  client_filters: ["myFeatureFilter1", "myFeatureFilter2"],
+                  requirement_type: "all",
```
jimmyca15, line 262:
```suggestion
                  requirement_type: "All",
```

jimmyca15: Although we support both, the schema defines it as "All" so we should use the correct casing in our comment.

src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
```diff
@@ -193,64 +210,182 @@ We support
...
+            {
+                throw new FeatureManagementException(
+                    FeatureManagementError.InvalidConfigurationSetting,
+                    $"Invalid value '{rawRequirementType}' for '{AzureAppConfigurationFeatureFlagConfigurationFields.RequirementType}' field of feature '{featureName}'.");
+            }
 
-            if (featureManagementConfigurationSection.Exists())
+            string rawEnabled = configurationSection[AzureAppConfigurationFeatureFlagConfigurationFields.Enabled];
+
+            bool enabled = false;
```
jimmyca15, line 301: Is there a need to initialize this variable?

src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
```diff
@@ -193,64 +210,182 @@ We support
...
+                    {
+                        //
+                        // Arrays in json such as "myKey": [ "some", "values" ]
+                        // Are accessed through the configuration system by using the array index as the property name, e.g. "myKey": { "0": "some", "1": "values" }
+                        if (int.TryParse(section.Key, out int _) && !string.IsNullOrEmpty(section[ConfigurationFields.NameKeyword]))
+                        {
+                            enabledFor.Add(new FeatureFilterConfiguration()
+                            {
+                                Name = section[ConfigurationFields.NameKeyword],
+                                Parameters = new ConfigurationWrapper(section.GetSection(ConfigurationFields.FeatureFilterConfigurationParameters))
```
jimmyca15, line 326: Shouldn't this code path be using AzureAppConfigurationFeatureFlagFields?

zhiyuanliang-ms: The "Name" and "Parameter" fields are used by both schemas to define feature filter. So I reused them. I can duplicate them in AzureAppConfigurationFeatureFlagFields.

src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
zhiyuanliang-ms: According to https://github.com/Azure/AppConfiguration/blob/main/docs/FeatureManagement/FeatureFlag.v1.1.0.schema.json
"enabled" field is required. Do you expect that if the "enabled" field cannot be found, an exception will be thrown? Or if the feature management cannot find the "enabled" field, we treat it as false.

src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
zhiyuanliang-ms: Sorry, I didn't understand. `_configuration` cannot be null, otherwise an exception will be thrown in the constructor. 
IConfiguration can always be casted to IConfigurationSection. I think there could not be a null-ref.

src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
```diff
@@ -22,13 +22,12 @@ public sealed class ConfigurationFeatureDefinitionProvider : IFeatureDefinitionP
...
         // provider to be marked for caching as well.
 
-        private const string FeatureFiltersSectionName = "EnabledFor";
-        private const string RequirementTypeKeyword = "RequirementType";
-        private const string FeatureManagementSectionName = "FeatureManagement";
         private readonly IConfiguration _configuration;
         private readonly ConcurrentDictionary<string, FeatureDefinition> _definitions;
         private IDisposable _changeSubscription;
         private int _stale = 0;
+        private int _schemaNotSet = 1;
```
RichardChen820, line 29: Is `_schemaSet` more straightforward?

src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
```diff
@@ -41,7 +40,10 @@ public ConfigurationFeatureDefinitionProvider(IConfiguration configuration)
 
             _changeSubscription = ChangeToken.OnChange(
                 () => _configuration.GetReloadToken(),
-                () => _stale = 1);
+                () => {
+                    _stale = 1;
+                    _schemaNotSet = 1;
```
RichardChen820, line 45: Any reason for not ensuring the thread safety here?

src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
jimmyca15: It's more appropriate to use `AzureAppConfigurationFeatureFlagFields`.

src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
```diff
@@ -193,64 +214,185 @@ We support
...
+                {
+                    Logger?.LogDebug($"No configuration section named '{ConfigurationFields.FeatureManagementSectionName}' was found.");
+
+                    return Enumerable.Empty<IConfigurationSection>();
+                }
+
+                featureManagementConfigurationSection = _configuration as IConfigurationSection;
+            }
+
+            IConfigurationSection featureFlagsConfigurationSection = featureManagementConfigurationSection.GetSection(AzureAppConfigurationFeatureFlagFields.FeatureFlagsSectionName);
```
jimmyca15, line 368: `as` implies null as a possibility. Meaning this is a possible null-ref.

src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
jimmyca15: The latest update does not solve the problem.

Once a public method is invoked and takes an action based off the schema, it should stick with that schema the entire time. Otherwise unexpected things will happen.

Originally I was thinking to just limit the schema to not be able to change during the runtime of the application. Is this acceptable?

If not, then we can have to separate internal feature definition providers that we invoke in the ConfigurationFeatureDefinitionProvider based off of the detected schema.
* DefaultConfigurationFeatureDefinitionProvider
* AacConfigurationFeatureDefinitionProvider

jimmyca15:
```
if (_aacProvider.HasFlags)
{
    _aacProvider.Xyz();
}
else
{
    _defaultProvider.Xyz();
}
```

---

_configuration.OnChange(() =>
{
    _aacProvider.MarkStale();

    _defaultProvider.MarkStale();
});

src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
jimmyca15: I mean, is there a need to write

```
bool enabled = false;
```

instead of

```
bool enabled;
```

Seems the second one should be fine.

jimmyca15:
In your description, those are not schemas. those may follow a particular schema. The file you linked to is a schema.

jimmyca15:
In the PR description, I'd like to highlight the "FeatureFlags" array's appearance. I don't particularly see a need to write any appsettings.json snippet other than

```
{
    "FeatureManagement": {
        "FeatureFlags": [
            // App Configuration Feature Flag objects
        ]
    }
}
```

src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
zhiyuanliang-ms: If we just declare 
``` C#
bool enabled
```
This line will give build error: use of unassigned local variable 'enabled'
``` C#
if (enabled)
```

src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
zhiyuanliang-ms: Are you expecting a null check after the type casting?
The _configuration is IConfiguration. The null check for _configuration happens in the constructor of ConfigurationFeatureDefinitionProvider. So _configuration cannot be null here and it can always be casted to IConfigurationSection.

src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
zhiyuanliang-ms: > Originally I was thinking to just limit the schema to not be able to change during the runtime of the application. Is this acceptable?
It depends on whether we consider the scenario that configuration is dynamically changed to a different schema during the runtime, is a valid user case.

Richard mentioned this potential user case:
https://github.com/microsoft/FeatureManagement-Dotnet/pull/319#discussion_r1408827330

I am thinking whether we should allow both schema to be recognized at the same time.

src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
jimmyca15: `as` is not a type cast.

this is a type cast

`(IConfigurationSection)_configuration`

Yes, if `as` is used, then a null check is needed. If we know that the object IS of a certain type then we shouldn't use `as` and we should type cast instead.

zhiyuanliang-ms: I think we can use type cast `IConfigurationSection featureFlagsConfigurationSection = (IConfigurationSection)_configuration` directly.

jimmyca15: Okay, type case it is.

src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
```diff
@@ -193,64 +212,224 @@ We support
...
+                {
+                    Logger?.LogDebug($"Configuration is empty.");
+
+                    return Enumerable.Empty<IConfigurationSection>();
+                }
+
+                featureManagementConfigurationSection = _configuration;
+            }
+
+            lock(_lock)
```
jimmyca15, line 386: Let's talk about locks.

zhiyuanliang-ms:
The latest change aims to achieve the following goals:
1. _azureAppConfigurationFeatureFlagSchemaEnabled can only be set for once during the runtime
2. If the initial configuration is empty which means we have no context to determine whether schema to use, it will not set _azureAppConfigurationFeatureFlagSchemaEnabled until the configuration changes.

src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
```diff
@@ -113,16 +116,57 @@ public async IAsyncEnumerable<FeatureDefinition> GetAllFeatureDefinitionsAsync()
...
                 //
                 // Underlying IConfigurationSection data is dynamic so latest feature definitions are returned
-                yield return  _definitions.GetOrAdd(featureSection.Key, (_) => ReadFeatureDefinition(featureSection));
+                yield return  _definitions.GetOrAdd(featureName, (_) => ReadFeatureDefinition(featureSection));
+            }
+        }
+
+        private void EnsureInit()
+        {
+            if (!_configuration.GetChildren().Any())
```
jimmyca15, line 134: EnsureInit should not return without being initialized.

src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
```diff
@@ -113,16 +116,57 @@ public async IAsyncEnumerable<FeatureDefinition> GetAllFeatureDefinitionsAsync()
...
+            {
+                Logger?.LogDebug($"Configuration is empty.");
+
+                return;
```
jimmyca15, line 138: EnsureInit should not return without being initialized.

src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
```diff
@@ -113,16 +116,57 @@ public async IAsyncEnumerable<FeatureDefinition> GetAllFeatureDefinitionsAsync()
...
+                .Any(section => string.Equals(section.Key, ConfigurationFields.FeatureManagementSectionName, StringComparison.OrdinalIgnoreCase));
+
+            if (!hasFeatureManagementSection && !RootConfigurationFallbackEnabled)
+            {
+                Logger?.LogDebug($"No configuration section named '{ConfigurationFields.FeatureManagementSectionName}' was found.");
+
+                return;
+            }
+
+            if (_schemaSet == 0)
```
jimmyca15, line 151: `_initialized`

src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
```diff
@@ -113,16 +116,53 @@ public async IAsyncEnumerable<FeatureDefinition> GetAllFeatureDefinitionsAsync()
...
+
+            if (_initialized == 0)
+            {
+                IConfiguration featureManagementConfigurationSection = hasFeatureManagementSection ?
+                    _configuration.GetSection(ConfigurationFields.FeatureManagementSectionName) :
+                    _configuration;
+
+                bool hasAzureAppConfigurationFeatureFlagSchema = HasAzureAppConfigurationFeatureFlagSchema(featureManagementConfigurationSection);
+
+                if (Interlocked.Exchange(ref _initialized, 1) != 1)
```
jimmyca15, line 155: Interlocked usage here can allow a simultaneously executing thread to bypass the critical section and return from the method without initializing. With the requirement to initialize here a lock cannot be avoided. Single usage of lock on initialization should be acceptable.

src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
```diff
@@ -113,16 +117,58 @@ public async IAsyncEnumerable<FeatureDefinition> GetAllFeatureDefinitionsAsync()
...
+            if (!hasFeatureManagementSection && !RootConfigurationFallbackEnabled)
+            {
+                Logger?.LogDebug($"No configuration section named '{ConfigurationFields.FeatureManagementSectionName}' was found.");
+            }
+
+            if (_initialized == 0)
+            {
+                IConfiguration featureManagementConfigurationSection = hasFeatureManagementSection ?
+                    _configuration.GetSection(ConfigurationFields.FeatureManagementSectionName) :
+                    _configuration;
```
jimmyca15, line 152: Why is there fallback to the root configuration without a check for `RootConfigurationFallbackEnabled`

src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
```diff
@@ -193,64 +251,197 @@ We support
...
+                .Any(section => string.Equals(section.Key, AzureAppConfigurationFeatureFlagFields.FeatureFlagsSectionName, StringComparison.OrdinalIgnoreCase));
+
+            if (hasFeatureFlagsSection)
+            {
+                IConfigurationSection featureFlagsConfigurationSection = featureManagementConfiguration.GetSection(AzureAppConfigurationFeatureFlagFields.FeatureFlagsSectionName);
+
+                //
+                // FeatureFlags section is an array
+                return string.IsNullOrEmpty(featureFlagsConfigurationSection.Value) && 
+                    featureFlagsConfigurationSection.GetChildren()
```
jimmyca15, line 440: If `featureFlagsConfigurationSection.GetChildren()` returns empty, then `All` would return true. Seems unintended.

src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
```diff
@@ -193,64 +251,197 @@ We support
...
+        }
+
+        private static bool HasAzureAppConfigurationFeatureFlagSchema(IConfiguration featureManagementConfiguration)
+        {
+            bool hasFeatureFlagsSection = featureManagementConfiguration.GetChildren()
+                .Any(section => string.Equals(section.Key, AzureAppConfigurationFeatureFlagFields.FeatureFlagsSectionName, StringComparison.OrdinalIgnoreCase));
+
+            if (hasFeatureFlagsSection)
+            {
+                IConfigurationSection featureFlagsConfigurationSection = featureManagementConfiguration.GetSection(AzureAppConfigurationFeatureFlagFields.FeatureFlagsSectionName);
```
jimmyca15, line 435: GetSection can't return null can it?

src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
```diff
@@ -193,64 +251,197 @@ We support
...
+
+                return featureFlagsSection.GetChildren();
+            }
+
+            return featureManagementConfigurationSection.GetChildren();
+        }
+
+        private static bool HasAzureAppConfigurationFeatureFlagSchema(IConfiguration featureManagementConfiguration)
+        {
+            bool hasFeatureFlagsSection = featureManagementConfiguration.GetChildren()
```
jimmyca15, line 430: A few suggestions in this section. Get the section while checking for it's existence. Check if feature flags section has any children while doing an all check.

```
            IConfigurationSection featureFlagsConfigurationSection = featureManagementConfiguration
                    .GetChildren()
                    .FirstOrDefault(section => 
                        string.Equals(
                            section.Key,
                            AzureAppConfigurationFeatureFlagFields.FeatureFlagsSectionName,

            if (featureFlagsConfigurationSection != null)
            {
                if (!string.IsNullOrEmpty(featureFlagsConfigurationSection.Value))
                {
                    return false;
                }

                featureFlagChildren = featureFlagsConfigurationSection.GetChildren();

                return featureFlagChildren.Any() && featureFlagChildren.All(section => int.TryParse(section.Key, out int _));
            }

            return false;
```

src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
jimmyca15: Perhaps this used to return above previously. It needs to be corrected given the change.

src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
```diff
@@ -193,64 +251,197 @@ We support
...
+        }
+
+        private IEnumerable<IConfigurationSection> GetFeatureDefinitionSections()
+        {
+            if (!_configuration.GetChildren().Any())
+            {
+                return Enumerable.Empty<IConfigurationSection>();
+            }
+
+            bool hasFeatureManagementSection = _configuration.GetChildren()
```
jimmyca15, line 406: Better to get the section with FirstOrDefault, rather than just check for it's existence, else you will just need to get it again later.

src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
```diff
@@ -193,64 +251,197 @@ We support
...
+
+            return new FeatureDefinition()
+            {
+                Name = featureName,
+                EnabledFor = enabledFor,
+                RequirementType = requirementType
+            };
+        }
+
+        private string GetFeatureFlagSectionName(IConfigurationSection section)
```
jimmyca15, line 389:
```suggestion
        private string GetFeatureFlagName(IConfigurationSection section)
```

jimmyca15: Or
```suggestion
        private string GetFeatureName(IConfigurationSection section)
```

Given the caller uses a variable named `featureName`

src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
zhiyuanliang-ms: This is intended. If there is "FeatureFlags" section, either it is empty or an array, we consider it as using the App Config schema.

src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
```diff
@@ -113,16 +117,55 @@ public async IAsyncEnumerable<FeatureDefinition> GetAllFeatureDefinitionsAsync()
...
+                {
+                    featureManagementConfigurationSection = _configuration;
+                }
+
+                bool hasAzureAppConfigurationFeatureFlagSchema = featureManagementConfigurationSection != null && 
+                    HasAzureAppConfigurationFeatureFlagSchema(featureManagementConfigurationSection);
+
+                lock (_lock)
+                {
+                    if (Interlocked.Read(ref _initialized) == 0)
```
RichardChen820, line 155: lock can sure only one thread can enter this block, what is the reason for using `Interlocked.Read(ref _initialized)` to atomically get the value?

Mark int32 type `_initialized` as `volatile` then judge `_initialized == 0` is fine. Am I missing anything?

RichardChen820: Or use ` if (Interlocked.CompareExchange(ref _initialized, 1, 0))==0 {_azureAppConfigurationFeatureFlagSchemaEnabled = hasAzureAppConfigurationFeatureFlagSchema; }` is also fine.

jimmyca15: Interlocked avoids cached read from CPU. Marking the variable as volatile will have lasting effect throughout the lifetime of the application whereas interlocked usage in the init check only happens during initialization.

mrm9084:
@jimmyca15 @zhiyuanliang-ms 
 > We want to augment the FeatureManagement-Dotnet to support the [Azure App Configuration feature flag schema] (https://github.com/Azure/AppConfiguration/blob/main/docs/FeatureManagement/FeatureFlag.v1.1.0.schema.json)
  {
      "FeatureManagement": {
          "FeatureFlags": [
              // App Configuration Feature Flag objects
          ]
      }
  }

Isn't this the "V2" schema. I thought we decided not to use it?

zhiyuanliang-ms:
@mrm9084 This is not the V2 schema because this PR is targeting on the main branch. The variant and telemetry part are implemented in the preview/release v4 branch which is for the ongoing preview release. We didn't plan the App Config feature flag schema support for the preview release.

zhiyuanliang-ms:
@zhenlan @jimmyca15 @RichardChen820 As discussed in the email "Microsoft Feature Flag schema should be the name we use going forward", I made some update to this PR.

Issue #325, opened 2023-12-01
Cache Entry does not set size


mjameson-se:
We're in the process of implementing a feature using Azure App Config to power feature gates, but hit a rather annoying snag.
```
System.InvalidOperationException: Cache entry must specify a value for Size when SizeLimit is set.
   at Microsoft.Extensions.Caching.Memory.MemoryCache.SetEntry(CacheEntry entry)
   at Microsoft.Extensions.Caching.Memory.CacheExtensions.Set[TItem](IMemoryCache cache, Object key, TItem value, MemoryCacheEntryOptions options)
   at Microsoft.FeatureManagement.FeatureManager.BindSettings(IFeatureFilterMetadata filter, FeatureFilterEvaluationContext context, Int32 filterIndex)
   at Microsoft.FeatureManagement.FeatureManager.IsEnabledAsync[TContext](String feature, TContext appContext, Boolean useAppContext)
   at Microsoft.FeatureManagement.Mvc.FeatureGateAttribute.<>c__DisplayClass10_0.<<OnActionExecutionAsync>b__0>d.MoveNext(
```
The Feature Manager is getting a shared IMemoryCache from the dependency injection container and tries to add cache entries to it without specifying a size. The only work-around for this seems to be to alter all of *our* code to register and pull a keyed IMemoryCache instead of the default unkeyed instance.

jimmyca15:
@mjameson-se

Thanks for raising this. Looks like an issue introduced in **3.1.0** release.

@zhiyuanliang-ms We need to specify size for all entries placed in the cache.

jimmyca15:
@zhiyuanliang-ms 

To repro this issue add the following in Startup.cs of the FeatureFlagDemo example project.

```
            services.AddMemoryCache(mo =>
            {
                mo.SizeLimit = 50_000_000;
            });
```

--- 

This issue occurs when the shared memory cache in DI is configured to have a size limit.

jimmyca15:
Hello @mjameson-se this issue has been fixed and released in version [3.1.1](https://github.com/microsoft/FeatureManagement-Dotnet/releases/tag/3.1.1).

PR #327, opened 2023-12-04
Add size for cache entry

zhiyuanliang-ms:
Try to fix the bug introduced by 3.1.0.
The feature manager uses the shared memory cache and it doesn't specify the entry size. This will cause error if the shared memory cache has a size limit. #325

### DIFF
--- a/src/Microsoft.FeatureManagement/FeatureManager.cs
+++ b/src/Microsoft.FeatureManagement/FeatureManager.cs
@@ -321,7 +321,7 @@ namespace Microsoft.FeatureManagement
                     {
                         SlidingExpiration = ParametersCacheSlidingExpiration,
                         AbsoluteExpirationRelativeToNow = ParametersCacheAbsoluteExpirationRelativeToNow
-                    });
+                    }.SetSize(1));
             }
             else
             {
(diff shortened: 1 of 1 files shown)


src/Microsoft.FeatureManagement/FeatureManager.cs
```diff
@@ -321,7 +321,7 @@ private void BindSettings(IFeatureFilterMetadata filter, FeatureFilterEvaluation
                     {
                         SlidingExpiration = ParametersCacheSlidingExpiration,
                         AbsoluteExpirationRelativeToNow = ParametersCacheAbsoluteExpirationRelativeToNow
-                    });
+                    }.SetSize(1));
```
jimmyca15, line 324: MemoryCacheEntryOptions has a size property. We should initialize it with the other properties rather than adding a new calling pattern.

```
,
                    new MemoryCacheEntryOptions
                    {
                        SlidingExpiration = ParametersCacheSlidingExpiration,
                        AbsoluteExpirationRelativeToNow = ParametersCacheAbsoluteExpirationRelativeToNow,
                        Size = 1
                    });
```

jimmyca15 (approved):

PR #329, opened 2023-12-05
Adjust title level in README

zhiyuanliang-ms:


README.md
```diff
@@ -141,7 +144,7 @@ A `RequirementType` of `All` changes the traversal. First, if there are no filte
 
 In the above example, `FeatureW` specifies a `RequirementType` of `All`, meaning all of it's filters must evaluate to true for the feature to be enabled. In this case, the feature will be enabled for 50% of users during the specified time window.
     
-### Service Registration
+## Service Registration
```
jimmyca15, line 147: I think the root title should be **Consumption** and then it goes over `featureManager.IsEnabledAsync` right away.

Then it Service Registration, and DI can be following subsections.

Feature check section would no longer be needed.

And I'd suggest moving the scoped services section last in the **Consumption** parent section.

So

Consumption
  * Service Registration
  * Dependency Injection
  * Scoped Feature Management Services

README.md
```diff
@@ -439,7 +447,7 @@ public void ConfigureServices(IServiceCollection services)
 **Advanced:** `IHttpContextAccessor`/`HttpContext` should not be used in the Razor components of server-side Blazor apps. [The recommended approach](https://learn.microsoft.com/en-us/aspnet/core/blazor/security/server/interactive-server-side-rendering?view=aspnetcore-7.0#ihttpcontextaccessorhttpcontext-in-razor-components) for passing http context in Blazor apps is to copy the data into a scoped service. For Blazor apps, `AddScopedFeatureManagement` should be used to register the feature management services.
 Please refer to the `Scoped Feature Management Services` section for more details.
 
-## Providing a Context For Feature Evaluation
+### Providing a Context For Feature Evaluation
```
jimmyca15, line 450: I don't consider this a subsection of implementing a filter.

jimmyca15: And perhaps the built-in filters section should be promoted a level.

jimmyca15 (approved):

PR #330, opened 2023-12-12
Mention TargetingContext and fix wording in Variants section of readme

amerjusupovic:


amerjusupovic:
I was considering adding a larger callout somewhere, like changing the `TargetingContextAccessor` section, but it would probably require some rearranging of sections. Open to suggestions there.

zhiyuanliang-ms:
My thought is that we have the "Targeting" sections structure like:
## Targeting
Introduce the concept of audience here.
#### Targeting Exclusion
Bring the concept of exclusion to the begining.

### ITargetingContextContextAccessor
Promote this section a level since it will be used by both allocation and targeting filter.

### Targeting in Variant Allocation
### Targeting in a Web Application
### Targeting in a Console Application
### Targeting Evaluation Options

zhiyuanliang-ms:
@amerjusupovic Did you lose some changes in the latest merge commit?

README.md
```diff
@@ -767,7 +767,7 @@ In the above example, if the feature is not enabled, `GetVariantAsync` would ret
...
 
-Allocation logic is similar to the [Microsoft.Targeting](./README.md#MicrosoftTargeting) feature filter, but there are some parameters that are present in targeting that aren't in allocation, and vice versa. Like targeting, allocation can only access user information once an implementation of  `ITargetingContextAccessor` has been added to the service collection. The outcomes of targeting and allocation are not related.
+Allocation logic is similar to the [Microsoft.Targeting](./README.md#MicrosoftTargeting) feature filter, but there are some parameters that are present in targeting that aren't in allocation, and vice versa. Like targeting, allocation can only access user information once an implementation of `ITargetingContextAccessor` has been added to the service collection. Without this service, allocation can only return default variants if they're specified. The outcomes of targeting and allocation are not related.
```
zhiyuanliang-ms, line 817: Is this change lost?

amerjusupovic: Talked with @jimmyca15 so I removed this, changed some titles to fit the sections better, and included targeting context earlier in the doc.

README.md
```diff
@@ -735,28 +737,43 @@ IConfigurationSection variantConfiguration = variant.Configuration;
...
-"Variants": [
-    { 
-        "Name": "Big", 
-        "ConfigurationReference": "ShoppingCart:Big" 
-    },  
-    { 
-        "Name": "Small", 
-        "ConfigurationValue": {
-            "Size": 300
+{
```
zhiyuanliang-ms, line 747:
``` JavaScript
"MyFlag":
{   
    "Variants": [
        { 
            "Name": "Big", 
            "ConfigurationReference": "ShoppingCart:Big" 
        },  
        { 
            "Name": "Small", 
            "ConfigurationValue": {
                "Size": 300
            }
        } 
    ],
    "EnabledFor": [
        {
            "Name": "AlwaysOn"
        }
    ]
}
```
It would be more consistent to remove the "FeatureManagement" section and use JavaScript format to get the highlight.

README.md
```diff
@@ -735,28 +737,43 @@ IConfigurationSection variantConfiguration = variant.Configuration;
 // Do something with the resulting variant and its configuration
 ```
 
-### Setting a Variant's Configuration
+### Defining a Feature's Variants
```
zhiyuanliang-ms, line 740: I think "Defining Variants" would be better since the above section is called "Getting Variants"

README.md
```diff
@@ -720,9 +720,11 @@ public class Variant
 }
 ```
 
-### Getting a Feature's Variant
+### Getting Variants
```
jimmyca15, line 723: I think it would be better if the code snippet came directly after this part "For each feature, a variant can be retrieved using the IVariantFeatureManager's GetVariantAsync method."

then after the code snippet the extra information was provided.

jimmyca15: I was thinking in our doc it might be helpful to mention that .NET makes it easy to bind IconfigurationSection to a type. After all, this is how I expect most people to consume a features variants. After seeing this, I think a second code snippet could be provided. something like below

---

Once a variant is retrieved, the configuration of a variant can be used directly as an `IConfigurationSection` from the variants `Configuration` property. Another option is to bind the configuration to a type using .NET's configuration binding pattern.

```
IConfigurationSection variantConfiguration = variant.Configuration;

MyFeatureSettings settings = variant
    .Configuration
    .Bind<MyFeatureSettings>();
```

README.md
jimmyca15: I like the suggestion.

jimmyca15: Same with `Allocating Variants`

README.md
```diff
@@ -804,21 +821,21 @@ The `Allocation` setting of a feature flag has the following properties:
...
+| `DefaultWhenEnabled` | Specifies which variant should be used when a variant is requested while the feature is considered enabled and no other variant was assigned to the user. |
+| `User` | Specifies a variant and a list of users to whom that variant should be assigned. | 
+| `Group` | Specifies a variant and a list of groups the current user has to be in for that variant to be assigned. |
+| `Percentile` | Specifies a variant and a percentage range the user's calculated percentage has to fit into for that variant to be assigned. |
 | `Seed` | The value which percentage calculations for `Percentile` are based on. The percentage calculation for a specific user will be the same across all features if the same `Seed` value is used. If no `Seed` is specified, then a default seed is created based on the feature name. |
 
-In the above example, if the feature is not enabled, `GetVariantAsync` would return the variant allocated by `DefaultWhenDisabled`, which is `Small` in this case. 
+In the above example, if the feature is not enabled, the feature manager will assign the variant allocated by `DefaultWhenDisabled` to the current user, which is `Small` in this case.
```
jimmyca15, line 830:
```suggestion
In the above example, if the feature is not enabled, the feature manager will assign the variant marked as `DefaultWhenDisabled` to the current user, which is `Small` in this case.

```

README.md
```diff
@@ -804,21 +821,21 @@ The `Allocation` setting of a feature flag has the following properties:
...
 
-If the feature is enabled, the feature manager will check the `User`, `Group`, and `Percentile` allocations in that order to allocate a variant for this feature. If the user being evaluated is named `Marsha`, in the group named `Ring1`, or the user happens to fall between the 0 and 10th percentile calculated with the given `Seed`, then the specified variant is returned for that allocation. In this case, all of these would return the `Big` variant. If none of these allocations match, the `DefaultWhenEnabled` variant is returned, which is `Small`.
+If the feature is enabled, the feature manager will check the `User`, `Group`, and `Percentile` allocations in that order to assign a variant. If the user being evaluated is named `Marsha`, in the group named `Ring1`, or the user happens to fall between the 0 and 10th percentile calculated with the given `Seed`, then the specified variant is assigned to the user. In this case, all of these would return the `Big` variant. If none of these allocations match, the user is assigned the `DefaultWhenEnabled` variant, which is `Small`.
```
jimmyca15, line 832:
```suggestion
If the feature is enabled, the feature manager will check the `User`, `Group`, and `Percentile` allocations in that order to assign a variant. For this particular example, if the user being evaluated is named `Marsha`, in the group named `Ring1`, or the user happens to fall between the 0 and 10th percentile, then the specified variant is assigned to the user. In this case, all of these would return the `Big` variant. If none of these allocations match, the user is assigned the `DefaultWhenEnabled` variant, which is `Small`.

```

jimmyca15: I would leave seed out of the explanation. In my opinion it raises the complexity too high.

jimmyca15: Also, can we leave seed out of the flag example that covers overriding enabled/disabled with variants.

README.md
```diff
@@ -720,9 +720,11 @@ public class Variant
 }
 ```
 
-### Getting a Feature's Variant
+### Getting Variants
 
-A feature's variant can be retrieved using the `IVariantFeatureManager`'s `GetVariantAsync` method.
+For each feature, a variant can be retrieved using the `IVariantFeatureManager`'s `GetVariantAsync` method. The variant returned is dependent on the
+user currently being evaluated, and that information is obtained from an instance of `TargetingContext`. This context can either be passed in when
+calling `GetVariantAsync` or it can be automatically retrieved from an implementation of `ITargetingContextAccessor` if one is registered.
```
jimmyca15, line 727: I feel that this should link to the `ITargetingContextAccessor` section for more details.

README.md
zhiyuanliang-ms: We can add an "**Advanced**" section to explain how seed works.

amerjusupovic: I'm not really sure what to add in an Advanced section since the table above this part describes how `Seed` works and I feel like I might be reiterating that. Did you have anything in mind?

README.md
zhiyuanliang-ms:
``` C#
MyFeatureSettings settings = variantConfiguration.Get<MyFeatureSettings>() ?? new MyFeatureSettings();
```
Or
```
MyFeatureSettings settings = new MyFeatureSettings();
variantConfiguration.Bind(settings);
```

README.md
zhiyuanliang-ms: Oh, I didn't notice the table. Then we don't need a dedicated section for the seed usage.

README.md
```diff
@@ -703,7 +703,7 @@ In the above example, the feature will be enabled for users named `Jeff` and `Al
...
 
-Variants enable a feature flag to become more than a simple on/off flag. A variant represents a value of a feature flag that can be a string, a number, a boolean, or even a configuration object. A feature flag that declares variants should define under what circumstances each variant should be used, which is covered in greater detail in the [Allocating a Variant](./README.md#allocating-a-variant) section.
+Variants enable a feature flag to become more than a simple on/off flag. A variant represents a value of a feature flag that can be a string, a number, a boolean, or even a configuration object. A feature flag that declares variants should define under what circumstances each variant should be used, which is covered in greater detail in the [Allocating a Feature's Variants](./README.md#allocating-a-features-variants) section.
```
jimmyca15, line 706: the reference should be Allocating Variants now.

jimmyca15 (approved):

README.md
jimmyca15: Given some other section names were updated, it's worth a double check if any links need updated pointing to those.

amerjusupovic: Ok I'll double check

PR #335, opened 2023-12-19
Feature-based Injection

### LINKED ISSUE #39: Feature-Based Injection
This is a migration of the issue https://github.com/Azure/AppConfiguration/issues/68 for better tracking and coming to a solution.

zhiyuanliang-ms:
This PR is targeting on the potential user case is to inject services based on variants. For example, if the variant was an algorithm name, and a controller constructor took IHashAlgorithm as a constructor parameter. How could we make it so that the implementation was dependent upon an assigned variant?

Also, it tries to resolve #39. 

This draft PR provides one APIs: `FeatureManagementBuilder.AddFeaturedService<TService>`

Example usage can be found here: examples/FeatureBasedInjectionPOC/Program.cs

Open to suggestion for the design of feature-based injection.


examples/FeatureBasedInjectionPOC/Program.cs
```diff
@@ -0,0 +1,77 @@
...
+    targetingContextAccessor.Current = new TargetingContext
+    {
+        UserId = "UserOmega"
+    };
+
+    algorithm = await featuredAlgorithm.GetAsync(CancellationToken.None);
+
+    variant = await featureManager.GetVariantAsync("MyFeature", CancellationToken.None);
+
+    Console.WriteLine($"Get algorithm {algorithm?.Name ?? "Null"} because the feature variant is {variant?.Name ?? "Null"}");
```
rossgrambo, line 76:
```suggestion
    string[] userIds = {"Guest", "UserBeta", "UserSigma", "UserOmega"};

    foreach (string userId in userIds)
    {
        targetingContextAccessor.Current = new TargetingContext
        {
            UserId = userId
        };

        IAlgorithm algorithm = await featuredAlgorithm.GetAsync(CancellationToken.None);

        Variant variant = await featureManager.GetVariantAsync("MyFeature", CancellationToken.None);

        Console.WriteLine($"Get algorithm {algorithm?.Name ?? "Null"} because the feature variant is {variant?.Name ?? "Null"}");
    }
```

src/Microsoft.FeatureManagement/ServiceCollectionExtensions.cs
```diff
@@ -12,6 +12,7 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Reflection;
```
amerjusupovic, line 15: I don't see any changes in this file related to System.Reflection, is this used?

src/Microsoft.FeatureManagement/IFeaturedService.cs
```diff
@@ -0,0 +1,13 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+//
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.FeatureManagement
+{
+    public interface IFeaturedService<TService> where TService : class
```
zhenlan, line 9: Can we add comments for each input/output parameter? I am wondering the reasoning for the choice between FeaturedService vs. FeatureService.

jimmyca15: I think the name wasn't discussed enough actually. Thinking on the overall design that we landed on I would suggest we call this `IVariantServiceProvider` with a single method GetService.

This is similar to built-in [IServiceProvider](https://learn.microsoft.com/en-us/dotnet/api/system.iserviceprovider?view=net-8.0).

zhiyuanliang-ms: Sounds good. I like the name `IVariantServiceProvider`

examples/FeatureBasedInjectionPOC/IAlgorithm.cs
```diff
@@ -0,0 +1,7 @@
+﻿namespace FeatureBasedInjectionPOC
+{
+    public interface IAlgorithm
+    {
+        public string Name {  get; }
```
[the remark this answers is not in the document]
zhiyuanliang-ms: Sounds good. "Style" is better. This is just a POC. So my previous plan is to remove this example when this PR is ready for review.

src/Microsoft.FeatureManagement.Telemetry.ApplicationInsights/TelemetryClientExtensions.cs
```diff
@@ -0,0 +1,57 @@
+﻿// Copyright (c) Microsoft Corporation.
```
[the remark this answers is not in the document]
zhiyuanliang-ms: I have merged with the latest preview branch. Now it should be removed.

src/Microsoft.FeatureManagement/ServiceCollectionExtensions.cs
zhiyuanliang-ms: Good catch! I forgot to remove it.

src/Microsoft.FeatureManagement/IFeaturedService.cs
zhiyuanliang-ms: @jimmyca15 Should the GetService method be async? Since our implementation need to get variant from feature manager through GetVariantAsync.

jimmyca15: @zhiyuanliang-ms yes it needs to be async. ValueTask.

src/Microsoft.FeatureManagement/VariantServiceProvider.cs
```diff
@@ -0,0 +1,86 @@
...
+        {
+            Debug.Assert(_variantFeatureName != null);
+
+            Variant variant = await _featureManager.GetVariantAsync(_variantFeatureName, cancellationToken);
+
+            TService implementation = null;
+
+            if (variant != null)
+            {
+                implementation = _services.FirstOrDefault(service =>
```
rossgrambo, line 63: We might want to do this work upfront- if we added an "Init()" function, we could call it in our BuilderExtension and have it place the services into a Map<string, TService> where the string is either the alias or the name. 

This way we don't have to do reflection on every request/time GetServiceAsync is called.

What do you think?

src/Microsoft.FeatureManagement/FeatureManagementBuilderExtensions.cs
```diff
@@ -39,6 +40,41 @@ public static class FeatureManagementBuilderExtensions
...
+        public static IFeatureManagementBuilder AddVariantServiceProvider<TService>(this IFeatureManagementBuilder builder, string featureName) where TService : class
+        {
+            if (string.IsNullOrEmpty(featureName))
+            {
+                throw new ArgumentNullException(nameof(featureName));
+            }
+
+            if (builder.Services.Any(descriptor => descriptor.ServiceType == typeof(IFeatureManager) && descriptor.Lifetime == ServiceLifetime.Scoped))
+            {
+                builder.Services.AddScoped<IVariantServiceProvider<TService>>(sp => new VariantServiceProvider<TService>(
```
rossgrambo, line 58: Does this service make sense as a scoped service? From my understanding- the variant service will be a Singleton, right? Or, if the variant service could be scoped, then perhaps this should check the descriptor.Lifetime of one of the variant services instead of FeatureManager.

src/Microsoft.FeatureManagement/VariantServiceProvider.cs
```diff
@@ -0,0 +1,86 @@
...
+            Debug.Assert(variant != null);
+
+            string implementationName = ((VariantServiceAliasAttribute)Attribute.GetCustomAttribute(implementationType, typeof(VariantServiceAliasAttribute)))?.Alias;
+
+            if (implementationName == null)
+            {
+                implementationName = implementationType.Name;
+            }
+
+            return string.Equals(implementationName, variant.Name, StringComparison.OrdinalIgnoreCase);
```
zhenlan, line 83: Just to confirm, is it case insensitive for feature filter too?

zhiyuanliang-ms: Yes. Feature filter name is case insensitive. https://github.com/microsoft/FeatureManagement-Dotnet/blob/main/src/Microsoft.FeatureManagement/FeatureManager.cs#L377

src/Microsoft.FeatureManagement/FeatureManagementBuilderExtensions.cs
zhiyuanliang-ms: VariantServiceProvider will consume feature manager. If the feature manager is scoped and we register the variant service provider as singleton, an exception will be thrown. 

> check the descriptor.Lifetime of one of the variant services instead of FeatureManager.
This will fail if we cannot find variant services in the service collection, because users may register their implementation after calling our API like:
```C#
            .AddVariantServiceProvider<IAlgorithm>("MyFeature");  
services.AddSingleton<IAlgorithm, AlgorithmAlpha>();
```

src/Microsoft.FeatureManagement/VariantServiceProvider.cs
zhiyuanliang-ms: > if we added an "Init()" function, we could call it in our BuilderExtension and have it place the services into a Map<string, TService> where the string is either the alias or the name.

This sounds not feasible for me. Could you provide more details about how this can be achieved?

If we want to have a Map<string, TService>, the value of each key is a TService instance. This could not be done during the service registration stage. In our BuilderExtension, the registered service are actually `ServiceDescriptor` in `IServiceCollection` and they are still not instantiated.

zhiyuanliang-ms: Added in 7120abd97a0ee8ed44157d174198b0aeabc926d6  @rossgrambo

src/Microsoft.FeatureManagement/FeatureManagementBuilderExtensions.cs
rossgrambo: To clarify, I'm curious about the actual variant services- not the provider. Ie:
```
        .AddVariantServiceProvider<IAlgorithm>("MyFeature");  
services.AddSingleton<IAlgorithm, AlgorithmAlpha>(); // This 
```
vs
```
        .AddVariantServiceProvider<IAlgorithm>("MyFeature");  
services.AddScoped<IAlgorithm, AlgorithmAlpha>(); // vs this
```

zhiyuanliang-ms: If the user registers variant service as scoped and registers feature management as singleton, then the user mess up the service lifetime. Scoped variant service will be consumed by the singleton variant service provider, an exception will occur.
```C#
        .AddVariantServiceProvider<IAlgorithm>("MyFeature");  
services.AddScoped<IAlgorithm, AlgorithmAlpha>();
```

The variant services are registered by user, so they should choose the right lifetime and the correct method to register feature management (AddScopedFeatureManagement or AddFeatureManagement).

For me, this is the similar case for targeting context accessor, before we provide `WithTargeting` API, user need to register `ITargetingContextAccessor` by themselves and they can register it as scoped/transient whatever they want. We didn't register the targeting filter or feature manager as scoped/transient if we find `ITargetingContextAccessor` is registered as scoped. Instead, if user really wants to use a scoped `ITargetingContextAccessor`, they should call `AddScopedFeatureManagement` to register the whole feature management as scoped.

src/Microsoft.FeatureManagement/FeatureManagementBuilderExtensions.cs
```diff
@@ -39,6 +40,41 @@ public static class FeatureManagementBuilderExtensions
             return builder;
         }
 
+        /// <summary>
+        /// Adds a <see cref="VariantServiceProvider{TService}"/> to the feature management system.
+        /// </summary>
+        /// <param name="builder">The <see cref="IFeatureManagementBuilder"/> used to customize feature management functionality.</param>
+        /// <param name="featureName">The variant feature flag used to assign variants. The <see cref="VariantServiceProvider{TService}"/> will return different implementations of TService according to the assigned variant.</param>
+        /// <returns>A <see cref="IFeatureManagementBuilder"/> that can be used to customize feature management functionality.</returns>
+        public static IFeatureManagementBuilder AddVariantServiceProvider<TService>(this IFeatureManagementBuilder builder, string featureName) where TService : class
```
jimmyca15, line 49:
```suggestion
        public static IFeatureManagementBuilder RegisterVariantService<TService>(this IFeatureManagementBuilder builder, string featureName) where TService : class
```

jimmyca15: I suggest the name because I think it breaks a way from typical `Add`. Here we require a parameter which is the feature name that is mapped to the variant service.

jimmyca15: I also wouldn't mind `WithVariantService` similar to `WithTargeting`

src/Microsoft.FeatureManagement/FeatureManagementBuilderExtensions.cs
```diff
@@ -39,6 +40,41 @@ public static class FeatureManagementBuilderExtensions
...
+                throw new ArgumentNullException(nameof(featureName));
+            }
+
+            if (builder.Services.Any(descriptor => descriptor.ServiceType == typeof(IFeatureManager) && descriptor.Lifetime == ServiceLifetime.Scoped))
+            {
+                builder.Services.AddScoped<IVariantServiceProvider<TService>>(sp => new VariantServiceProvider<TService>(
+                    sp.GetRequiredService<IEnumerable<TService>>(),
+                    sp.GetRequiredService<IVariantFeatureManager>())
+                {
+                    VariantFeatureName = featureName,
```
jimmyca15, line 62:
```suggestion
                    FeatureName = featureName,
```

src/Microsoft.FeatureManagement/FeatureManagementBuilderExtensions.cs
```diff
@@ -39,6 +40,41 @@ public static class FeatureManagementBuilderExtensions
             return builder;
         }
 
+        /// <summary>
+        /// Adds a <see cref="VariantServiceProvider{TService}"/> to the feature management system.
+        /// </summary>
+        /// <param name="builder">The <see cref="IFeatureManagementBuilder"/> used to customize feature management functionality.</param>
+        /// <param name="featureName">The variant feature flag used to assign variants. The <see cref="VariantServiceProvider{TService}"/> will return different implementations of TService according to the assigned variant.</param>
```
jimmyca15, line 47:
```suggestion
        /// <param name="featureName">The feature flag that should be used to determine which variant of the service should be used. The <see cref="VariantServiceProvider{TService}"/> will return different implementations of TService according to the assigned variant.</param>
```

src/Microsoft.FeatureManagement/FeatureManagementBuilderExtensions.cs
zhiyuanliang-ms: I like `WithVariantService` which makes our API's method name style more consistent.

src/Microsoft.FeatureManagement/VariantServiceProvider.cs
```diff
@@ -0,0 +1,90 @@
...
+        /// <summary>
+        /// The variant feature flag used to assign variants.
+        /// </summary>
+        public string VariantFeatureName
+        {
+            get => _variantFeatureName;
+
+            init
+            {
+                _variantFeatureName = value ?? throw new ArgumentNullException(nameof(value));
```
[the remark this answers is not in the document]
zhiyuanliang-ms: > It's a required for the successful usage of the object isn't it?

Yes.
I used to use the feature flag name as a constructor parameter in this commit: 176cf78616366b29309c80c3cf1dad9ddb18b3e6

But then I decide to put it in a public property, because in this way, which variant feature flag is bound to this variant service provider can be known from the outside. 
I think the interface type and the feature flag name are equal in status. We can know the interface type from the type `VariantServiceProvider<TService>` but we won't know the feature flag name if we put it in a private property.

I know that it is unnecessary to expose the feature flag name to public at this moment , the `VariantServiceProvider` class is actually internal. So I won't mind putting the feature flag name in the constructor parameter.

zhiyuanliang-ms:
An exception should be thrown if there are multiple `IVariantServiceProvider<TService>` being registered.

src/Microsoft.FeatureManagement/FeatureManagementBuilderExtensions.cs
```diff
@@ -8,6 +8,7 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Reflection;
```
rossgrambo, line 11: I don't think this is used.

src/Microsoft.FeatureManagement/FeatureManagementBuilderExtensions.cs
```diff
@@ -39,6 +40,48 @@ public static class FeatureManagementBuilderExtensions
...
         }
 
+        /// <summary>
+        /// Adds a <see cref="VariantServiceProvider{TService}"/> to the feature management system.
+        /// </summary>
+        /// <param name="builder">The <see cref="IFeatureManagementBuilder"/> used to customize feature management functionality.</param>
+        /// <param name="featureName">The feature flag that should be used to determine which variant of the service should be used. The <see cref="VariantServiceProvider{TService}"/> will return different implementations of TService according to the assigned variant.</param>
+        /// <returns>A <see cref="IFeatureManagementBuilder"/> that can be used to customize feature management functionality.</returns>
+        /// <exception cref="ArgumentNullException">Thrown if feature name parameter is null.</exception>
+        /// <exception cref="InvalidOperationException">Thrown if the variant service of the type has been added.</exception>
```
rossgrambo, line 50:
```suggestion
        /// <exception cref="InvalidOperationException">Thrown if a variant service provider of the type has already been added.</exception>
```

src/Microsoft.FeatureManagement/FeatureManagementBuilderExtensions.cs
zhiyuanliang-ms: Oh, my bad. I forget to remove it. Good catch.

rossgrambo (approved):

src/Microsoft.FeatureManagement/VariantServiceProvider.cs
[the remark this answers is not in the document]
jimmyca15: Please put it in the constructor. It's especially unexpected that `GetServiceAsync` asserts that feature name is not null when nothing about the instantiation of the type enforces it to be non-null.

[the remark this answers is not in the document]
jimmyca15: `Debug.Assert` should only be used to assert conditions that are validated by public entry point and a reader can verify to be valid.

src/Microsoft.FeatureManagement/FeatureManagementBuilderExtensions.cs
```diff
@@ -39,6 +39,48 @@ public static class FeatureManagementBuilderExtensions
...
+        public static IFeatureManagementBuilder WithVariantService<TService>(this IFeatureManagementBuilder builder, string featureName) where TService : class
+        {
+            if (string.IsNullOrEmpty(featureName))
+            {
+                throw new ArgumentNullException(nameof(featureName));
+            }
+            
+            if (builder.Services.Any(descriptor => descriptor.ServiceType == typeof(IVariantServiceProvider<TService>)))
+            {
+                throw new InvalidOperationException($"Variant services of {typeof(TService)} has been added.");
```
jimmyca15, line 59:
```suggestion
                throw new InvalidOperationException($"A variant service of {typeof(TService)} has already been added.");
```

jimmyca15:
```suggestion
                throw new InvalidOperationException($"Variant services of {typeof(TService).FullName} has been added.");
```

jimmyca15 (approved):
2 unresolved minor comments on my end. Everything else looks good. Thanks @zhiyuanliang-ms !

PR #336, opened 2023-12-19
Update telemetry schema in feature flags.

jimmyca15:
This PR updates the schema for telemetry in feature flags

---

### Before

```
{
    "TelemetryEnabled": true,
    "TelemetryMetadata": {
        "Foo": "Bar"
    }
}
```

### After

```
{
    "Telemetry": {
        "Enabled": true,
        "Metadata": {
            "Foo": "Bar"
        }
    }
}
```


README.md
```diff
@@ -889,7 +891,14 @@ For flags defined in `appsettings.json`, that is done by using the `TelemetryEna
 }
 ```
 
-The appsettings snippet above defines a flag named `MyFlag` that is enabled for telemetry.
+The appsettings snippet above defines a flag named `MyFlag` that is enabled for telemetry. This is indicated by the `Telemetry` object which sets `Enabled` to true. The value of the `Enabled` property must be `true` to publish telemetry for the flag.
```
zhiyuanliang-ms, line 894:
```suggestion
The appsettings snippet above defines a feature flag named `MyFlag` that is enabled for telemetry. This is indicated by the `Telemetry` object which sets `Enabled` to true. The value of the `Enabled` property must be `true` to publish telemetry for the flag.

```

README.md
```diff
@@ -889,7 +891,14 @@ For flags defined in `appsettings.json`, that is done by using the `TelemetryEna
...
 
-The appsettings snippet above defines a flag named `MyFlag` that is enabled for telemetry.
+The appsettings snippet above defines a flag named `MyFlag` that is enabled for telemetry. This is indicated by the `Telemetry` object which sets `Enabled` to true. The value of the `Enabled` property must be `true` to publish telemetry for the flag.
+
+The `Telemetry` section of a feature flag has the following properties:
+
+| Property | Description |
+| ---------------- | ---------------- |
+| `Enabled` | Specifies whether telemetry should be published for the feature flag. |
+| `Metadata` | A collection of key-value pairs, modeled as a dictionary, that can be used to add custom metadata on evaluation events that are generated by the flag. |
```
zhiyuanliang-ms, line 901: I feel confusing about the word "generated". My understanding is that the metadata dictionary usually works like custom tag/label to provide some extra information about the feature flag.

I would suggest something like " ..., that can be used to attach custom metadata about the feature flag to evaluation events."

src/Microsoft.FeatureManagement.Telemetry.ApplicationInsights/ApplicationInsightsTelemetryPublisher.cs
```diff
@@ -41,9 +41,9 @@ public ValueTask PublishEvent(EvaluationEvent evaluationEvent, CancellationToken
                 properties["Variant"] = evaluationEvent.Variant.Name;
             }
 
-            if (featureDefinition.TelemetryMetadata != null)
+            if (featureDefinition.Telemetry.Metadata != null)
```
zhiyuanliang-ms, line 44: Is it necessary to add null check for `featureDefinition.Telemetry`?

src/Microsoft.FeatureManagement/ConfigurationFields.cs
```diff
@@ -37,5 +36,10 @@ internal static class ConfigurationFields
         // Other keywords
         public const string NameKeyword = "Name";
         public const string FeatureManagementSectionName = "FeatureManagement";
+
+        // Telemetry keywords
+        public const string Telemetry = "Telemetry";
+        public const string Enabled = "Enabled";
```
zhiyuanliang-ms, line 42: I suggest "TelemetryEnabled" and ”TelemetryMetadata" for two reasons:
1. We had "FeatureStatus" rather than "Status" in Enum keywords
2. It makes the following code more readable
``` C#
telemetryEnabled = telemetrySection.GetValue<bool>(ConfigurationFields.Enabled);
IConfigurationSection telemetryMetadataSection = telemetrySection.GetSection(ConfigurationFields.Metadata);
```

src/Microsoft.FeatureManagement/TelemetryConfiguration.cs
```diff
@@ -0,0 +1,20 @@
...
+
+namespace Microsoft.FeatureManagement
+{
+    /// <summary>
+    /// Defines telemetry related configuration settings available for features.
+    /// </summary>
+    public class TelemetryConfiguration
+    {
+        /// <summary>
+        /// A flag to enable or disable sending telemetry events to the registered <see cref="Telemetry.ITelemetryPublisher"/>.
```
zhiyuanliang-ms, line 11: I would like to drop the word "events". I suggest "publishing/sending telemetry".

zhiyuanliang-ms (approved):

src/Microsoft.FeatureManagement.Telemetry.ApplicationInsights/ApplicationInsightsTelemetryPublisher.cs
jimmyca15: This method should not be invoked if telemetry is null. I will update to throw an argument exception if telemetry is null.

src/Microsoft.FeatureManagement/ConfigurationFields.cs
jimmyca15: I'd stay away from this pattern. That's setting us up to keep an entire object model hierarchy as part of the name of a field. The name of the fields are enabled and metadata. If other parts of the configuration of a flag can use it they should. Ideally we shouldn't have two different keywords for the concept of Enabled and Metadata.

src/Microsoft.FeatureManagement/TelemetryConfiguration.cs
jimmyca15: Why would you like to drop the word events. This flag controls whether `EvaluationEvent` type is pushed to registered telemetry publishers. We already use event in what seems the most important piece of the puzzle.

jimmyca15: Do you think it would be better to instead go in the other direction and be more specific. Such as "enable or disable sending <cref="EvaluationEvent"/> events to registered <cref="ITelemetryPublisher"/>s.

README.md
jimmyca15: I like the suggestion.

README.md
jimmyca15: Updated.

README.md
jimmyca15: Updated.

src/Microsoft.FeatureManagement.Telemetry.ApplicationInsights/ApplicationInsightsTelemetryPublisher.cs
jimmyca15: Updated.

src/Microsoft.FeatureManagement/TelemetryConfiguration.cs
zhiyuanliang-ms: > Do you think it would be better to instead go in the other direction and be more specific. Such as "enable or disable sending <cref="EvaluationEvent"/> events to registered <cref="ITelemetryPublisher"/>s.

Sounds good. This is much better.

PR #347, opened 2024-01-05
Adds TargetingContext to EvaluationEvent & Emits TargetingId in the App Insights Publisher

rossgrambo:


zhiyuanliang-ms (approved):

src/Microsoft.FeatureManagement.Telemetry.ApplicationInsights/ApplicationInsightsTelemetryPublisher.cs
```diff
@@ -42,6 +42,11 @@ public ValueTask PublishEvent(EvaluationEvent evaluationEvent, CancellationToken
                 { "Enabled", evaluationEvent.Enabled.ToString() }
             };
 
+            if (evaluationEvent.TargetingContext != null)
+            {
+                properties["TargetingId"] = evaluationEvent.TargetingContext.UserId;
```
zhiyuanliang-ms, line 47: I am ok with the current implementation. Just curious about what will happen if the TargetingContext.UserId is null.
How will the experimentation link be established in this case? Will all metrics and events with a null "TargetingId" be linked together?

zhiyuanliang-ms:
@rossgrambo  We have not yet discussed how targeting id is set, right? In this PR, using TargetingContext.UserId is just a work-around solution, right?

src/Microsoft.FeatureManagement.Telemetry.ApplicationInsights/ApplicationInsightsTelemetryPublisher.cs
rossgrambo: That's a good question. The metrics will still be emitted, but linking them together from a null targetingId would not provide much value. Split will ignore the data for Experimentation if the id is null. The only value of the field is to explicitly indicate that there was no value, so it cannot be associated with metrics.

rossgrambo:
> @rossgrambo We have not yet discussed how targeting id is set, right? In this PR, using TargetingContext.UserId is just a work-around solution, right?

Another good question, no. This is the correct way to set Targeting Id for now. We could update TargetingContext in the future and essentially rename "UserId" to "TargetingId" but they are functionally the same.

src/Microsoft.FeatureManagement/Telemetry/EvaluationEvent.cs
```diff
@@ -13,6 +15,11 @@ public class EvaluationEvent
         /// </summary>
         public FeatureDefinition FeatureDefinition { get; set; }
 
+        /// <summary>
+        /// The targeting context used to evaluate the feature.
+        /// </summary>
+        public ITargetingContext TargetingContext { get; set; }
```
jimmyca15, line 21: I'd like to use `TargetingContext` instead of ITargetingContext, similarly to how `GetVariantAsync` takes `TargetingContext` rather than `ITargetingContext`.

https://github.com/microsoft/FeatureManagement-Dotnet/blob/2d7073a2006f478836290cfe4aae7b7bd920bbbd/src/Microsoft.FeatureManagement/IVariantFeatureManager.cs#L54

`ITargetingContext` exists to satisfy ContextualTargetingFilter.

rossgrambo: Good point, adjusted.

jimmyca15 (approved):

PR #348, opened 2024-01-09
Adds extension methods for TrackEvent and TrackMetric

rossgrambo:
Adds TargetingContext extension method for:
1. TrackEvent
2. TrackMetric

There are other track calls, but those calls are not a part of the happy path. `TargetingId` is only helpful in the happy path. Does not add extensions for:
1. TrackTrace
2. TrackException
3. TrackDependency
4. TrackAvailability

### DIFF
--- a/src/Microsoft.FeatureManagement.Telemetry.ApplicationInsights/ApplicationInsightsTelemetryExtensions.cs
+++ b/src/Microsoft.FeatureManagement.Telemetry.ApplicationInsights/ApplicationInsightsTelemetryExtensions.cs
@@ -1,4 +1,7 @@
-﻿using Microsoft.ApplicationInsights;
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+//
+using Microsoft.ApplicationInsights;
 using Microsoft.ApplicationInsights.DataContracts;
 using Microsoft.FeatureManagement.FeatureFilters;
 
@@ -10,7 +13,7 @@ namespace Microsoft.FeatureManagement.Telemetry.ApplicationInsights
     public static class ApplicationInsightsTelemetryExtensions
     {
         /// <summary>
-        /// Extension method to track an event with TargetingContext.
+        /// Extension method to track an event with <see cref="TargetingContext"/>.
         /// </summary>
         public static void TrackEvent(this TelemetryClient telemetryClient, string eventName, TargetingContext targetingContext, IDictionary<string, string> properties = null, IDictionary<string, double> metrics = null)
         {
@@ -19,13 +22,13 @@ namespace Microsoft.FeatureManagement.Telemetry.ApplicationInsights
                 properties = new Dictionary<string, string>();
             }
 
-            properties.Add("TargetingId", targetingContext.UserId);
+            properties["TargetingId"] = targetingContext.UserId;
 
             telemetryClient.TrackEvent(eventName, properties, metrics);
         }
 
         /// <summary>
-        /// Extension method to track an EventTelemetry with TargetingContext.
+        /// Extension method to track an <see cref="EventTelemetry"/> with <see cref="TargetingContext"/>.
         /// </summary>
         public static void TrackEvent(this TelemetryClient telemetryClient, EventTelemetry telemetry, TargetingContext targetingContext)
         {
@@ -40,7 +43,7 @@ namespace Microsoft.FeatureManagement.Telemetry.ApplicationInsights
         }
 
         /// <summary>
-        /// Extension method to track a metric with TargetingContext.
+        /// Extension method to track a metric with <see cref="TargetingContext"/>.
         /// </summary>
         public static void TrackMetric(this TelemetryClient telemetryClient, string name, double value, TargetingContext targetingContext, IDictionary<string, string> properties = null)
         {
@@ -49,13 +52,13 @@ namespace Microsoft.FeatureManagement.Telemetry.ApplicationInsights
                 properties = new Dictionary<string, string>();
             }
 
-            properties.Add("TargetingId", targetingContext.UserId);
+            properties["TargetingId"] = targetingContext.UserId;
 
             telemetryClient.TrackMetric(name, value, properties);
         }
 
         /// <summary>
-        /// Extension method to track a MetricTelemetry with TargetingContext.
+        /// Extension method to track a <see cref="MetricTelemetry"/> with <see cref="TargetingContext"/>.
         /// </summary>
         public static void TrackMetric(this TelemetryClient telemetryClient, MetricTelemetry telemetry, TargetingContext targetingContext)
         {
--- a/src/Microsoft.FeatureManagement.Telemetry.ApplicationInsights/ApplicationInsightsTelemetryExtensions.cs
+++ b/src/Microsoft.FeatureManagement.Telemetry.ApplicationInsights/ApplicationInsightsTelemetryExtensions.cs
@@ -1,0 +1,72 @@
+﻿using Microsoft.ApplicationInsights;
+using Microsoft.ApplicationInsights.DataContracts;
+using Microsoft.FeatureManagement.FeatureFilters;
+
+namespace Microsoft.FeatureManagement.Telemetry.ApplicationInsights
+{
+    /// <summary>
+    /// Provides extension methods for tracking events with TargetingContext.
+    /// </summary>
+    public static class ApplicationInsightsTelemetryExtensions
+    {
+        /// <summary>
+        /// Extension method to track an event with TargetingContext.
+        /// </summary>
+        public static void TrackEvent(this TelemetryClient telemetryClient, string eventName, TargetingContext targetingContext, IDictionary<string, string> properties = null, IDictionary<string, double> metrics = null)
+        {
+            if (properties == null)
+            {
+                properties = new Dictionary<string, string>();
+            }
+
+            properties.Add("TargetingId", targetingContext.UserId);
+
+            telemetryClient.TrackEvent(eventName, properties, metrics);
+        }
+
+        /// <summary>
+        /// Extension method to track an EventTelemetry with TargetingContext.
+        /// </summary>
+        public static void TrackEvent(this TelemetryClient telemetryClient, EventTelemetry telemetry, TargetingContext targetingContext)
+        {
+            if (telemetry == null)
+            {
+                telemetry = new EventTelemetry();
+            }
+
+            telemetry.Properties["TargetingId"] = targetingContext.UserId;
+
+            telemetryClient.TrackEvent(telemetry);
+        }
+
+        /// <summary>
+        /// Extension method to track a metric with TargetingContext.
+        /// </summary>
+        public static void TrackMetric(this TelemetryClient telemetryClient, string name, double value, TargetingContext targetingContext, IDictionary<string, string> properties = null)
+        {
+            if (properties == null)
+            {
+                properties = new Dictionary<string, string>();
+            }
+
+            properties.Add("TargetingId", targetingContext.UserId);
+
+            telemetryClient.TrackMetric(name, value, properties);
+        }
+
+        /// <summary>
+        /// Extension method to track a MetricTelemetry with TargetingContext.
+        /// </summary>
+        public static void TrackMetric(this TelemetryClient telemetryClient, MetricTelemetry telemetry, TargetingContext targetingContext)
+        {
+            if (telemetry == null)
+            {
+                telemetry = new MetricTelemetry();
+            }
+
+            telemetry.Properties["TargetingId"] = targetingContext.UserId;
+
+            telemetryClient.TrackMetric(telemetry);
+        }
+    }
+}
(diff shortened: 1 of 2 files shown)


src/Microsoft.FeatureManagement.Telemetry.ApplicationInsights/ApplicationInsightsTelemetryExtensions.cs
```diff
@@ -0,0 +1,72 @@
+﻿using Microsoft.ApplicationInsights;
```
zhenlan, line 1: Copyright header.

src/Microsoft.FeatureManagement.Telemetry.ApplicationInsights/ApplicationInsightsTelemetryExtensions.cs
```diff
@@ -0,0 +1,72 @@
...
+using Microsoft.ApplicationInsights.DataContracts;
+using Microsoft.FeatureManagement.FeatureFilters;
+
+namespace Microsoft.FeatureManagement.Telemetry.ApplicationInsights
+{
+    /// <summary>
+    /// Provides extension methods for tracking events with TargetingContext.
```
zhiyuanliang-ms, line 8:
```suggestion
    /// Provides extension methods for tracking events with <see cref="TargetingContext"/>.
```

zhiyuanliang-ms: As mentioned in https://github.com/microsoft/FeatureManagement-Dotnet/pull/347#discussion_r1444962105
Is there any reason to use `TargetingContext" instead of "ITargetingContext" here?

src/Microsoft.FeatureManagement.Telemetry.ApplicationInsights/ApplicationInsightsTelemetryExtensions.cs
```diff
@@ -0,0 +1,72 @@
...
+                properties = new Dictionary<string, string>();
+            }
+
+            properties.Add("TargetingId", targetingContext.UserId);
+
+            telemetryClient.TrackEvent(eventName, properties, metrics);
+        }
+
+        /// <summary>
+        /// Extension method to track an EventTelemetry with TargetingContext.
```
zhiyuanliang-ms, line 28:
```suggestion
        /// Extension method to track an <see cref="EventTelemetry"/> with TargetingContext.
```

src/Microsoft.FeatureManagement.Telemetry.ApplicationInsights/ApplicationInsightsTelemetryExtensions.cs
```diff
@@ -0,0 +1,72 @@
...
+                properties = new Dictionary<string, string>();
+            }
+
+            properties.Add("TargetingId", targetingContext.UserId);
+
+            telemetryClient.TrackMetric(name, value, properties);
+        }
+
+        /// <summary>
+        /// Extension method to track a MetricTelemetry with TargetingContext.
```
zhiyuanliang-ms, line 58:
```suggestion
        /// Extension method to track a <see cref="MetricTelemetry"/> with TargetingContext.
```

src/Microsoft.FeatureManagement.Telemetry.ApplicationInsights/ApplicationInsightsTelemetryExtensions.cs
```diff
@@ -0,0 +1,72 @@
...
+        /// Extension method to track an event with TargetingContext.
+        /// </summary>
+        public static void TrackEvent(this TelemetryClient telemetryClient, string eventName, TargetingContext targetingContext, IDictionary<string, string> properties = null, IDictionary<string, double> metrics = null)
+        {
+            if (properties == null)
+            {
+                properties = new Dictionary<string, string>();
+            }
+
+            properties.Add("TargetingId", targetingContext.UserId);
```
zhiyuanliang-ms, line 22: What will happen if the key "TargetingId" has already existed in the `properties` dictionary?

src/Microsoft.FeatureManagement.Telemetry.ApplicationInsights/ApplicationInsightsTelemetryExtensions.cs
jimmyca15: Contextual filters describe an interface that they handle. This allows a single app context type to satisfy multiple contextual feature filters.

rossgrambo: Don't take this as the actual answer, because I'm not sure what use case was in mind originally, but my understanding is that it enables something like:

```csharp
featureManager.IsEnabledAsync("Flag", new ContextRetriever());
```

```csharp
public class ContextRetriever: ITargetingContext
{
    string UserId { 
        get { ReadFromSomeLocalContext() } 
    }

    IEnumerable<string> Groups { 
        get { ReadFromSomeServer() }; 
    }
}
```

Which isn't needed in the Accessor Pattern, as the Accessor is essentially solving the same problem.

src/Microsoft.FeatureManagement.Telemetry.ApplicationInsights/ApplicationInsightsTelemetryExtensions.cs
```diff
@@ -0,0 +1,75 @@
...
+using Microsoft.ApplicationInsights;
+using Microsoft.ApplicationInsights.DataContracts;
+using Microsoft.FeatureManagement.FeatureFilters;
+
+namespace Microsoft.FeatureManagement.Telemetry.ApplicationInsights
+{
+    /// <summary>
+    /// Provides extension methods for tracking events with TargetingContext.
+    /// </summary>
+    public static class ApplicationInsightsTelemetryExtensions
```
jimmyca15, line 13:
```suggestion
    public static class TelemetryClientExtensions
```

src/Microsoft.FeatureManagement.Telemetry.ApplicationInsights/ApplicationInsightsTelemetryExtensions.cs
```diff
@@ -0,0 +1,75 @@
...
+        /// Extension method to track an <see cref="EventTelemetry"/> with <see cref="TargetingContext"/>.
+        /// </summary>
+        public static void TrackEvent(this TelemetryClient telemetryClient, EventTelemetry telemetry, TargetingContext targetingContext)
+        {
+            if (telemetry == null)
+            {
+                telemetry = new EventTelemetry();
+            }
+
+            telemetry.Properties["TargetingId"] = targetingContext.UserId;
```
jimmyca15, line 40: if `targetingContext` is null, this is a null-ref. I recommend we require it for this overload.

src/Microsoft.FeatureManagement.Telemetry.ApplicationInsights/ApplicationInsightsTelemetryExtensions.cs
```diff
@@ -0,0 +1,75 @@
...
+
+            telemetryClient.TrackEvent(telemetry);
+        }
+
+        /// <summary>
+        /// Extension method to track a metric with <see cref="TargetingContext"/>.
+        /// </summary>
+        public static void TrackMetric(this TelemetryClient telemetryClient, string name, double value, TargetingContext targetingContext, IDictionary<string, string> properties = null)
```
jimmyca15, line 48: At the moment, I believe only Event is in scope.

src/Microsoft.FeatureManagement.Telemetry.ApplicationInsights/ApplicationInsightsTelemetryExtensions.cs
```diff
@@ -0,0 +1,75 @@
...
+
+            telemetryClient.TrackEvent(eventName, properties, metrics);
+        }
+
+        /// <summary>
+        /// Extension method to track an <see cref="EventTelemetry"/> with <see cref="TargetingContext"/>.
+        /// </summary>
+        public static void TrackEvent(this TelemetryClient telemetryClient, EventTelemetry telemetry, TargetingContext targetingContext)
+        {
+            if (telemetry == null)
```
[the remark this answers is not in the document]
rossgrambo: Yes it does weirdly. Base method is:

```csharp
//
// Summary:
//     Send an Microsoft.ApplicationInsights.DataContracts.EventTelemetry for display
//     in Diagnostic Search and in the Analytics Portal. Create a separate Microsoft.ApplicationInsights.DataContracts.EventTelemetry
//     instance for each call to Microsoft.ApplicationInsights.TelemetryClient.TrackEvent(Microsoft.ApplicationInsights.DataContracts.EventTelemetry).
//
// Parameters:
//   telemetry:
//     An event log item.
//
// Remarks:
//     Learn more
public void TrackEvent(EventTelemetry telemetry)
{
    if (telemetry == null)
    {
        telemetry = new EventTelemetry();
    }

    Track(telemetry);
}
```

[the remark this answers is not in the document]
jimmyca15: I see. Very odd. I can't see any reason for it, and because of that I feel like we should deviate and block null EventTelemetry.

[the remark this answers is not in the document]
rossgrambo: I prefer to be consistent in the rare case someone is using it. It does allow for tracking events in the minimalist way possible: `telementryClient.trackEvent(null);`. And the resulting event will have everything an event has, except an eventName.

src/Microsoft.FeatureManagement.Telemetry.ApplicationInsights/ApplicationInsightsTelemetryExtensions.cs
rossgrambo: That is true for the whole system, but I think FeatureManagement should support this as well.

jimmyca15: There's two things that come to mind.

* we can always add we can never take away.
* Metrics are usually aggregated so I see complications when they are used in conjunction with targeting id.

src/Microsoft.FeatureManagement.Telemetry.ApplicationInsights/ApplicationInsightsTelemetryExtensions.cs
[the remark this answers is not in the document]
jimmyca15: Okay, sounds reasonable if an event makes it's way to app insights.

src/Microsoft.FeatureManagement.Telemetry.ApplicationInsights/ApplicationInsightsTelemetryExtensions.cs
rossgrambo: Valid points. Removed.

src/Microsoft.FeatureManagement.Telemetry.ApplicationInsights/TelemetryClientExtensions.cs
```diff
@@ -0,0 +1,62 @@
...
+        {
+            if (targetingContext == null)
+            {
+                throw new ArgumentNullException(nameof(targetingContext));
+            }
+        }
+
+        private static void AddTargetingProperties(IDictionary<string, string> properties, TargetingContext targetingContext)
+        {
+            properties["TargetingId"] = targetingContext.UserId;
```
jimmyca15, line 59: I push to avoid one-line private methods. There's not really a saving between a one line method call and just inlining the method implementation. But one thing that does happen is reviewing a method results in the need to jump to the private method.

jimmyca15 (approved):

src/Microsoft.FeatureManagement.Telemetry.ApplicationInsights/TelemetryClientExtensions.cs
```diff
@@ -0,0 +1,62 @@
...
+// Licensed under the MIT license.
+//
+using Microsoft.ApplicationInsights;
+using Microsoft.ApplicationInsights.DataContracts;
+using Microsoft.FeatureManagement.FeatureFilters;
+
+namespace Microsoft.FeatureManagement.Telemetry.ApplicationInsights
+{
+    /// <summary>
+    /// Provides extension methods for tracking events with TargetingContext.
```
jimmyca15, line 11:
```suggestion
    /// Provides extension methods for tracking events with <see cref="TargetingContext"/>.
```

src/Microsoft.FeatureManagement.Telemetry.ApplicationInsights/TelemetryClientExtensions.cs
rossgrambo: Fair point. To me, the one line had justtt too much that might change in the near future to repeat in 4 functions. 

But we ended up removing 2 of the functions (TrackMetric), so I'll remove this function as well for now. I'll still probably pull the string out though.

PR #350, opened 2024-01-12
Adds targeting middleware and targeting initializer

rossgrambo:
This PR introduces the idea of `TargetingId`. This is separate from UserId, although commonly they may be the same value. `TargetingId` is used to reflect which id was actually used for Feature Evaluation. This `TargetingId` can be tagged on events & metrics to connect events to evaluation events, emitted from a telemetry publisher. 

### `TargetingTelemetryInitializer` 

This was added in a new project, `Microsoft.FeatureManagement.Telemetry.ApplicationInsights.AspNetCore`. `TargetingTelemetryInitializer` is only relevant for Application Insights telemetry in ASP.NET Core.

### `TargetingHttpContextMiddleware`

This was added to `Microsoft.FeatureManagement.AspNetCore`. `TargetingHttpContextMiddleware` simply adds the TargetingId to HttpContext to enable Synchronous code to quickly access it. This is helpful for things like Telemetry Initializers. 

### Example

The example will be updated in a separate PR. To see how the example will change / how customers would use these new classes, see https://github.com/microsoft/FeatureManagement-Dotnet/compare/rossgrambo/targeting-id-example...rossgrambo/example-update-targeting-id?expand=1

### DIFF
--- a/src/Microsoft.FeatureManagement.AspNetCore/TargetingHttpContextMiddleware.cs
+++ b/src/Microsoft.FeatureManagement.AspNetCore/TargetingHttpContextMiddleware.cs
@@ -1,0 +1,54 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+//
+
+using Microsoft.AspNetCore.Http;
+using Microsoft.FeatureManagement.FeatureFilters;
+using System;
+using System.Threading.Tasks;
+
+namespace EvaluationDataToApplicationInsights.Telemetry
+{
+    /// <summary>
+    /// Used to add targeting information to http context. This allows synronous code to access targeting information.
+    /// </summary>
+    public class TargetingHttpContextMiddleware
+    {
+        private readonly RequestDelegate _next;
+
+        /// <summary>
+        /// Creates an instance of the TargetingHttpContextMiddleware
+        /// </summary>
+        public TargetingHttpContextMiddleware(RequestDelegate next) { 
+            _next = next; 
+        }
+
+        /// <summary>
+        /// Adds targeting information to the http context.
+        /// </summary>
+        /// <param name="context">The <see cref="HttpContext"/> to add the targeting information to.</param>
+        /// <param name="targetingContextAccessor">The <see cref="ITargetingContextAccessor"/> to retrieve the targeting information from.</param>
+        /// <exception cref="ArgumentNullException">Thrown if the provided context or targetingContextAccessor is null.</exception>
+        public async Task InvokeAsync(HttpContext context, ITargetingContextAccessor targetingContextAccessor)
+        {
+            if (context == null)
+            {
+                throw new ArgumentNullException(nameof(context));
+            }
+
+            if (targetingContextAccessor == null)
+            {
+                throw new ArgumentNullException(nameof(targetingContextAccessor));
+            }
+
+            TargetingContext targetingContext = await targetingContextAccessor.GetContextAsync().ConfigureAwait(false);
+
+            if (targetingContext != null)
+            {
+                context.Items["TargetingId"] = targetingContext.UserId;
+            }
+
+            await _next(context);
+        }
+    }
+}
--- a/src/Microsoft.FeatureManagement.AspNetCore/TargetingHttpContextMiddleware.cs
+++ b/src/Microsoft.FeatureManagement.AspNetCore/TargetingHttpContextMiddleware.cs
@@ -8,7 +8,7 @@ using Microsoft.FeatureManagement.FeatureFilters;
 using System;
 using System.Threading.Tasks;
 
-namespace EvaluationDataToApplicationInsights.Telemetry
+namespace Microsoft.FeatureManagement
 {
     /// <summary>
     /// Used to add targeting information to HTTP context.
@@ -24,7 +24,7 @@ namespace EvaluationDataToApplicationInsights.Telemetry
         /// Creates an instance of the TargetingHttpContextMiddleware
         /// </summary>
         public TargetingHttpContextMiddleware(RequestDelegate next, ILoggerFactory loggerFactory) {
-            _next = next;
+            _next = next ?? throw new ArgumentNullException(nameof(next));
             _logger = loggerFactory?.CreateLogger<TargetingHttpContextMiddleware>() ?? throw new ArgumentNullException(nameof(loggerFactory));
         }
 
--- a/src/Microsoft.FeatureManagement.AspNetCore/TargetingHttpContextMiddleware.cs
+++ b/src/Microsoft.FeatureManagement.AspNetCore/TargetingHttpContextMiddleware.cs
@@ -11,7 +11,7 @@ using System.Threading.Tasks;
 namespace EvaluationDataToApplicationInsights.Telemetry
 {
     /// <summary>
-    /// Used to add targeting information to http context. This allows synronous code to access targeting information.
+    /// Used to add targeting information to HTTP context.
     /// </summary>
     public class TargetingHttpContextMiddleware
     {
@@ -29,7 +29,7 @@ namespace EvaluationDataToApplicationInsights.Telemetry
         }
 
         /// <summary>
-        /// Adds targeting information to the http context.
+        /// Adds targeting information to the HTTP context.
         /// </summary>
         /// <param name="context">The <see cref="HttpContext"/> to add the targeting information to.</param>
         /// <param name="targetingContextAccessor">The <see cref="ITargetingContextAccessor"/> to retrieve the targeting information from.</param>
--- a/src/Microsoft.FeatureManagement.AspNetCore/TargetingHttpContextMiddleware.cs
+++ b/src/Microsoft.FeatureManagement.AspNetCore/TargetingHttpContextMiddleware.cs
@@ -3,6 +3,7 @@
 //
 
 using Microsoft.AspNetCore.Http;
+using Microsoft.Extensions.Logging;
 using Microsoft.FeatureManagement.FeatureFilters;
 using System;
 using System.Threading.Tasks;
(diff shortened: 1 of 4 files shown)


src/Microsoft.FeatureManagement.AspNetCore/TargetingHttpContextMiddleware.cs
```diff
@@ -0,0 +1,54 @@
...
+    /// Used to add targeting information to http context. This allows synronous code to access targeting information.
+    /// </summary>
+    public class TargetingHttpContextMiddleware
+    {
+        private readonly RequestDelegate _next;
+
+        /// <summary>
+        /// Creates an instance of the TargetingHttpContextMiddleware
+        /// </summary>
+        public TargetingHttpContextMiddleware(RequestDelegate next) {
```
zhiyuanliang-ms, line 22: I suggest we add a logger here.
```suggestion
        public TargetingHttpContextMiddleware(RequestDelegate next, , ILoggerFactory loggerFactory) { 
```
When targetingContext == null, we can log something for debug.

src/Microsoft.FeatureManagement.AspNetCore/TargetingHttpContextMiddleware.cs
```diff
@@ -0,0 +1,54 @@
...
+            if (targetingContextAccessor == null)
+            {
+                throw new ArgumentNullException(nameof(targetingContextAccessor));
+            }
+
+            TargetingContext targetingContext = await targetingContextAccessor.GetContextAsync().ConfigureAwait(false);
+
+            if (targetingContext != null)
+            {
+                context.Items["TargetingId"] = targetingContext.UserId;
```
zhiyuanliang-ms, line 48: It seems like the context.Items["TargetingId"] is expected to consume by the TargetingTelemetryInitializer. 
So maybe we should seperate the keyname to a constant string or even put it in a file e.g. TelemetryProperties.cs. "FeatureName", "Enabled", "TargetingId", "Variant" and "VariantAssignmentReason" should also be put in that file.

My another concern is that httpcontext is a shared place which can be modified by any custom middlewares. So the question is whether "TargetingId" is a concept which will only be used by FeatureManagement lib. What if users want to set their own "TargetingId" in HttpContext. This may not be a valid user case. But we have a precedent for feature filter settings cache which will be stored in the shared memory cache. What we did is to add a "FeatureManagement" prefix to the key name: https://github.com/microsoft/FeatureManagement-Dotnet/blob/preview/src/Microsoft.FeatureManagement/FeatureManager.cs#L666

src/Microsoft.FeatureManagement.Telemetry.ApplicationInsights.AspNetCore/TargetingTelemetryInitializer.cs
```diff
@@ -0,0 +1,62 @@
...
+            {
+                throw new ArgumentNullException("requestTelemetry");
+            }
+
+            if (httpContext == null)
+            {
+                throw new ArgumentNullException("httpContext");
+            }
+
+            string targetingId = (string) httpContext.Items["TargetingId"];
```
zhiyuanliang-ms, line 48: httpContext.Items["TargetingId"] is expected to be set as a string only by the TargetingHttpContextMiddleware. It seems like we cannot prevent other middleware from modifying it. I suggest we add some error handling here in case that httpContext.Items["TargetingId"] is not a string.

src/Microsoft.FeatureManagement.AspNetCore/TargetingHttpContextMiddleware.cs
rossgrambo: > It seems like the context.Items["TargetingId"] is expected to consume by the TargetingTelemetryInitializer.
So maybe we should seperate the keyname to a constant string or even put it in a file e.g. TelemetryProperties.cs. "FeatureName", "Enabled", "TargetingId", "Variant" and "VariantAssignmentReason" should also be put in that file.

This seemed like a good idea, so I started adding it. The issue is the publisher and initializer are in different projects. We could make TelemetryProperties but it would have it be public which is an interface change I doubt we want.

src/Microsoft.FeatureManagement.Telemetry.ApplicationInsights.AspNetCore/TargetingTelemetryInitializer.cs
rossgrambo: Great point. I went ahead and adjusted it to `?.ToString();`, which doesn't have error handling necessarily, but it'll either get a string result or null from whatever is placed in there. This seems right to me, what do you think?

src/Microsoft.FeatureManagement.AspNetCore/TargetingHttpContextMiddleware.cs
rossgrambo: > What we did is to add a "FeatureManagement" prefix to the key name

Went ahead and added `Microsoft.FeatureManagement.` as a prefix.

src/Microsoft.FeatureManagement.AspNetCore/TargetingHttpContextMiddleware.cs
```diff
@@ -0,0 +1,63 @@
...
+using Microsoft.AspNetCore.Http;
+using Microsoft.Extensions.Logging;
+using Microsoft.FeatureManagement.FeatureFilters;
+using System;
+using System.Threading.Tasks;
+
+namespace EvaluationDataToApplicationInsights.Telemetry
+{
+    /// <summary>
+    /// Used to add targeting information to http context. This allows synronous code to access targeting information.
```
jimmyca15, line 14:
```suggestion
    /// Used to add targeting information to HTTP context.
```

src/Microsoft.FeatureManagement.AspNetCore/TargetingHttpContextMiddleware.cs
```diff
@@ -0,0 +1,63 @@
...
+// Licensed under the MIT license.
+//
+
+using Microsoft.AspNetCore.Http;
+using Microsoft.Extensions.Logging;
+using Microsoft.FeatureManagement.FeatureFilters;
+using System;
+using System.Threading.Tasks;
+
+namespace EvaluationDataToApplicationInsights.Telemetry
```
jimmyca15, line 11: wrong namespace

src/Microsoft.FeatureManagement.AspNetCore/TargetingHttpContextMiddleware.cs
```diff
@@ -0,0 +1,63 @@
...
+        private readonly RequestDelegate _next;
+        private readonly ILogger _logger;
+
+        private const string TargetingIdKey = $"Microsoft.FeatureManagement.TargetingId";
+
+        /// <summary>
+        /// Creates an instance of the TargetingHttpContextMiddleware
+        /// </summary>
+        public TargetingHttpContextMiddleware(RequestDelegate next, ILoggerFactory loggerFactory) {
+            _next = next;
```
jimmyca15, line 27: check `next` for null

src/Microsoft.FeatureManagement.AspNetCore/TargetingHttpContextMiddleware.cs
```diff
@@ -0,0 +1,63 @@
...
+            _logger = loggerFactory?.CreateLogger<TargetingHttpContextMiddleware>() ?? throw new ArgumentNullException(nameof(loggerFactory));
+        }
+
+        /// <summary>
+        /// Adds targeting information to the http context.
```
jimmyca15, line 32:
```suggestion
        /// Adds targeting information to the HTTP context.
```

src/Microsoft.FeatureManagement.Telemetry.ApplicationInsights.AspNetCore/TargetingTelemetryInitializer.cs
```diff
@@ -0,0 +1,65 @@
...
+                throw new ArgumentNullException("requestTelemetry");
+            }
+
+            if (httpContext == null)
+            {
+                throw new ArgumentNullException("httpContext");
+            }
+
+            // Extract the targeting id from the http context
+            string targetingId = httpContext.Items[TargetingIdKey]?.ToString();
```
jimmyca15, line 51: Doesn't accessing a non-existent key throw a keynotfoundexception?

src/Microsoft.FeatureManagement.Telemetry.ApplicationInsights.AspNetCore/TargetingTelemetryInitializer.cs
```diff
@@ -0,0 +1,65 @@
...
+
+            if (!string.IsNullOrEmpty(targetingId))
+            {
+                requestTelemetry.Properties["TargetingId"] = targetingId;
+
+                // Telemetry.Properties is deprecated in favor of ISupportProperties
+                if (telemetry is ISupportProperties telemetryWithSupportProperties)
+                {
+                    telemetryWithSupportProperties.Properties["TargetingId"] = targetingId;
```
jimmyca15, line 60: This would set in two places. is that necessary?

src/Microsoft.FeatureManagement.AspNetCore/TargetingHttpContextMiddleware.cs
rossgrambo: Adjusted.

Didn't realize next would ever be null. Nearly all examples I see, the middleware completely assumes it wont be null, but looking into standard middlewares (like routing and authentication), I see they throw when it's null. Just surprised null handling isn't anywhere in the docs on writing middleware.

jimmyca15: Always check for null if null isn't accounted for. Regardless of anything else. In this case, the consumption of _next does not check for null so it should be validated in the constructor.

src/Microsoft.FeatureManagement.Telemetry.ApplicationInsights.AspNetCore/TargetingTelemetryInitializer.cs
rossgrambo: You're right, adjusted

src/Microsoft.FeatureManagement.Telemetry.ApplicationInsights.AspNetCore/TargetingTelemetryInitializer.cs
rossgrambo: So it was my understanding that this Interface was called differently, as it is given both a RequestTelemetry and a Telemetry. I didn't think it would be called at any point with those two items being the same, but I just did some testing, and I'm incorrect!

So good callout, I removed the requestTelemetry TargetingId setting, as it was redundant. I also removed the null check on requestTelemetry, since it being null no longer matters to this function.

rossgrambo:
Let me know if y'all see any other issues here.

zhiyuanliang-ms (approved):

PR #351, opened 2024-01-15
Add feature management prefix for cache key

zhiyuanliang-ms:
I found that the cache key for feature filter settings is different in preview branch. The cache key will have a prefix of "FeatureManagement". I think this is better.
https://github.com/microsoft/FeatureManagement-Dotnet/blob/preview/src/Microsoft.FeatureManagement/FeatureManager.cs#L666


jimmyca15 (approved):

jimmyca15:
Somehow the branches diverged?

PR #356, opened 2024-01-24
Support Microsoft Feature Management schema for preview branch

zhiyuanliang-ms:
Follow up #319 
Support [Microsoft Feature Flag schema v2.0.0](https://github.com/Azure/AppConfiguration/blob/main/docs/FeatureManagement/FeatureFlag.v2.0.0.schema.json)


src/Microsoft.FeatureManagement/MicrosoftFeatureFlagFields.cs
```diff
@@ -5,7 +5,7 @@
 namespace Microsoft.FeatureManagement
 {
     //
-    // Microsoft feature flag schema: https://github.com/Azure/AppConfiguration/blob/main/docs/FeatureManagement/FeatureFlag.v1.1.0.schema.json
+    // Microsoft feature flag schema: https://github.com/Azure/AppConfiguration/blob/main/docs/FeatureManagement/FeatureFlag.v2.0.0.schema.json
     internal static class MicrosoftFeatureFlagFields
     {
         public const string FeatureFlagsSectionName = "FeatureFlags";
```
rossgrambo, line 11: Lets add support for "feature_flags" as well here.

zhiyuanliang-ms: I think we should only support snake casing for Microsoft Feature Management schema. 
@jimmyca15

rossgrambo: The rest of this file is snake case, but perhaps I'm misunderstanding how this file is used?

PR #359, opened 2024-02-01
Update FilterCollectionExtensions

zhiyuanliang-ms:
## Why this PR?

Update `FilterCollectionExtensions`  #560

## Visible change

I really want to change the return type of this API to `FilterCollection`:

``` C#
public static IFilterMetadata AddForFeature<TFilterType>(this FilterCollection filters, string feature) where TFilterType : IAsyncActionFilter
{
    IFilterMetadata filterMetadata = null;

    filters.Add(new FeatureGatedAsyncActionFilter<TFilterType>(RequirementType.Any, false, feature));

    return filterMetadata;
}
```

Because, filterMetadata will always be returned as `null`. But as @rossgrambo mentioned, this is a breaking change. 

New APIs are added:

```C#
public static FilterCollection AddForFeature<TFilterType>(this FilterCollection filters, params string[] features)

public static FilterCollection AddForFeature<TFilterType>(this FilterCollection filters, RequirementType requirementType, params string[] features)

public static FilterCollection AddForFeature<TFilterType>(this FilterCollection filters, RequirementType requirementType, bool negate, params string[] features)
```

Usage:
```C#
services.AddMvcCore(o =>
{
    o.Filters.AddForFeature<MvcFilter1>("feature1")
                .AddForFeature<MvcFilter2>(RequirementType.All, "feature1", "feature2");
});
```

### DIFF
--- a/src/Microsoft.FeatureManagement.AspNetCore/FilterCollectionExtensions.cs
+++ b/src/Microsoft.FeatureManagement.AspNetCore/FilterCollectionExtensions.cs
@@ -16,12 +16,12 @@ namespace Microsoft.FeatureManagement
         /// <typeparam name="TFilterType">The MVC filter to add and use if the feature is enabled.</typeparam>
         /// <param name="filters">The filter collection to add to.</param>
         /// <param name="feature">The feature that will need to enabled to trigger the execution of the MVC filter.</param>
-        /// <returns></returns>
+        /// <returns>The reference to the added filter metadata.</returns>
         public static IFilterMetadata AddForFeature<TFilterType>(this FilterCollection filters, string feature) where TFilterType : IAsyncActionFilter
         {
-            IFilterMetadata filterMetadata = null;
+            IFilterMetadata filterMetadata = new FeatureGatedAsyncActionFilter<TFilterType>(RequirementType.Any, false, feature);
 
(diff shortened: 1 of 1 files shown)


rossgrambo:
This is a "breaking" change so I'm not sure if we can do it. The reason "breaking" is in quotes is because it looks like we only ever return `null` here... But in the rare case where someone has:

```csharp
services.AddMvc(o => 
{
    IFilterMetadata metadata = o.Filters.AddForFeature<ThirdPartyActionFilter>("MyFeatureFlag");
});
```

We would break them.

zhiyuanliang-ms:
@jimmyca15 How about make this PR targeting on preview branch to avoid making breaking changes on our current stable release?

zhiyuanliang-ms:
> This is a "breaking" change so I'm not sure if we can do it. The reason "breaking" is in quotes is because it looks like we only ever return `null` here... But in the rare case where someone has:
> 
> ```cs
> services.AddMvc(o => 
> {
>     IFilterMetadata metadata = o.Filters.AddForFeature<ThirdPartyActionFilter>("MyFeatureFlag");
> });
> ```
> 
> We would break them.

`AddForFeature` always return `null`. This is a bug.

jimmyca15 (approved):

Issue #358, opened 2024-01-25
Should we have a feature flag schema specification?


zhiyuanliang-ms:
Something like this: https://github.com/Azure/AppConfiguration/blob/main/docs/FeatureManagement/FeatureFlag.v2.0.0.schema.json

rossgrambo:
I think you just linked to it- meaning we do have a feature flag specification 😄. It's just not in this repo- which I think makes sense because it's for more languages than .net.

jimmyca15:
Yes we should.

PR #362, opened 2024-02-02
Add feature management schema for main branch

zhiyuanliang-ms:
#358 

Before release 3.2.0. We need to specify the schema we supported. The schemas specify what the "FeatureManagement" section should look like, which is actually a missing piece of this repo's documentation.

This PR is associated with this PR: https://github.com/Azure/AppConfiguration/pull/871. The Microsoft Feature Management configuration schema mentioned in the README is placed in the Azure App Config repo.

The Json schema (file) names need to be discussed.

I found this website: https://www.jsonschemavalidator.net/ which can validate the json schema.


schemas/FeatureManagement.Dotnet.v1.0.0.schema.json
```diff
@@ -0,0 +1,91 @@
...
+  "patternProperties": {
+    "^[^:]*$": {
+      "description": "Declares a feature flag.",
+      "oneOf": [
+        {
+          "type": "boolean"
+        },
+        {
+          "type": "object",
+          "required": [
```
amerjusupovic, line 16: I was looking at `ConfigurationFeatureDefinitionProvider` and I saw [this comment](https://github.com/microsoft/FeatureManagement-Dotnet/blob/d19c9ad045968099cca87cea024ff08ad3744307/src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs#L207) about how we support strings as the value. If so, then maybe that needs to be included in the `oneOf` section. If it isn't supported maybe we could remove those couple lines of the comment?

README.md
```diff
@@ -144,6 +146,12 @@ A `RequirementType` of `All` changes the traversal. First, if there are no filte
 
 In the above example, `FeatureW` specifies a `RequirementType` of `All`, meaning all of its filters must evaluate to true for the feature to be enabled. In this case, the feature will be enabled for 50% of users during the specified time window.
 
+#### Microsoft Feature Flag Schema
+
+The feature management library also supports the usage of [`Microsoft Feature Flag schema``](https://github.com/Azure/AppConfiguration/blob/main/docs/FeatureManagement/FeatureFlag.v1.1.0.schema.json) to declare feature flags.
```
amerjusupovic, line 151:
```suggestion
The feature management library also supports the usage of [`Microsoft Feature Flag schema`](https://github.com/Azure/AppConfiguration/blob/main/docs/FeatureManagement/FeatureFlag.v1.1.0.schema.json) to declare feature flags.

```

schemas/FeatureManagement.Dotnet.v1.0.0.schema.json
zhiyuanliang-ms:
```
myDisabledFeature3: "",
```
I think we don't really like users to declare a feature flag in this way, but under our current implementation, even if the user writes it like this, we will parse this feature flag as always disabled. 

> If it isn't supported maybe we could remove those couple lines of the comment?

I agree. I will make some updates to the comment.

zhiyuanliang-ms:
@jimmyca15 

https://github.com/microsoft/FeatureManagement-Dotnet/blob/2.6.1/src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs#L115

/*
              
We support

myFeature: {
  enabledFor: [ "myFeatureFilter1", "myFeatureFilter2" ]
},
myDisabledFeature: {
  enabledFor: [  ]
},
myFeature2: {
  enabledFor: "myFeatureFilter1;myFeatureFilter2"
},
myDisabledFeature2: {
  enabledFor: ""
},
myFeature3: "myFeatureFilter1;myFeatureFilter2",
myDisabledFeature3: "",
myAlwaysEnabledFeature: true,
myAlwaysDisabledFeature: false // removing this line would be the same as setting it to false
myAlwaysEnabledFeature2: {
  enabledFor: true
},
myAlwaysDisabledFeature2: {
  enabledFor: false
},
myAllRequiredFilterFeature: {
    requirementType: "all"
    enabledFor: [ "myFeatureFilter1", "myFeatureFilter2" ],
},
*/

These are all valid/invalid usages we mentioned in the comment in `ConfigurationFeatureDefinitionProvider.cs`. 

First, I've tested these usages (which was told as valid in the comment, but actually invalid):
1. myFeature3: "myFeatureFilter1;myFeatureFilter2",
2. myFeature2: {
  enabledFor: "myFeatureFilter1;myFeatureFilter2"
}
They cannot be recognized by our current library. I think we should remove them in the comments.

The following usages can be supported by the current library, but we only mentioned it in the comments not in the README or anywhere else:
myAlwaysEnabledFeature2: {
  enabledFor: true
},
myAlwaysDisabledFeature2: {
  enabledFor: false
},

I will add this usage in the schema.

zhiyuanliang-ms:
@jimmyca15 The schema names need to be decided. Do you have any idea about the name?

schemas/FeatureManagement.Dotnet.v1.0.0.schema.json
```diff
@@ -0,0 +1,98 @@
+{
+  "definitions": {},
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "type": "object",
+  "title": ".NET Feature Management Configuration",
```
jimmyca15, line 5:
```suggestion
  "title": ".NET Feature Flag Definition",
```

schemas/FeatureManagement.Dotnet.v2.0.0.schema.json
```diff
@@ -0,0 +1,19 @@
+{
```
jimmyca15, line 1: I wouldn't consider this v2.0.0 of anything. This is separate

jimmyca15: I would say that this should be defined in the Azure/AppConfiguration repo where the Microsoft FeatureFlag schema is defined, and it should contain the root `FeatureManagement` node.

README.md
```diff
@@ -144,6 +146,12 @@ A `RequirementType` of `All` changes the traversal. First, if there are no filte
 
 In the above example, `FeatureW` specifies a `RequirementType` of `All`, meaning all of its filters must evaluate to true for the feature to be enabled. In this case, the feature will be enabled for 50% of users during the specified time window.
 
+#### Microsoft Feature Flag Schema
+
+The feature management library also supports the usage of [`Microsoft Feature Flag schema`](https://github.com/Azure/AppConfiguration/blob/main/docs/FeatureManagement/FeatureFlag.v1.1.0.schema.json) to declare feature flags.
+
+If you want to use the `Microsoft Feature Flag schema`, you will need to follow [this schema](./schemas/FeatureManagement.Dotnet.v2.0.0.schema.json) to set up the `FeatureManagement` section of the configuration.
+
```
jimmyca15, line 154: Perhaps an example would be helpful?

schemas/FeatureManagement.Dotnet.v2.0.0.schema.json
zhiyuanliang-ms: PR sent out: https://github.com/Azure/AppConfiguration/pull/871

README.md
```diff
@@ -144,6 +146,51 @@ A `RequirementType` of `All` changes the traversal. First, if there are no filte
 
 In the above example, `FeatureW` specifies a `RequirementType` of `All`, meaning all of its filters must evaluate to true for the feature to be enabled. In this case, the feature will be enabled for 50% of users during the specified time window.
 
+#### Microsoft Feature Flag Schema
+
+The feature management library also supports the usage of [`Microsoft Feature Flag schema`](https://github.com/Azure/AppConfiguration/blob/main/docs/FeatureManagement/FeatureFlag.v1.1.0.schema.json) to declare feature flags.
+
+If you want to use the `Microsoft Feature Flag schema`, you will need to follow the Azure App Configuration Feature Management Configuration schema to set up feature flags. The schema can be found [here](https://github.com/Azure/AppConfiguration/tree/main/docs/FeatureManagement). The feature flag declarations should be listed in the `FeatureFlags` array under the `FeatureManagement` section, as shown in the example below:
+
+```json
```
jimmyca15, line 155: I would reduce the complexity of this to a single feature flag. with a single filter. I think the biggest gain of the example is the fact that there is a `FeatureFlags` property and it is an `array`. Other than that we can include that single flag to demonstrate that the flag indeed uses the microsoft feature flag schema rather than the .net feature management schema.

README.md
```diff
@@ -144,6 +146,51 @@ A `RequirementType` of `All` changes the traversal. First, if there are no filte
...
 
+#### Microsoft Feature Flag Schema
+
+The feature management library also supports the usage of [`Microsoft Feature Flag schema`](https://github.com/Azure/AppConfiguration/blob/main/docs/FeatureManagement/FeatureFlag.v1.1.0.schema.json) to declare feature flags.
+
+If you want to use the `Microsoft Feature Flag schema`, you will need to follow the Azure App Configuration Feature Management Configuration schema to set up feature flags. The schema can be found [here](https://github.com/Azure/AppConfiguration/tree/main/docs/FeatureManagement). The feature flag declarations should be listed in the `FeatureFlags` array under the `FeatureManagement` section, as shown in the example below:
+
+```json
+{
+    "FeatureManagement": {
```
jimmyca15, line 157: Looking at this I'm thinking, it's kind of odd that `FeatureManagement` and `FeatureFlags` have a separate naming convention than the property names of individual flags. Especially given that this is the `Microsoft Feature Management Schema` & the `Microsoft Feature Flag Schema`.

Seems it should be `feature_management` and `feature_flags`.

jimmyca15: @rossgrambo

README.md
```diff
@@ -54,7 +54,7 @@ The .NET Core configuration system is used to determine the state of feature fla
 
 The feature management library supports appsettings.json as a feature flag source since it is a provider for .NET Core's IConfiguration system. Below we have an example of the format used to set up feature flags in a json file.
 
-``` JavaScript
+``` json
```
rossgrambo, line 57:
```suggestion
``` JavaScript
```
This should stay JavaScript. Its a trick to allow "//" comments without github highlighting it in red. (See rich diff)

README.md
```diff
@@ -91,12 +91,14 @@ The feature management library supports appsettings.json as a feature flag sourc
...
 
+The detailed schema of the `FeatureManagement` section can be found [here](./schemas/FeatureManagement.Dotnet.v1.0.0.schema.json).
+
 **Advanced:** The usage of colon ':' in feature flag names is forbidden.
 
 #### On/Off Declaration
  
 The following snippet demonstrates an alternative way to define a feature that can be used for on/off features. 
-``` JavaScript
+``` Json
```
rossgrambo, line 101: Same as above

README.md
zhiyuanliang-ms: Oh, thanks. I never noticed this difference.

README.md
```diff
@@ -144,6 +146,51 @@ A `RequirementType` of `All` changes the traversal. First, if there are no filte
 
 In the above example, `FeatureW` specifies a `RequirementType` of `All`, meaning all of its filters must evaluate to true for the feature to be enabled. In this case, the feature will be enabled for 50% of users during the specified time window.
 
+#### Microsoft Feature Flag Schema
+
+The feature management library also supports the usage of [`Microsoft Feature Flag schema`](https://github.com/Azure/AppConfiguration/blob/main/docs/FeatureManagement/FeatureFlag.v1.1.0.schema.json) to declare feature flags.
+
+If you want to use the `Microsoft Feature Flag schema`, you will need to follow the Azure App Configuration Feature Management Configuration schema to set up feature flags. The schema can be found [here](https://github.com/Azure/AppConfiguration/tree/main/docs/FeatureManagement). The feature flag declarations should be listed in the `FeatureFlags` array under the `FeatureManagement` section, as shown in the example below:
```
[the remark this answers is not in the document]
zhiyuanliang-ms: Also updated in this [PR](https://github.com/Azure/AppConfiguration/pull/871)

README.md
zhiyuanliang-ms: @jimmyca15 @rossgrambo I will take this. The json schema and the code base will both need to change.

zhiyuanliang-ms: @jimmyca15 PR to let our code base to recognize snake_casing: #370

schemas/FeatureManagement.Dotnet.v1.0.0.schema.json
```diff
@@ -0,0 +1,102 @@
+{
+  "definitions": {},
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "type": "object",
+  "title": "A.NET Feature Management Configuration",
```
rossgrambo, line 5:
```suggestion
  "title": "A .NET Feature Management Configuration",
```

schemas/FeatureManagement.Dotnet.v1.0.0.schema.json
```diff
@@ -0,0 +1,102 @@
...
+          "description": "A feature flag which value is dynamic based on a set of feature filters",
+          "required": [
+            "EnabledFor"
+          ],
+          "properties": {
+            "RequirementType": {
+              "type": "string",
+              "title": "Requirement Type",
+              "description": "Determines whether any or all registered feature filters must be enabled for the feature to be considered enabled.",
+              "enum": [
```
rossgrambo, line 28: This is not ideal because enums enforce case- and this value is case insensitive. We could just make it a string with any value (and supply "example"s of All and Any). Or use regex- but that workaround isn't great. 

Alternatively- we leave it like this or change the enums to lower case- then update our examples & tests to adhere to the case (currently they dont)

rossgrambo (approved):

README.md
```diff
@@ -144,6 +146,38 @@ A `RequirementType` of `All` changes the traversal. First, if there are no filte
 
 In the above example, `FeatureW` specifies a `RequirementType` of `All`, meaning all of its filters must evaluate to true for the feature to be enabled. In this case, the feature will be enabled for 50% of users during the specified time window.
 
+#### Microsoft Feature Flag Schema
+
+The feature management library also supports the usage of [`Microsoft Feature Flag schema`](https://github.com/Azure/AppConfiguration/blob/main/docs/FeatureManagement/FeatureFlag.v1.1.0.schema.json) to declare feature flags.
```
jimmyca15, line 151:
```suggestion
The feature management library also supports the usage of the [`Microsoft Feature Management schema`](https://github.com/Azure/AppConfiguration/blob/main/docs/FeatureManagement/FeatureFlag.v1.1.0.schema.json) to declare feature flags.

```

README.md
```diff
@@ -144,6 +146,38 @@ A `RequirementType` of `All` changes the traversal. First, if there are no filte
 
 In the above example, `FeatureW` specifies a `RequirementType` of `All`, meaning all of its filters must evaluate to true for the feature to be enabled. In this case, the feature will be enabled for 50% of users during the specified time window.
 
+#### Microsoft Feature Flag Schema
+
+The feature management library also supports the usage of [`Microsoft Feature Flag schema`](https://github.com/Azure/AppConfiguration/blob/main/docs/FeatureManagement/FeatureFlag.v1.1.0.schema.json) to declare feature flags.
+
+If you want to use the `Microsoft Feature Flag schema`, you will need to follow the `Microsoft Feature Management Configuration schema` to set up feature flags. The schema can be found [here](https://github.com/Azure/AppConfiguration/tree/main/docs/FeatureManagement). The feature flag declarations should be listed in the `feature_flags` array under the `feature_management` section, as shown in the example below:
```
jimmyca15, line 153: > If you want to use the `Microsoft Feature Flag schema`, you will need to follow the `Microsoft Feature Management Configuration schema`

I feel like we should only talk about the Microsoft Feature Management schema. The flag schema simply falls in line as a result.

README.md
```diff
@@ -144,6 +146,38 @@ A `RequirementType` of `All` changes the traversal. First, if there are no filte
...
+                        }
+                    ]
+                }
+            }
+        ]
+    }
+}
+```
+
+**Note:** If `feature_management` section can be found in the configuration, the `FeatureManagement` section will be ignored.
```
jimmyca15, line 179:
```suggestion
**Note:** If the `feature_management` section can be found in the configuration, the `FeatureManagement` section will be ignored.

```

schemas/FeatureManagement.Dotnet.v1.0.0.schema.json
jimmyca15, line 5:
```suggestion
  "title": "The .NET Feature Management Schema",
```

schemas/FeatureManagement.Dotnet.v1.0.0.schema.json
jimmyca15: We should declare the schema the proper way.

schemas/FeatureManagement.Dotnet.v1.0.0.schema.json
```diff
@@ -0,0 +1,102 @@
...
+                "All"
+              ],
+              "default": "Any"
+            },
+            "EnabledFor": {
+              "oneOf": [
+                {
+                  "type": "array",
+                  "title": "Feature Filter Collection",
+                  "description": "Feature Filters that must be evaluated as true for the feature to be considered enabled.",
```
jimmyca15, line 39: This wording is a listtle sensitive, sense for default requirement type only a single filter needs to be true. I would say instead

```suggestion
                  "description": "Feature filters that are evaluated to conditionally enable the flag.",
```

README.md
jimmyca15: I also feel it's worth explaining why this exists since it may be odd to see it declared immediately after the .NET schema.

This schema is language agnostic in origin and is supported by all Microsoft feature management libraries.

schemas/FeatureManagement.Dotnet.v1.0.0.schema.json
rossgrambo: > We should declare the schema the proper way, with specific enum value.

We should update our tests and examples to adhere to it then.

jimmyca15: Completely agree. What ones did you find?

rossgrambo: FeatureUsesFiltersWithDuplicatedAlias and AllFeatureFilter.

All valid now! Updated in https://github.com/microsoft/FeatureManagement-Dotnet/pull/372

Issue #364, opened 2024-02-05
Test suit should also target on .NET8.0


zhiyuanliang-ms:
Since .NET 8.0 is the latest LTS, we should make our test suite target on it too.

PR #365, opened 2024-02-05
Target on .NET 8.0

zhiyuanliang-ms:
#364


tests/Tests.FeatureManagement.AspNetCore/Tests.FeatureManagement.AspNetCore.csproj
```diff
@@ -30,6 +30,11 @@
     <PackageReference Include="Microsoft.AspNetCore.TestHost" Version="7.0.11" />
   </ItemGroup>
 
+  <ItemGroup Condition="'$(TargetFramework)' == 'net8.0'">
+    <PackageReference Include="Microsoft.Extensions.Configuration.Json" Version="7.0.0" />
+    <PackageReference Include="Microsoft.AspNetCore.TestHost" Version="7.0.11" />
```
rossgrambo, line 35: Should probably do the 8.- packages here.

examples/ConsoleApp/ConsoleApp-Charles.csproj
```diff
@@ -0,0 +1,25 @@
+<Project Sdk="Microsoft.NET.Sdk">
```
rossgrambo, line 1: What does this .csproj do? Was it included by accident?

zhiyuanliang-ms: Oh, I mis-uploaded my test application. I will remove it.

tests/Tests.FeatureManagement.AspNetCore/Tests.FeatureManagement.AspNetCore.csproj
zhiyuanliang-ms: I follow the previous one.
https://github.com/microsoft/FeatureManagement-Dotnet/blob/main/tests/Tests.FeatureManagement/Tests.FeatureManagement.csproj#L28

Should I also change them?

rossgrambo: I think yes, if it doesn't cause any issues. These are the test packages so they don't matter for customers, so we might as well be on latest.

zhiyuanliang-ms: I think we are good to use "net8.0", this is the correct TFM according to [doc](https://learn.microsoft.com/en-us/dotnet/standard/frameworks#how-to-specify-a-target-framework).

rossgrambo: Sorry, my comment might have been misleading. I meant we should do 

```
<ItemGroup Condition="'$(TargetFramework)' == 'net8.0'">
    <PackageReference Include="Microsoft.Extensions.Configuration.Json" Version="8.0.0" />
    <PackageReference Include="Microsoft.AspNetCore.TestHost" Version="8.0.1" />
</ItemGroup>
```

I didn't know what the latest sub-versions of the TestHost and Json packages were so I only wrote the "8.-" part. 😅

zhiyuanliang-ms: The latest version of [Microsoft.Extensions.Configuration.Json](https://www.nuget.org/packages/Microsoft.Extensions.Configuration.Json) package is 8.0.0. It seems like there is no minor/patch version bump since 3.1.32. It is quite stable. So It is good for us to use 7.0.0/6.0.0 for .NET 7/6.

The latest version of [Microsoft.AspNetCore.TestHost](https://www.nuget.org/packages/Microsoft.AspNetCore.TestHost) is 8.0.1 and I will also update the version for NET 6 and NET 7.

rossgrambo (approved):

jimmyca15 (approved):

PR #370, opened 2024-02-18
Support Microsoft Feature Management schema for main branch

zhiyuanliang-ms:
Adjust Microsoft Feature Management schema to use snake casing: "feature_management" and "feature_flags" so that the fields name will be more consistent.

As discussed in the email, we will use snake_casing and not support FeatureManagement.FeatureFlags at all. When people have both "feature_management" and "FeatureManagement", "FeatureManagement" will be ignored.

We are going to support
```json
{
  "feature_management": {
    "feature_flags": [
      {
        // Microsoft Feature Flag schema
      }
    ]
  }
}

```
#319 will be deprecated.

Related PRs: #362 
Microsoft Feature Management schema: https://github.com/Azure/AppConfiguration/pull/871


src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
```diff
@@ -134,21 +134,38 @@ private void EnsureInit()
...
+                }
 
                 if (featureManagementConfigurationSection == null && RootConfigurationFallbackEnabled)
                 {
                     featureManagementConfigurationSection = _configuration;
-                }
 
-                bool hasMicrosoftFeatureFlagSchema = featureManagementConfigurationSection != null && 
-                    HasMicrosoftFeatureFlagSchema(featureManagementConfigurationSection);
+                    hasMicrosoftFeatureFlagSchema = featureManagementConfigurationSection != null &&
```
rossgrambo, line 166: Do we need to fallback for Microsoft Feature Flag schema? I don't think we would ever have it without the `feature_management` wrappers. But I suppose we could support customers who want to do that? I prefer we don't support it.

src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
```diff
@@ -134,21 +134,38 @@ private void EnsureInit()
...
                         string.Equals(
                             section.Key,
                             ConfigurationFields.FeatureManagementSectionName,
                             StringComparison.OrdinalIgnoreCase));
+                }
 
                 if (featureManagementConfigurationSection == null && RootConfigurationFallbackEnabled)
                 {
                     featureManagementConfigurationSection = _configuration;
-                }
```
jimmyca15, line 165: The check `featureManagementConfigurationSection != null` seems unnecessary since `_configuration` is unable to be null as validated in the constructor.

jimmyca15: Seems it should be 

```
if (featureManagementConfigurationSection == null && RootConfigurationFallbackEnabled)
{
    hasMicrosoftFeatureFlagSchema = HasMicrosoftFeatureFlagSchema(_configuration);
}
```

jimmyca15 (approved):

src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
zhiyuanliang-ms: @jimmyca15 
I am kind of convinced by Ross. I agree that we might never have it without the "feature_management" wrappers.

In the Microsoft Feature Management schema, "feature_management" is a required property. So I think we should not support fallback for it.

Previously, "FeatureFlags" array was under "FeatureManagement" section. With the complexity brought by the RootConfigurationFallBack behavior, we have to use the "EnsureInit" pattern. Now, the Microsoft Feature Management schema is under "feature_management", we can tell which schema is being used in the constructor. We can simplify the code if we don't support the RootConfigurationFallBack for Microsoft Feature Management schema.

src/Microsoft.FeatureManagement/MicrosoftFeatureManagementFields.cs
```diff
@@ -4,14 +4,13 @@
 
 namespace Microsoft.FeatureManagement
 {
-    //
-    // Microsoft feature flag schema: https://github.com/Azure/AppConfiguration/blob/main/docs/FeatureManagement/FeatureFlag.v1.1.0.schema.json
-    internal static class MicrosoftFeatureFlagFields
+    internal static class MicrosoftFeatureManagementFields
     {
```
jimmyca15, line 8: Why remove the schema link?

jimmyca15 (approved):

src/Microsoft.FeatureManagement/MicrosoftFeatureManagementFields.cs
zhiyuanliang-ms: Re-added. Now the link is Microsoft Feature Management schema instead of Feature Flag schema.

Issue #378, opened 2024-02-26
Improve non-DI usage of built-in feature filters


zhiyuanliang-ms:
After Microsoft.FeatureManagement 3.1.0, we exposed `FeatureManager` and `ConfigurationFeatureDefintionProvider` classes to public and the non-DI usage of Feature Management is supported:
```C#
IConfiguration configuration = new ConfigurationBuilder()
    .AddJsonFile("appsettings.json")
    .Build();

var featureDefintionProvider = new ConfigurationFeatureDefinitionProvider(configuration);

var featureManager = new FeatureManager(featureDefintionProvider)();

bool res = await featureManager.IsEnabledAsync("MyFeature");
```

However, the built-in feature filters are designed for DI usage originally. If users want to use feature filters in non-DI Feature Management. The usage will be like:

``` C#
IConfiguration configuration = new ConfigurationBuilder()
    .AddJsonFile("appsettings.json")
    .Build();

var featureDefintionProvider = new ConfigurationFeatureDefinitionProvider(configuration);

ILoggerFactory loggerFactory = new LoggerFactory();

var targetingFilter = new ContextualTargetingFilter(Options.Create(new TargetingEvaluationOptions()), loggerFactory);

var timeWindowFilter = new TimeWindowFilter(loggerFactory);

var featureManager = new FeatureManager(featureDefintionProvider)
{
    FeatureFilters = new List<IFeatureFilterMetadata>() { targetingFilter, timeWindowFilter },
    Logger = loggerFactory.CreateLogger<FeatureManager>()
};

bool res = await featureManager.IsEnabledAsync("MyFeature");
```

The current non-DI usage looks kind of messy.

Notice that the current constructor of built-in filters are:
``` C#
public PercentageFilter(ILoggerFactory loggerFactory)
{
    _logger = loggerFactory.CreateLogger<PercentageFilter>();
}

public TimeWindowFilter(ILoggerFactory loggerFactory)
{
    _logger = loggerFactory?.CreateLogger<TimeWindowFilter>() ?? throw new ArgumentNullException(nameof(loggerFactory));
}

public ContextualTargetingFilter(IOptions<TargetingEvaluationOptions> options, ILoggerFactory loggerFactory)
{
    _options = options?.Value ?? throw new ArgumentNullException(nameof(options));
    _logger = loggerFactory?.CreateLogger<ContextualTargetingFilter>() ?? throw new ArgumentNullException(nameof(loggerFactory));
}

[...]

PR #386, opened 2024-03-07
Update constructors of built-in feature filters

zhiyuanliang-ms:
#378

### DIFF
--- a/src/Microsoft.FeatureManagement/FeatureFilters/PercentageFilter.cs
+++ b/src/Microsoft.FeatureManagement/FeatureFilters/PercentageFilter.cs
@@ -21,9 +21,9 @@ namespace Microsoft.FeatureManagement.FeatureFilters
         /// Creates a percentage based feature filter.
         /// </summary>
         /// <param name="loggerFactory">A logger factory for creating loggers.</param>
-        public PercentageFilter(ILoggerFactory loggerFactory)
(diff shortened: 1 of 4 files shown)


src/Microsoft.FeatureManagement/FeatureFilters/PercentageFilter.cs
```diff
@@ -21,9 +21,9 @@ public class PercentageFilter : IFeatureFilter, IFilterParametersBinder
         /// Creates a percentage based feature filter.
         /// </summary>
         /// <param name="loggerFactory">A logger factory for creating loggers.</param>
-        public PercentageFilter(ILoggerFactory loggerFactory)
+        public PercentageFilter(ILoggerFactory loggerFactory = null)
         {
-            _logger = loggerFactory.CreateLogger<PercentageFilter>();
+            _logger = loggerFactory?.CreateLogger<PercentageFilter>();
```
Eskibear, line 26: Prior to this PR, before we have a default `null` value for loggerFactory, am I allowed to explicitly pass a null as parameter, i.e. `new PercentageFilter(null)`? If so, this PR also fixes the potential NPE, right?

zhiyuanliang-ms: Yes.

Eskibear (approved):

zhenlan (approved):

PR #388, opened 2024-03-08
Update README to mention the usage of feature-based injection (variant service)

zhiyuanliang-ms:


README.md
```diff
@@ -970,6 +972,52 @@ If you are using a feature flag with binary variants, the `StatusOverride` prope
 
 In the above example, the feature is enabled by the `AlwaysOn` filter. If the current user is in the calculated percentile range of 10 to 20, then the `On` variant is returned. Otherwise, the `Off` variant is returned and because `StatusOverride` is equal to `Disabled`, the feature will now be considered disabled.
 
+### Variant Service
```
jimmyca15, line 975: I would highlight Dependency Injection here.

Something like

## Use variant services in dependency injection

jimmyca15: Also, probably makes sense to be in the ToC at the top of the readme.

README.md
```diff
@@ -970,6 +972,52 @@ If you are using a feature flag with binary variants, the `StatusOverride` prope
 
 In the above example, the feature is enabled by the `AlwaysOn` filter. If the current user is in the calculated percentile range of 10 to 20, then the `On` variant is returned. Otherwise, the `Off` variant is returned and because `StatusOverride` is equal to `Disabled`, the feature will now be considered disabled.
 
+### Variant Service
+
+Dependency injection can be wired up with the variant feature flag. Different implementations of a service interface can be considered as variant services. Specific variant of an interface can be retrieved from the dependency injection container based on the allocated variant of a variant feature flag. This could be done by using `IVariantServiceProvider<TService>`.
```
rossgrambo, line 977:
```suggestion
Dependency injection can be wired up with a variant feature flag. Different implementations of a service interface can be considered as variant services. Specific variant of an interface can be retrieved from the dependency injection container based on the allocated variant of a variant feature flag. This could be done by using `IVariantServiceProvider<TService>`.

```

README.md
```diff
@@ -924,7 +924,7 @@ The `Allocation` setting of a feature flag has the following properties:
 | `DefaultWhenDisabled` | Specifies which variant should be used when a variant is requested while the feature is considered disabled. |
 | `DefaultWhenEnabled` | Specifies which variant should be used when a variant is requested while the feature is considered enabled and no other variant was assigned to the user. |
 | `User` | Specifies a variant and a list of users to whom that variant should be assigned. | 
-| `Group` | Specifies a variant and a list of groups the current user has to be in for that variant to be assigned. |
+| `Group` | Specifies a variant and a list of groups the current user has to be in at least one of for that variant to be used. |
```
rossgrambo, line 927:
```suggestion
| `Group` | Specifies a variant and a list of groups. The variant will be assigned if the user is at least one of the groups.|
```

README.md
```diff
@@ -970,6 +972,52 @@ If you are using a feature flag with binary variants, the `StatusOverride` prope
...
+        "Variants": [ 
+            { 
+                "Name": "AlgorithmBeta" 
+            }
+        ] 
+    }
+}
+```
+
+**Note:** The `WithvariantService<TService>` method will not register any implementation of `TService`. You need to register variant serices as `TService` to allow `IVariantServiceProvider<TService>` to retrieve them from the dependency injection container.
```
rossgrambo, line 1007:
```suggestion
**Note:** The `WithVariantService<TService>` method will not register any implementation of `TService`. You need to register variant serices as `TService` to allow `IVariantServiceProvider<TService>` to retrieve them from the dependency injection container.

```

rossgrambo: I feel like we should include a snippet of registering the AlgorithmBeta class as defined in this note. Just

```csharp
services.AddSingleton<IAlgorithm, AlgorithmBeta>();
```

rossgrambo (approved):

README.md
zhiyuanliang-ms: User can register AlgorithemBeta in many ways:
``` C#
services.AddSingleton<IAlgorithm, AlgorithmBeta>()
services.AddSingleton<IAlgorithm>(sp => new AlgorithmBeta())
services.AddSingleton<IAlgorithm>(new AlgorithmBeta())
```
I don't want to leave an impression here that AlgorithmBeta has to be registered like this. My thought is that when the example application is ready, we can put a link here and say "For more details, please refer to this example."

zhiyuanliang-ms: How about this:
Which extension method to use for registration depends on your needs. Here is an example.
``` C#
services.AddSingleton<IAlgorithm, AlgorithmBeta>();
```

README.md
```diff
@@ -36,6 +36,7 @@ Here are some of the benefits of using this library:
 * [Targeting](#targeting)
   * [Targeting Exclusion](#targeting-exclusion)
 * [Variants](#variants)
+* [Variants in Dependency Injection](#variants-in-dependency-injection)
```
jimmyca15, line 39: Sub bullet.

README.md
```diff
@@ -934,7 +935,9 @@ If the feature is enabled, the feature manager will check the `User`, `Group`, a
 
 Allocation logic is similar to the [Microsoft.Targeting](./README.md#MicrosoftTargeting) feature filter, but there are some parameters that are present in targeting that aren't in allocation, and vice versa. The outcomes of targeting and allocation are not related.
 
-#### Overriding Enabled State with a Variant
+**Note:** To allow allocating feature variants, you need to register `ITargetingContextAccessor`. This could be done by calling `WithTargeting<T>` method.
```
jimmyca15, line 938:
```suggestion
**Note:** To allow allocating feature variants, you need to register `ITargetingContextAccessor`. This can be done by calling the `WithTargeting<T>` method.

```

README.md
```diff
@@ -970,6 +973,57 @@ If you are using a feature flag with binary variants, the `StatusOverride` prope
 
 In the above example, the feature is enabled by the `AlwaysOn` filter. If the current user is in the calculated percentile range of 10 to 20, then the `On` variant is returned. Otherwise, the `Off` variant is returned and because `StatusOverride` is equal to `Disabled`, the feature will now be considered disabled.
 
+## Variants in Dependency Injection
```
jimmyca15, line 976: Seems like it should be a triple `#` no?

README.md
```diff
@@ -970,6 +973,57 @@ If you are using a feature flag with binary variants, the `StatusOverride` prope
 
 In the above example, the feature is enabled by the `AlwaysOn` filter. If the current user is in the calculated percentile range of 10 to 20, then the `On` variant is returned. Otherwise, the `Off` variant is returned and because `StatusOverride` is equal to `Disabled`, the feature will now be considered disabled.
 
+## Variants in Dependency Injection
+
+Dependency injection can be wired up with a variant feature flag. You can use a variant feature flag to control which implementation of a service should be used by the dependency injection, based on the allocated variant of the feature flag. This could be done by using `IVariantServiceProvider<TService>`.
```
jimmyca15, line 978:
```suggestion
Variant feature flags can be used in conjunction with dependency injection to surface different implementations of a service for different users. This is accomplished through the use of the `IVariantServiceProvider<TService>` interface.

```

README.md
```diff
@@ -970,6 +973,57 @@ If you are using a feature flag with binary variants, the `StatusOverride` prope
...
+
+``` C#
+IVariantServiceProvider<IAlgorithm> algorithmServiceProvider;
+...
+
+IAlgorithm forecastAlgorithm = await algorithmServiceProvider.GetServiceAsync(cancellationToken); 
+```
+
+The `IVaraintServiceProvider<IAlgorithm>` will retrieve a specific implementation of `IAlgorithm` from the dependency injection container. It can be registered by calling `WithVariantService<TService>` on the `IFeatureManagementBuilder`.
```
jimmyca15, line 987:
```suggestion
In the snippet above, the `IVariantServiceProvider<IAlgorithm>` will retrieve an implementation of `IAlgorithm` from the dependency injection container. The chosen implementation is dependent upon:
* The feature that the `IAlgorithm` service was registered with
* The allocated variant for that feature

The `IVariantServiceProvider<T>` is made available to the application by calling `IFeatureManagementBuilder.WithVariantService<T>(string featureName)`. See below for an example.

```

README.md
```diff
@@ -970,6 +973,57 @@ If you are using a feature flag with binary variants, the `StatusOverride` prope
...
+            { 
+                "Name": "AlgorithmBeta" 
+            },
+            ...
+        ] 
+    }
+}
+```
+
+**Note:** The `WithvariantService<TService>` method will not register any implementation of `TService`. You need to register variant serices as `TService` to allow `IVariantServiceProvider<TService>` to retrieve them from the dependency injection container. Which extension method to use for registration depends on your needs. Here is an example.
```
jimmyca15, line 1009:
```suggestion
**Note:** The `WithVariantService<TService>` method will not register any implementation of `TService`. You need to register variant services as `TService` to allow `IVariantServiceProvider<TService>` to retrieve them from the dependency injection container. Which extension method to use for registration depends on your needs. Here is an example.

```

README.md
```diff
@@ -970,6 +973,57 @@ If you are using a feature flag with binary variants, the `StatusOverride` prope
...
+
+``` C#
+services.AddSingleton<IAlgorithm, AlgorithmBeta>();
```
jimmyca15, line 1012:
```suggestion
// Adds an implementation for `IAlgorithm`, which is consumed as a variant service above.
services.AddSingleton<IAlgorithm, AlgorithmBeta>();

```

jimmyca15: What happens if we forget to register implementation for the service?

README.md
```diff
@@ -934,7 +935,9 @@ If the feature is enabled, the feature manager will check the `User`, `Group`, a
...
 
-#### Overriding Enabled State with a Variant
+**Note:** To allow allocating feature variants, you need to register `ITargetingContextAccessor`. This could be done by calling `WithTargeting<T>` method.
+
+### Overriding Enabled State with a Variant
 
 You can use variants to override the enabled state of a feature flag. This gives variants an opportunity to extend the evaluation of a feature flag. If a caller is checking whether a flag that has variants is enabled, the feature manager will check if the variant assigned to the current user is set up to override the result. This is done using the optional variant property `StatusOverride`. By default, this property is set to `None`, which means the variant doesn't affect whether the flag is considered enabled or disabled. Setting `StatusOverride` to `Enabled` allows the variant, when chosen, to override a flag to be enabled. Setting `StatusOverride` to `Disabled` provides the opposite functionality, therefore disabling the flag when the variant is chosen. A feature with a `Status` of `Disabled` cannot be overridden.
```
rossgrambo, line 942: Maybe we should add a "why?" here, because there's only a narrow reason they would want to use the overrides. It's if a customer wants to allocate via variants without changing their existing IsEnabled code. Meaning they want to return bool results but have more than 2 "buckets".

README.md
zhiyuanliang-ms: It will return null if there is no matched implementation found in the DI container.

README.md
zhiyuanliang-ms: > Today, feature filters can be used to conditionally enable feature flags. Variants should be able to extend this functionality by also playing a role in the decision to enable a feature flag. A property can be included in a feature flag’s variants giving them the opportunity to override the flag’s enabled state. This is important considering the enabled state of a flag is used in our integration points with ASP.NET Core such as the FeatureGateAttribute

quoted from the [design doc](https://microsoft.sharepoint.com/:w:/r/teams/azconfig/_layouts/15/Doc.aspx?sourcedoc=%7BBA41C52A-B66F-4B9A-A3CA-D5CF6283BE06%7D&file=Variants%20as%20an%20extension%20of%20feature%20flags.docx&action=default&mobileredirect=true&share=IQEqxUG6b7aaS6PK1c9ig74GAQjMj0RepCJ3cjLLW_AeLf0)

This is the original purpose for this design.

> This gives variants an opportunity to extend the evaluation of a feature flag.

I think this sentence is a solid reason but we may come up a good story or a attracting use case of statusOverride.

rossgrambo: > This is important considering the enabled state of a flag is used in our integration points with ASP.NET Core such as the FeatureGateAttribute

This part is valid, but it's still not a big use case. Essentially, customers can always use a normal flag with filters to achieve whatever boolean targeting they want to achieve. Variants don't offer an advantage to boolean targeting over the standard TargetingFilter. 

One reason someone would use Variants in the bool world is to adding more telemetry "buckets" so they can split their traffic (although this is also not necessary as the split traffic of "true" results should be equal). 

Another reason is if a customer's plan is to call both GetVariant and IsEnabled(like using FeatureGateAttribute) on a single flag, then they might use it, but this is likely overloading what a flag should be in charge of. 

My fear is that users will see this section in the readme and think that they **should** be using it- instead of using it to address a specific problem.

README.md
jimmyca15: How about

The call above makes `IVariantServiceProvider<IAlgorithm>` available in the service collection. Implementation(s) of `IAlgorithm` must be added separately via an add method such as `services.AddSingleton<IAlgorithm, SomeImplementation>()`. The implementation of `IAlgorithm` that the `IVariantServiceProvider` uses depends on the `ForecastAlgorithm` feature. If no implementation of `IAlgorithm` is added to the service collection, then the `IVariantServiceProvider<IAlgorithm>.GetServiceAsync()` will return a task with a null result.

jimmyca15 (approved):

Issue #93, opened 2020-09-03
Custom FeatureDefinitionProvider still depends on IConfiguration


pardahlman:
I've been look at the latest pre-release of this package (`2.2.0-preview`) - it is a great improvement in terms of migration from other feature management system 👍

However, when implementing a custom `IFeatureDefinitionProvider` I noticed that `FeatureFilterConfiguration` still expects its data to be modeled as `IConfiguration`. There are a few reasons why I think this is problematic:

* **Non-transparent API**. It is not transparent what settings are mandatory or even expected, `IConfiguration` is very broad.
* **Cumbersome API**. In my case, I have loaded feature toggle data that I want to map to percentage filter, groups etc. The best way I have found is to look at the source code for the filter classes and mimic the expected configuration in an in-memory configuration collection.
* **Surprising API**. Configuration is a concept that is tied to application startup. Whenever code is reliant on configuration, the first step is usually to map it the configuration to a strongly typed class and let downstream classes rely on it.

It would be great if the API for implementing a custom data source/adapter for the feature management system would be strongly typed and not reliant on `IConfiguration`.

jimmyca15:
Thank you for the feedback @pardahlman . I can see why you have listed these points. We chose to stick with IConfiguration for two main reasons. First, the library is designed by default around pulling feature flags from configuration. Secondly, the feature filter parameters are the configuration for an individual instance of a feature filter.

We will discuss how we can improve this and see if the requirement to use IConfiguration can be laxed in favor of another way to specify feature filter parameters. This will not make it into 2.2.0 as we have already started the process of promoting 2.2.0-preview.

DovPet:
Hi, I don't know where I have to write this, but it is very unclear how to create CustomFeatureDefinitionProvider and choose my provider to be database not appsettings.json file. In documentation it is mentioned, but it is fuzzy how to use it in code. I think, just a simple example of database implementation would be very great .

jimmyca15:
Hello @DovPet we can look into adding an example like that. For the time being though, we do have [this test](https://github.com/microsoft/FeatureManagement-Dotnet/blob/2240941480f833fea27601368eb910260ce71966/tests/Tests.FeatureManagement/FeatureManagement.cs#L488) that uses an in-memory feature definition provider. It should provide a starting point, the only difference would be that your features would come from a database instead of in-memory.

jimmyca15:
[...]

PR #389, opened 2024-03-09
Add getting features from database example

fffffatah:
Simple example for getting features from a database. Example includes,

- A custom FeatureDefinitionProvider.
- Getting features using EFCore and SQLite.

### DIFF
--- a/examples/GettingFeaturesFromDatabase/Database/ServiceCollectionExtensions.cs
+++ b/examples/GettingFeaturesFromDatabase/Database/ServiceCollectionExtensions.cs
@@ -1,0 +1,20 @@
+﻿using GettingFeaturesFromDatabase.Database.Services;
+using Microsoft.EntityFrameworkCore;
+
+namespace GettingFeaturesFromDatabase.Database;
+
+public static class ServiceCollectionExtensions
+{
+    public static void AddDatabase(this IServiceCollection services)
+    {
+        services.AddDbContext<SqliteDbContext>(options =>
+        {
+            options.UseSqlite("Data Source=example.db");
+        });
+    }
+    
+    public static void AddFeatureService(this IServiceCollection services)
+    {
+        services.AddScoped<IFeatureService, FeatureService>();
+    }
+}
--- a/examples/GettingFeaturesFromDatabase/Program.cs
+++ b/examples/GettingFeaturesFromDatabase/Program.cs
@@ -1,0 +1,23 @@
+using GettingFeaturesFromDatabase;
+using GettingFeaturesFromDatabase.Database;
+using Microsoft.FeatureManagement;
+
+var builder = WebApplication.CreateBuilder(args);
+
+builder.Services.AddControllers();
+builder.Services.AddEndpointsApiExplorer();
+builder.Services.AddDatabase();
+builder.Services.AddFeatureService();
+
+/* Add feature management */
+builder.Services.AddScoped<IFeatureDefinitionProvider, CustomFeatureDefinitionProvider>().AddScopedFeatureManagement();
+
+var app = builder.Build();
+
+app.UseHttpsRedirection();
+
+app.UseAuthorization();
+
+app.MapControllers();
+
+app.Run();
--- a/Microsoft.FeatureManagement.sln
+++ b/Microsoft.FeatureManagement.sln
@@ -25,6 +25,8 @@ Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "TargetingConsoleApp", "exam
 EndProject
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "BlazorServerApp", "examples\BlazorServerApp\BlazorServerApp.csproj", "{12BAB5A6-4EEB-4917-B5D9-4AFB6253008E}"
 EndProject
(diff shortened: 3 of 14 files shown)


fffffatah:

examples/GettingFeaturesFromDatabase/Program.cs
```diff
@@ -0,0 +1,23 @@
...
+
+var builder = WebApplication.CreateBuilder(args);
+
+builder.Services.AddControllers();
+builder.Services.AddEndpointsApiExplorer();
+builder.Services.AddDatabase();
+builder.Services.AddFeatureService();
+
+/* Add feature management */
+builder.Services.AddScoped<IFeatureDefinitionProvider, CustomFeatureDefinitionProvider>().AddScopedFeatureManagement();
```
zhiyuanliang-ms, line 13: Hi, @fffffatah Maybe this is a stupid question. May I know the reason why CustomFeatureDefinitionProvider is registered as scoped here?

fffffatah: Hi @zhiyuanliang-ms 👋, CustomFeatureDefinitionProvider is scoped here because the life-time of DbContext and the service which is used to read the features from DB are scoped. If I'm not wrong, setting the life time of DbContext as singleton might lead to issues when many concurrent calls are made. Although, DbContext can be registered as singleton and the Feature set can be marked AsNoTracking while querying, but it felt a bit risky.

zhiyuanliang-ms: Thank you!

zhiyuanliang-ms: Thank you!

examples/GettingFeaturesFromDatabase/Database/ServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,20 @@
...
+
+namespace GettingFeaturesFromDatabase.Database;
+
+public static class ServiceCollectionExtensions
+{
+    public static void AddDatabase(this IServiceCollection services)
+    {
+        services.AddDbContext<SqliteDbContext>(options =>
+        {
+            options.UseSqlite("Data Source=example.db");
```
zhiyuanliang-ms, line 12: What's example.db here?

zhiyuanliang-ms:
@fffffatah When I tried to run this application, it gave the result of "localhost page can not be found". I guess this is because there is no database existed and the feature manager will consider the unexisted feature flag as disabled.

Would you mind including a README? ([example 1](https://github.com/microsoft/FeatureManagement-Dotnet/blob/main/examples/BlazorServerApp/README.md) [example 2](https://github.com/microsoft/FeatureManagement-Dotnet/blob/preview/examples/EvaluationDataToApplicationInsights/README.md))

examples/GettingFeaturesFromDatabase/Database/ServiceCollectionExtensions.cs
fffffatah: It's the SQLite database file.

fffffatah:
@zhiyuanliang-ms I simplified the example for initial run and added Swagger UI for ease of use. Also, added a README to get started. Sorry about the exception, it happened due to a package mismatch.

zhiyuanliang-ms:
@fffffatah The application works for me now. But It seems like the CI pipeline failed. I will take some to look into it.

Thank you for your interest and support for this repo. Appreciate you spending the time on this PR.
I want to point out the significant value of this PR to us, even if it ultimately might not be merged.

This PR fills a missing piece of examples in this repo. We provides the IFeatureDefinitionProvider interface and this interface provides the extensibility to allow users to implement their own feature definition provider. But we don't have any detailed tutorial or example about how to implement IFeatureDefinitionProvider, which was called out 3 years ago. (Previous issue can be found in #93)

This PR also lets us know how actually user will use our FM lib to work with feature flags from database. The use case of importing feature flag from database is common especially when users are migrating from other third-party feature management libs. This is super valuable for us. Really appreciate that. So even if this PR cannot be merged, it brings a good chance for us to seriously consider how to build an example app of implementing a custom feature defintion provider from a database. 

Another interesting thing is that this PR uses the `AddScopedFeatureManagement` API which we presented in 3.1.0. The very initial idea of this API is to support Blazor server application to use Feature Management. But it turns out this API can be helpful when working with database according to your comment: https://github.com/microsoft/FeatureManagement-Dotnet/pull/389#discussion_r1518834820 I will spend some time to study why database usage requires scoped services.

Thanks again for your effort. @fffffatah This PR is super valuable for us.

examples/GettingFeaturesFromDatabase/Database/ServiceCollectionExtensions.cs
zhiyuanliang-ms: Thanks. Now I see the application will create an example.db during initialization.

zhiyuanliang-ms:
@fffffatch 
The Azure pipeline failed and it reported the following error. It seems the dll of sqlite is not compliant.
image

But I saw your first commit can pass the pipeline and you also used sqlite there.

zhiyuanliang-ms:
@fffffatah 

I guess `SQL Server` will not pass the Azure pipeline either. I do remember I met some trouble while including SQL stuff in some example projects.
I tried to include login/register function in an example project, which used SQL server to store user info. That PR failed to pass the Azure pipeline. But I am not 100% sure about whether using SQL server will make the PR fail to pass the CI check. 

Even if a PR can not pass the pipeline check, it still can be merged. So, don't be too worried about it.

Comparing to SQL server, I think SQLite may be a better choice, because it is lighter and user can use it directly as long as the nuget package is installed.

Issue #376, opened 2024-02-23
Consider adding default for CancellationToken parameters in interfaces / classes in 4.0.0


brental:
It is a common pattern for interface and class methods with CancellationToken parameters to specify a default for the parameter so that it can be called without the parameter by consumers. This includes interface and class methods with CancellationToken parameters in various Microsoft libraries such as (but not limited to) the dotnet runtime and AspNetCore as well as in other libraries. For consistency with this established pattern, consider adding this to the interface and class methods where CancellationToken parameters were added in the 4.0.0 preview.

As an example:
IAsyncEnumerable<string> GetFeatureNamesAsync(CancellationToken cancellationToken **= default**);

zhenlan:
Agreed. This is an oversight. Thanks @brental.

@zhiyuanliang-ms @rossgrambo can one you take this on. Please check all interfaces/types applicable.

zhiyuanliang-ms:
@zhenlan Jimmy, Ross and I had a discussion about this. I just shared the chat history to you.

zhiyuanliang-ms:
@zhenlan Jimmy, Ross and I had a discussion about this. I just shared the chat history to you.

zhiyuanliang-ms:
@zhenlan @brental 
Not providing default value for cancellationToken is by design.Here is the discussion of this API: https://github.com/microsoft/FeatureManagement-Dotnet/pull/250#discussion_r1291625266

Last month, when we saw @brental opened this issue, we had another discussion about it. 

We agree that it will make the caller code look simpler if we provide default value for cancellation token. 

But, omitting cancellation token on async method invocation is not a good practice. Caller should be explicit that it does't intend to be canceled. It is terrible to accidentally omit the propagation of cancellation.

If .NET has established the paradigm, we should not go against the it. So we did some investigation about how common the default value pattern is. 

Cancellation token with default value is not common in the Azure .NET SDK.
: https://github.com/search?q=repo%3AAzure%2Fazure-sdk-for-net+%22%2C+CancellationToken+token%22&type=code

System.IO lib provides default value for cancellation token: https://learn.microsoft.com/en-us/dotnet/api/system.io.stream.readasync?view=net-8.0&source=docs

[...]

PR #395, opened 2024-03-11
Optional Cancellation Token

zhiyuanliang-ms:
According to #376  

BTW, all AppConfig SDK public APIs made the CancellationToken optional: https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/appconfiguration/Azure.Data.AppConfiguration/src/ConfigurationClient.cs

This PR makes cancellation token optional for async methods of IVariantFeatureManager and adds cancellation token to async methods of ISessionManager


Eskibear (approved):
LGTM

PR #400, opened 2024-03-14
Remove redundant if check for logger

zhiyuanliang-ms:
Maintain a consistent code style.
https://github.com/microsoft/FeatureManagement-Dotnet/blob/main/src/Microsoft.FeatureManagement/FeatureManager.cs#L219
https://github.com/microsoft/FeatureManagement-Dotnet/blob/main/src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs#L367


Eskibear (approved):

Eskibear:
Just some thoughts: Is it possible to setup linting check in CI? That can be helpful.

Issue #403, opened 2024-03-18
Targeting evaluation is CPU dependent.


jimmyca15:
Targeting evaluation uses BitConvert.ToInt32 as seen [here](https://github.com/microsoft/FeatureManagement-Dotnet/blob/5dfdb23c22eecc6662afc4700966113b20e1f822/src/Microsoft.FeatureManagement/Targeting/ContextualTargetingFilter.cs#L162). According to `ToInt32`'s [documentation](https://learn.microsoft.com/en-us/dotnet/api/system.bitconverter.toint32?view=net-8.0), the result is dependent upon the CPU's endianness. 

Targeting evaluation should be consistent across CPU architectures. Given the current documented and intended usage is server scenarios, which typically run on x86, I propose to use a replacement that would give results consistent with `BitConvert.ToInt32` run on x86 systems.

rossgrambo:
Not sure if there's a good replacement. We could explicitly check and reverse the bits?

```csharp
if (BitConverter.IsLittleEndian)
{
    Array.Reverse(bytes);
}
```

zhiyuanliang-ms:
@rossgrambo I think the method you proposed is fine. I don't have any better idea. 

BTW, X86 and other common CPU architectures are all little-endian. So we'd better treat big-endian as special case.

jimmyca15:
Sounds reasonable, however no need to reverse the whole array. Only the bytes used for the int32 are needed.

zhiyuanliang-ms:
This issue also exists in Java FM: https://github.com/Azure/azure-sdk-for-java/blob/main/sdk/spring/spring-cloud-azure-feature-management/src/main/java/com/azure/spring/cloud/feature/management/filters/TargetingFilter.java#L256

The default endianness used by ByteBuffer is big-endian: https://docs.oracle.com/javase/8/docs/api/java/nio/ByteBuffer.html
@mrm9084 

The python FM is fine since it explicitly uses little-endian:
https://github.com/microsoft/FeatureManagement-Python/blob/main/featuremanagement/_defaultfilters.py#L86

PR #405, opened 2024-03-20
Ensure the consistency of targeting evaluation across CPU architectures.

zhiyuanliang-ms:
#403

### DIFF
--- a/src/Microsoft.FeatureManagement/Targeting/TargetingEvaluator.cs
+++ b/src/Microsoft.FeatureManagement/Targeting/TargetingEvaluator.cs
@@ -349,9 +349,21 @@ namespace Microsoft.FeatureManagement.Targeting
             }
 
             //
-            // Use first 4 bytes for percentage calculation
+            // The first 4 bytes of the hash will be used for percentage calculation
+            byte[] relevantBytes = new byte[4];
+
+            Array.Copy(hash, 0, relevantBytes, 0, 4);
+
+            //
+            // If the system is not little-endian, reverse the bytes.
+            if (!BitConverter.IsLittleEndian)
+            {
+                Array.Reverse(relevantBytes);
+            }
+
+            //
             // Cryptographic hashing algorithms ensure adequate entropy across hash
-            uint contextMarker = BitConverter.ToUInt32(hash, 0);
+            uint contextMarker = BitConverter.ToUInt32(relevantBytes, 0);
 
             double contextPercentage = (contextMarker / (double)uint.MaxValue) * 100;
 
--- a/src/Microsoft.FeatureManagement/Targeting/TargetingEvaluator.cs
+++ b/src/Microsoft.FeatureManagement/Targeting/TargetingEvaluator.cs
@@ -343,6 +343,8 @@ namespace Microsoft.FeatureManagement.Targeting
         {
             byte[] hash;
 
+            //
+            // Cryptographic hashing algorithms ensure adequate entropy across hash
             using (HashAlgorithm hashAlgorithm = SHA256.Create())
             {
                 hash = hashAlgorithm.ComputeHash(Encoding.UTF8.GetBytes(contextId));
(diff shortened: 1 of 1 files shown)


rossgrambo (approved):

src/Microsoft.FeatureManagement/Targeting/TargetingEvaluator.cs
```diff
@@ -348,8 +348,16 @@ private static bool IsTargeted(string contextId, double from, double to)
...
+            // Endianness check ensures the consistency accross different architectures
+            if (!BitConverter.IsLittleEndian)
+            {
+                Array.Reverse(hash, 0, 4);
+            }
+
             //
-            // Use first 4 bytes for percentage calculation
             // Cryptographic hashing algorithms ensure adequate entropy across hash
             uint contextMarker = BitConverter.ToUInt32(hash, 0);
```
rossgrambo, line 362: I kind of want to explicitly subset the hash to the first 4 bytes just to make things clear. But not needed.

src/Microsoft.FeatureManagement/Targeting/TargetingEvaluator.cs
```diff
@@ -348,8 +348,16 @@ private static bool IsTargeted(string contextId, double from, double to)
                 hash = hashAlgorithm.ComputeHash(Encoding.UTF8.GetBytes(contextId));
             }
 
+
+            //
+            // The first 4 bytes of the hash will be used for percentage calculation
+            // Endianness check ensures the consistency accross different architectures
```
juniwang, line 354: `accross `. typo?

src/Microsoft.FeatureManagement/Targeting/TargetingEvaluator.cs
zhiyuanliang-ms: Do you mean:

``` c#
byte[] relevantBytes = new byte[4];

Array.Copy(hash, 0, relevantBytes, 0, 4);


if (!BitConverter.IsLittleEndian)
{
    Array.Reverse(relevantBytes);
}

uint contextMarker = BitConverter.ToUInt32(relevantBytes, 0);
```

I used to implement it like this in [30ec65c43b35ebe46ef7c7327b327b150e7a0f18](https://github.com/microsoft/FeatureManagement-Dotnet/pull/405/commits/30ec65c43b35ebe46ef7c7327b327b150e7a0f18)

rossgrambo: Something like that, but I was hoping for something more like:

```csharp
uint contextMarker = BitConverter.ToUInt32(hash.Take(4), 0);
```

jimmyca15: That's allocation for no reason.

rossgrambo: Reason is for the next developer who's reading it to understand it- the way it is now makes it look like we reverse the first 4 bytes but somehow use all of the bytes to determine a resulting uint.

If we'd rather not add the extra array, maybe we could adjust the comments to make it more clear to me. What if we reordered/adjusted them like this:

```csharp
//
// Cryptographic hashing algorithms ensure adequate entropy across hash
using (HashAlgorithm hashAlgorithm = SHA256.Create())
{
    hash = hashAlgorithm.ComputeHash(Encoding.UTF8.GetBytes(contextId));
}

//
// Endianness check ensures the consistency of targeting evaluation result across different architectures
if (!BitConverter.IsLittleEndian)
{
    Array.Reverse(hash, 0, 4);
}

//
// Only the first 4 bytes of the hash will converted to a uint
uint contextMarker = BitConverter.ToUInt32(hash, 0);
```

jimmyca15: Sure comments, but no need to add allocation for it.

jimmyca15 (approved):

PR #407, opened 2024-03-20
Support variant for FeatureTagHelper

zhiyuanliang-ms:
We've already offer `<feature>` tags which can be used to conditionally render content based on whether a feature is enabled or not.

```cshtml
<feature name="FeatureX">
  <p>This can only be seen if 'FeatureX' is enabled.</p>
</feature>
```

We want to extend the feature tag helper to support variant.

The usage will be like:
```cshtml
<feature name=”Alpha” variant=”V1”> Content only for Variant 1</feature> 
<feature name=”Alpha” variant=”V2> Content only for Variant 2</feature> 
```


src/Microsoft.FeatureManagement.AspNetCore/TagHelpers/FeatureTagHelper.cs
```diff
@@ -52,16 +59,35 @@ public override async Task ProcessAsync(TagHelperContext context, TagHelperOutpu
...
+
+                    IEnumerable<string> variants = Variant.Split(',').Select(n => n.Trim());
+
+                    enabled = await variants.Any(
+                        async variant => {
+                            Variant assignedVariant = await _featureManager.GetVariantAsync(features.First());
+
+                            return variant == assignedVariant?.Name;    
+                        });
+                }
```
rossgrambo, line 90: I think we should support Negate in the variants case too.

zhenlan: I can think of scenarios where variant negate can be useful.

zhenlan:
It will be helpful if you can add description and examples of usage.

src/Microsoft.FeatureManagement.AspNetCore/TagHelpers/FeatureTagHelper.cs
zhiyuanliang-ms: Negate can be treated as just reverse the original behavior of the tag helper. In this case, I agree Negate can also apply to variant.

rossgrambo: Normal Feature Flag Negate:
```
<feature negate="true" name="ShouldEnable">
  <p>This can only be seen if 'ShouldEnable' is not enabled.</p>
</feature>
```

Variant Feature Flag Negate
```
<feature negate="true" name=”Alpha” variant=”V1”>
  <p>This can only be seen if 'Alpha' is not V1.</p>
</feature>
```

jimmyca15: Ah, it does make sense.

src/Microsoft.FeatureManagement.AspNetCore/TagHelpers/FeatureTagHelper.cs
```diff
@@ -30,13 +31,19 @@ public class FeatureTagHelper : TagHelper
         /// </summary>
         public bool Negate { get; set; }
 
+        /// <summary>
+        /// A variant name, or comma separated list of variant names. If any of specified variants is assigned, the content should be rendered.
+        /// If variant is specified, <see cref="Negate"/> and <see cref="Requirement"/> will be ignored. Besides, <see cref="Name"/> should be exactly one feature name.
```
juniwang, line 36: what's the reason of ignoring `Negate`?

zhiyuanliang-ms: I forget to update the comment. `Negate` is supported.

src/Microsoft.FeatureManagement.AspNetCore/TagHelpers/FeatureTagHelper.cs
```diff
@@ -30,13 +31,19 @@ public class FeatureTagHelper : TagHelper
         /// </summary>
         public bool Negate { get; set; }
 
+        /// <summary>
+        /// A variant name, or comma separated list of variant names. If any of specified variants is assigned, the content should be rendered.
+        /// If variant is specified, <see cref="Requirement"/> will be ignored. Besides, <see cref="Name"/> should be exactly one feature name.
```
rossgrambo, line 36:
```suggestion
        /// If variant is specified, <see cref="Name"/> must contain only one feature name and <see cref="Requirement"/> will have no effect.
```

zhiyuanliang-ms:
README update is WIP.

zhiyuanliang-ms:
@jimmyca15 @rossgrambo Could you help review/approve this PR?

README.md
```diff
@@ -248,31 +248,51 @@ public interface IDisabledFeaturesHandler
 
 ### View
 
-In MVC views `<feature>` tags can be used to conditionally render content based on whether a feature is enabled or not.
+In MVC views `<feature>` tags can be used to conditionally render content based on whether a feature is enabled or whether specific variant of a feature is assigned. For more information about variant, please refer to the [variants](./README.md#Variants) section.
```
rossgrambo, line 251:
```suggestion
In MVC views `<feature>` tags can be used to conditionally render content based on whether a feature is enabled or whether specific variant of a feature is assigned. For more information about variants, please refer to the [variants](./README.md#Variants) section.

```

README.md
```diff
@@ -248,31 +248,51 @@ public interface IDisabledFeaturesHandler
...
-By default, all listed features must be enabled for the feature tag to be rendered. This behavior can be overidden by adding the `requirement` attribute as seen in the example below.
+``` HTML+Razor
+<feature name="FeatureX" variant="Alpha,Beta">
+  <p>This can only be seen if variant 'Alpha' or 'Beta' of 'FeatureX' is assigned.</p>
+</feature>
+```
+
+**Note:** if `variant` is specified, only *one* feature should be specified. 
+
+By default, all listed features must be enabled for the feature tag to be rendered. This behavior can be overidden by adding the `requirement` attribute as seen in the example below. Note that the `requirement` attribute will have no effect on `variant`.
```
rossgrambo, line 295:
```suggestion
By default, all listed features must be enabled for the feature tag to be rendered. This behavior can be overidden by adding the `requirement` attribute as seen in the example below. `variant` will always use *any* logic, meaning the `requirement` attribute will have no effect on it.
```

src/Microsoft.FeatureManagement.AspNetCore/TagHelpers/FeatureTagHelper.cs
```diff
@@ -52,11 +59,35 @@ public override async Task ProcessAsync(TagHelperContext context, TagHelperOutpu
...
+                else
+                {
+                    if (features.Count() != 1)
+                    {
+                        throw new ArgumentException("Variant cannot be associated with multiple feature flags.", nameof(Name));
+                    }
+
+                    IEnumerable<string> variants = Variant.Split(',').Select(n => n.Trim());
+
+                    if (variants.Count() != 1 && Requirement == RequirementType.All)
```
rossgrambo, line 79: This check goes against the "Have no effect" we said in the readme. We should either always assume RequirementType.Any or we should throw. I believe we aligned on throwing.

zhiyuanliang-ms: I remember I have removed "Have no effect" in readme.

zhiyuanliang-ms: Updated.

rossgrambo (approved):

Issue #408, opened 2024-03-23
Remove the support for variants and telemetry from the .NET feature flag schema


zhenlan:
Open a tracking issue following the discussion: https://github.com/microsoft/FeatureManagement-Dotnet/issues/393#issuecomment-1989037173

We should do this before v4.0.0 turns stable.

rossgrambo:
Deprecating the .NET FF schema is not as straightforward as that discussion makes it sound. For the FM library, it can make the switch, but the provider today only returns the Dotnet schema and can't make the switch as easily. 

This change means the provider needs to switch to snake_case, which means newer versions of the provider will not work with old versions of FeatureManagement. Unless the provider can do some version determination but that doesn't sound great.

cc; @jimmyca15 @amerjusupovic

jimmyca15:
There is a way to support it that I thought of. Throwing it out there. 

Flags with variants are put into IConfiguration with the Microsoft FF schema. Flags without variants are put into IConfiguration with the .NET FF schema.

I.e.

```
{
    "FeatureManagement": {
        "Beta": true
    },
    "feature_management": {
        "feature_flags": [
            {
                "id": "Gamma",
                "Variants": [
                    ...
                ]
            }
        ]
    }
}
```

The .NET feature management library would need to respect both sections being defined.

rossgrambo:
We could go that route, but I'm concerned someone would add a variant to an existing flag, and now suddenly that flag disappears. I suppose the UI doesn't support it- but it doesn't sound ideal.

rossgrambo:
Without making code changes, if a customer has <V4 of FM and they use portal to edit the flag and add a variant, refresh would make the flag disappear.

zhenlan:
You added variant to an existing feature flag, but you didn't update your application to use the FM library that supports variant. What's your expected outcome for this scenario?

rossgrambo:
The current functionality, which is their flag continues to work as a flag without variants until they update FM to read the "Variants" & other fields. This update lets GetVariant work and lets the StatusOverride work.

jimmyca15:
> Without making code changes, if a customer has <V4 of FM and they use portal to edit the flag and add a variant, refresh would make the flag disappear in their running application. (If they had latest provider)

We don't allow updating an existing flag to add variants though.

rossgrambo:
> We don't allow updating an existing flag to add variants though.

[...]

PR #414, opened 2024-04-02
Promote Microsoft FM schema in README

zhiyuanliang-ms:
#408 

We are planning to retire the usage of .NET Feature Management schema. We want to no longer support new features for the .NET FM schema starting from v4.0.0. All new features will only be added to the Microsoft FM schema. To promote the Microsoft schema, this PR replaced all examples in the README with Microsoft feature flags.


README.md
```diff
@@ -198,37 +103,47 @@ The feature management library also supports the usage of the [`Microsoft Featur
...
+The `feature_management` section of the json document is used by convention to load feature flag settings. Feature flag objects must be listed in the `feature_flags` array under the `feature_management` section. In the section above, we see that we have provided three different features. A feature flag has `id` and `enabled` properties. The `id` is the name used to identify and reference the feature flag. The `enabled` is the enabled state of the feature flag. A feature is *OFF* if `enabled` is false. If `enabled` is true, then the feature is *ON* if there are no `conditions` (null or empty) or if the `conditions` are satisfied. The `conditions` property declare the conditions used to dynamically enabled the feature. Features define their feature filters in the `client_filters` array under the `conditions`. `FeatureV` specifies a feature filter named `Microsoft.TimeWindow`. This is an example of a configurable feature filter. We can see in the example that the filter has a `Parameters` property. This is used to configure the filter. In this case, the start and end times for the feature to be active are configured.
```
jimmyca15, line 106:
```suggestion
The `feature_management` section of the json document is used by convention to load feature flag settings. Feature flag objects must be listed in the `feature_flags` array under the `feature_management` section. In the section above, we see that we have provided three different features. A feature flag has `id` and `enabled` properties. The `id` is the name used to identify and reference the feature flag. The `enabled` property specifies the enabled state of the feature flag. A feature is *OFF* if `enabled` is false. If `enabled` is true, then the state of the feature depends on `conditions`. If there are no `conditions` then the feature is *ON*. If there are `conditions` and the conditions are met then the feature is *ON*. If there are `conditions` and they are not met then the feature is *OFF*. The `conditions` property declares the conditions used to dynamically enabled the feature. Features define their feature filters in the `client_filters` array. `FeatureV` specifies a feature filter named `Microsoft.TimeWindow`. This is an example of a configurable feature filter. We can see in the example that the filter has a `Parameters` property. This is used to configure the filter. In this case, the start and end times for the feature to be active are configured.
```

README.md
```diff
@@ -198,37 +103,47 @@ The feature management library also supports the usage of the [`Microsoft Featur
...
     }
 }
 ```
 
-**Note:** If the `feature_management` section can be found in the configuration, the `FeatureManagement` section will be ignored.
+In the above example, `FeatureW` specifies a `requirement_type` of `All`, meaning all of it's filters must evaluate to true for the feature to be enabled. In this case, the feature will be enabled for 50% of users during the specified time window.
+
+### .NET Feature Management schema
+
+The feature management library used to use the [.NET Feature Management schema](./schemas/FeatureManagement.Dotnet.v1.0.0.schema.json) as the format used to set up feature flags. Starting from v4.0.0, new features including Variants and Telemetry will not be supported for the .NET Feature Managemen schema.
```
jimmyca15, line 146:
```suggestion
In previous versions, the primary schema for the feature management library was the [.NET feature management schema](./schemas/FeatureManagement.Dotnet.v1.0.0.schema.json). Starting from v4.0.0, new features including variants and telemetry will not be supported for the .NET feature management schema.
```

schemas/FeatureManagement.Dotnet.v2.0.0.schema.json
```diff
@@ -1,313 +0,0 @@
-{
```
jimmyca15, line 1: I suppose we'll also need a code update in FeatureManager to not respect variants and telemetry when defined in .NET schema, right?

jimmyca15: Is it planned as a separate PR?

zhiyuanliang-ms: > Is it planned as a separate PR?

Yes.

README.md
```diff
@@ -58,52 +58,11 @@ The .NET Core configuration system is used to determine the state of feature fla
 
 ### Feature Flag Declaration
 
-The feature management library supports appsettings.json as a feature flag source since it is a provider for .NET Core's IConfiguration system. Below we have an example of the format used to set up feature flags in a json file.
+The feature management library supports appsettings.json as a feature flag source since it is a provider for .NET Core's IConfiguration system. The feature management library uses the [`Microsoft Feature Management schema`](https://github.com/Azure/AppConfiguration/blob/main/docs/FeatureManagement/FeatureManagement.v1.0.0.schema.json) to declare feature flags. This schema is language agnostic in origin and is supported by all Microsoft feature management libraries.
```
jimmyca15, line 61:
```suggestion
The feature management library supports appsettings.json as a feature flag source since it is a provider for .NET Core's IConfiguration system. Feature flags are declared using the [`Microsoft Feature Management schema`](https://github.com/Azure/AppConfiguration/blob/main/docs/FeatureManagement/FeatureManagement.v1.0.0.schema.json). This schema is language agnostic in origin and is supported by all Microsoft feature management libraries.

```

jimmyca15 (approved):

jimmyca15 (approved):

schemas/FeatureManagement.Dotnet.v2.0.0.schema.json
zhiyuanliang-ms: This PR: https://github.com/microsoft/FeatureManagement-Dotnet/pull/421

Issue #415, opened 2024-04-03
Add Default Http Targeting Context Accessor


rossgrambo:
Today, users must define their own implementation of `ITargetingContextAccessor`. Users need to build it from scratch, base it off one of [our examples](https://github.com/microsoft/FeatureManagement-Dotnet/blob/main/examples/FeatureFlagDemo/HttpContextTargetingContextAccessor.cs), or [follow our guide on targeting](https://learn.microsoft.com/en-us/azure/azure-app-configuration/howto-targetingfilter-aspnet-core#update-the-web-application-code-to-use-targetingfilter) (which builds something similar to our example).

We could ease this pain by offering a default implementation `ITargetingContextAccessor`.

Issue #399, opened 2024-03-14
The TargetingId should be included in the FeatureEvaluation event even if it's null/empty.


zhenlan:
From what I can see in the telemetry data, TargetingId appears to be skipped if it's null/empty, but we do send a FeatureEvaluation event for it. We should include it even if it's null/empty.

(We've discussed this. Open an issue to track it.)

rossgrambo:
Currently null and empty string are treated as different targets. Should we continue that? I believe python is treating them as the same. If we're going to emit it, the languages need to do the same thing. cc; @mrm9084

jimmyca15:
> Currently null and empty string are treated as different targets

Can you explain what you mean here? Do you mean if someone sets TargetingContext.UserId to null vs "", there may be a different variant returned?

rossgrambo:
They get the same variant right now, but if we removed the null or empty check on the outgoing telemetry- we can emit events where the userId is null and other events where the userId is "". 

Python would never emit an event for null, as it defaults to "". So a null userId in C# would have a differently shaped impression event than that same "null" user in python. 

tldr; should we commit to "" as the null/empty targetingId?

jimmyca15:
As far as I'm concerned, there is no such thing as a null user id. Our system should effectively treat it as empty string.

rossgrambo:
Turns out both null and empty are ignored by Application Insights SDK and the field won't be added. So unfortunately we can't include this on the impression event. 

We decided to have the ActivityEvent optionally include these as well to follow suit.

PR #416, opened 2024-04-03
Adds default http targeting context accessor

rossgrambo:
## Why this PR?

1. Addresses issues 
  a. https://github.com/microsoft/FeatureManagement-Dotnet/issues/415
  b. https://github.com/microsoft/FeatureManagement-Dotnet/issues/399
2. Also makes HttpContext.Items cache consistent- so all classes check the same location. 

Next Steps:
1. Remove unremarkable HttpTargetingContextAccessors from examples in favor of DefaultHttpTargetingContextAccessor.
2. Consider a different cache pattern that doesn't rely on a shared string key.

## Visible Changes

Developers will now have a new type available to them, `DefaultHttpTargetingContextAccessor`. Instead of defining their own, they can use this type when adding targeting via `WithTargeting` like:

```csharp
    .WithTargeting<DefaultHttpTargetingContextAccessor>()
```

The default accessor will extract the targeting info from `HttpContext.User`. `Groups` will be extracted from claims of type `Role`. `UserId` is taken from the `Identity.Name` field.


rossgrambo:
If someone has an idea for a shared place for the `TargetingContextLookup` string, let me know. I couldn't decide on a clean spot. Decided to leave them as separate strings for now since we're hoping to adjust the caching to not need them anyway.

zhenlan:
Just for ideas... you either make it a public property or use InternalsVisibleToAttribute between the two packages. Like you said, none of them are clean ways.

jimmyca15:
I know adding a scoped service will be an approach that we will want to propose to float the resolved targeting context along, but I also thing we should consider exposing a public extension method on HttpContext from the `Microsoft.FeatureManagement.AspNetCore` package that gets the resolved targeting context.

```
static class HttpContextExtensions
{
    //
    // Gets the targeting context resolved by feature management during a request.
    public static TargetingContext GetTargetingContext(this httpContext);
}
```

src/Microsoft.FeatureManagement.AspNetCore/DefaultHttpTargetingContextAccessor.cs
```diff
@@ -0,0 +1,68 @@
...
+//
+using Microsoft.AspNetCore.Http;
+using Microsoft.FeatureManagement.FeatureFilters;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Security.Claims;
+using System.Threading.Tasks;
+
+namespace FeatureFlagDemo
```
zhiyuanliang-ms, line 12: namespace is wrong

src/Microsoft.FeatureManagement.AspNetCore/DefaultHttpTargetingContextAccessor.cs
```diff
@@ -0,0 +1,68 @@
...
+            // Try cache lookup
+            if (httpContext.Items.TryGetValue(TargetingContextLookup, out object value))
+            {
+                return new ValueTask<TargetingContext>((TargetingContext)value);
+            }
+
+            //
+            // Treat user identity name as user id
+            ClaimsPrincipal user = httpContext.User;
+            string userId = user?.Identity?.Name;
```
zhiyuanliang-ms, line 47: Could you add an empty line before this line?

zhiyuanliang-ms:
@jimmyca15 @rossgrambo Are we going to include this PR into the next preview release (preview3)?

rossgrambo:
Would be nice- but I don't think it's a blocker.

src/Microsoft.FeatureManagement.AspNetCore/DefaultHttpTargetingContextAccessor.cs
```diff
@@ -0,0 +1,69 @@
...
+using System.Linq;
+using System.Security.Claims;
+using System.Threading.Tasks;
+
+namespace Microsoft.FeatureManagement
+{
+    /// <summary>
+    /// Provides a default implementation of <see cref="ITargetingContextAccessor"/> that creates <see cref="TargetingContext"/> using info from the current HTTP request.
+    /// </summary>
+    public class DefaultHttpTargetingContextAccessor : ITargetingContextAccessor
```
WeihanLi, line 17: Maybe could be `sealed` since there's no member to override, it would be fine from `sealed` to non-sealed but breaking change from non-sealed to `sealed`

```suggestion
    public sealed class DefaultHttpTargetingContextAccessor : ITargetingContextAccessor
```

src/Microsoft.FeatureManagement.AspNetCore/TargetingHttpContextMiddleware.cs
```diff
@@ -18,7 +18,7 @@ public class TargetingHttpContextMiddleware
         private readonly RequestDelegate _next;
         private readonly ILogger _logger;
 
-        private const string TargetingIdKey = $"Microsoft.FeatureManagement.TargetingId";
+        private const string TargetingContextLookup = "FeatureManagement.TargetingContext";
```
zhiyuanliang-ms, line 21: Is there any reason why we removed `Microsoft` prefix here?

zhiyuanliang-ms:
Is there a plan to use this built-in targeting context accessor in our FeatureFlagDemo example? (include the change in this PR or create a new PR after this PR is merged?)

src/Microsoft.FeatureManagement.AspNetCore/TargetingHttpContextMiddleware.cs
rossgrambo: Not that I can recall... I'll add it back.

rossgrambo:
> Is there a plan to use this built-in targeting context accessor in our FeatureFlagDemo example? (include the change in this PR or create a new PR after this PR is merged?)

Yes, I'm planning on a separate to update examples.

jimmyca15:
There has been some developments as we have discussed the middleware and telemetry initializer across different PRs/issues. My expectation at this point is that

* The middleware moves to the same project as the initializer
  * If we stick to http context items cache, the key is project internal
* The default targeting context accessor uses a cache key that is not accessible outside
  * `private object _targetingContextKey = new object();`

As such, I don't expect any changes in the middleware/initializer as part of this PR.

rossgrambo:
> As such, I don't expect any changes in the middleware/initializer as part of this PR.

Agreed. This PR is on hold until the caching rework is completed.

zhiyuanliang-ms:
Open an issue: https://github.com/microsoft/FeatureManagement-Dotnet/issues/432

This PR can help resolve this issue.

src/Microsoft.FeatureManagement.AspNetCore/DefaultHttpTargetingContextAccessor.cs
```diff
@@ -0,0 +1,73 @@
...
+
+namespace Microsoft.FeatureManagement
+{
+    /// <summary>
+    /// Provides a default implementation of <see cref="ITargetingContextAccessor"/> that creates <see cref="TargetingContext"/> using info from the current HTTP request.
+    /// </summary>
+    public sealed class DefaultHttpTargetingContextAccessor : ITargetingContextAccessor
+    {
+        /// <summary>
+        /// The key used to store and retrieve the TargetingContext from the HttpContext items.
```
zhiyuanliang-ms, line 20:
```suggestion
        /// The key used to store and retrieve the <see cref="TargetingContext"/> from the <see cref="HttpContext"/> items.
```

zhiyuanliang-ms (approved):

src/Microsoft.FeatureManagement.Telemetry.ApplicationInsights.AspNetCore/Microsoft.FeatureManagement.Telemetry.ApplicationInsights.AspNetCore.csproj
```diff
@@ -38,6 +38,7 @@
 
   <ItemGroup>
     <ProjectReference Include="..\Microsoft.FeatureManagement\Microsoft.FeatureManagement.csproj" />
```
zhiyuanliang-ms, line 40: Is this reference still necessary?

rossgrambo: Good catch, no. Updated!

src/Microsoft.FeatureManagement.Telemetry.ApplicationInsights.AspNetCore/TargetingTelemetryInitializer.cs
```diff
@@ -37,27 +36,23 @@ protected override void OnInitializeTelemetry(HttpContext httpContext, RequestTe
...
             if (httpContext == null)
             {
                 throw new ArgumentNullException(nameof(httpContext));
             }
 
-            // Extract the targeting id from the http context
-            string targetingId = null;
+            //
+            // Extract the targeting info from the http context
+            TargetingContext targetingContext = httpContext.Items[DefaultHttpTargetingContextAccessor.TargetingContextLookup] as TargetingContext;
```
zhenlan, line 51: Is it intentional to let this throw if the cache doesn't exist? If so, we should throw something more meaningful to users.

src/Microsoft.FeatureManagement.AspNetCore/DefaultHttpTargetingContextAccessor.cs
```diff
@@ -0,0 +1,73 @@
...
+{
+    /// <summary>
+    /// Provides a default implementation of <see cref="ITargetingContextAccessor"/> that creates <see cref="TargetingContext"/> using info from the current HTTP request.
+    /// </summary>
+    public sealed class DefaultHttpTargetingContextAccessor : ITargetingContextAccessor
+    {
+        /// <summary>
+        /// The key used to store and retrieve the <see cref="TargetingContext"/> from the <see cref="HttpContext"/> items.
+        /// </summary>
+        public const string TargetingContextLookup = $"Microsoft.FeatureManagement.TargetingContext";
```
zhenlan, line 22: It's debatable who should own this cache key. It's odd to see the more generic initializer and middleware reference the cache key from a very specific targeting context accessor.

rossgrambo: Agreed... I still am unable to come up with a satisfying place to put this. I suppose the owner could be the middleware- because the middleware's whole goal is to ensure that the `TargetingContext` is available at that key.

src/Microsoft.FeatureManagement.Telemetry.ApplicationInsights.AspNetCore/TargetingTelemetryInitializer.cs
rossgrambo: My undersatnding is that the `as` casting won't throw- it'll either give the requested object or return null. From https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/operators/type-testing-and-cast?redirectedfrom=MSDN#as-operator

I might be ignoring different versions or something where that's the case.

src/Microsoft.FeatureManagement.AspNetCore/DefaultHttpTargetingContextAccessor.cs
jimmyca15: The targeting context accessor doesn't need to share a key.

`private static object _cacheKey = new object();` should work fine.

src/Microsoft.FeatureManagement.Telemetry.ApplicationInsights.AspNetCore/TargetingTelemetryInitializer.cs
zhenlan: I was talking about `httpContext.Items[DefaultHttpTargetingContextAccessor.TargetingContextLookup]`. Will it throw if the cache doesn't exist?

rossgrambo: Oh thank you- you're right. Updated

src/Microsoft.FeatureManagement.AspNetCore/DefaultHttpTargetingContextAccessor.cs
rossgrambo: There's 3 classes that interact with HttpContext.Items for targeting context. `Initializer`, `Middleware`, and the `TargetingContextAccessor`.

`Initializer` depends upon the `Middleware`. The `Middleware` depends upon a `TargetingContextAccessor` existing. The `Middleware` and `Initializer` are tightly coupled- so they definitely need to share a key. The `TargetingContextAccessor` doesn't have to share the same key, but it reduces our HttpContext.Items footprint if it does- and is a clear shared cache location.

So for `Initializer` and `Middleware`, we are going to have a shared key- it could be either an object ref or the string key `"Microsoft.FeatureManagement.TargetingContext"`. I slightly prefer the string key since it's public (on the HttpContext.Items and as a public property). The key needs a home, and if it was just these two classes it would either be on the middleware or a shared constants class.

Now the `DefaultTargetingContextAccessor` comes in- which is our default implementation. It needs a cache key as well. It could either define it's own or used a shared constants class. 

I think the shared constants class makes things the most clear- as FM is defining where `TargetingContext` should live on `HttpContext.Items`. I ended up not liking adding the constants class because it introduces yet another public class, and I ended up going with the `DefaultTargetingContextAccessor`.

rossgrambo:
This PR has been dissolved into two:
* https://github.com/microsoft/FeatureManagement-Dotnet/pull/466 - Add DefaultTargetingContextAccessor to `main`
* https://github.com/microsoft/FeatureManagement-Dotnet/pull/467 - Adjusts targetingId caching to use Activity

Issue #392, opened 2024-03-10
Namespace of ApplicationInsightsTelemetryPublisher


zhenlan:
The namespace of `ApplicationInsightsTelemetryPublisher` is **Microsoft.FeatureManagement.Telemetry.ApplicationInsights** today. It's cumbersome for users to add long namespaces like this. I'd propose we move it under **Microsoft.FeatureManagement.Telemetry**. In the future, if we add any new publishers, we can put them all under the same namespace. This way users don't have to add namespaces for each publisher individually.

jimmyca15:
Sounds reasonable. Will be more discoverable. Shouldn't cause bloat since a user would have to specifically target this package to have this type appear.

rossgrambo:
To clarify- we're just changing the namespace and not the package name- correct? Meaning in the future we would have two packages:

1. `Microsoft.FeatureManagement.Telemetry.ApplicationInsights`
2. `Microsoft.FeatureManagement.Telemetry.LogAnalytics`

Both having:
```csharp
namespace Microsoft.FeatureManagement.Telemetry;
```

Additionally, how do we feel about `Microsoft.FeatureManagement.Telemetry.ApplicationInsights.AspNetCore` ? It holds the `TargetingTelemetryInitializer`. 

Should it follow suit and be:
```csharp
namespace Microsoft.FeatureManagement.Telemetry.AspNetCore;
```
?

jimmyca15:
> Additionally, how do we feel about Microsoft.FeatureManagement.Telemetry.ApplicationInsights.AspNetCore ? It holds the TargetingTelemetryInitializer

I don't think it should change. Here's why. The publisher is an implementation of a generic concept. Any telemetry destination would need it's own publisher. So standardizing to simple namespace allows the implementations to be discoverable easy as Zhenlan mentioned.

But, ITelemetryInitializers are App Insights specific components. I don't think they generalize well.

rossgrambo:
Makes sense. Updated PR.

PR #417, opened 2024-04-04
Adjusts namespace to simply ~.Telemetry

rossgrambo:
## Why this PR?

Addresses https://github.com/microsoft/FeatureManagement-Dotnet/issues/392

## Visible Changes

Developers using any of our offered telemetry publishers will use `using Microsoft.FeatureManagement.Telemetry;` and developers using any of our offered telemetry initializers will use `using Microsoft.FeatureManagement.Telemetry.AspNetCore;`. 

We will no longer have service specific namespaces like `using ~.ApplicationInsights.~;` except for extension methods.

### DIFF
--- a/examples/EvaluationDataToApplicationInsights/Program.cs
+++ b/examples/EvaluationDataToApplicationInsights/Program.cs
@@ -5,7 +5,7 @@ using Microsoft.FeatureManagement.Telemetry;
 using Microsoft.FeatureManagement;
 using EvaluationDataToApplicationInsights;
 using Microsoft.ApplicationInsights.Extensibility;
-using Microsoft.FeatureManagement.Telemetry.AspNetCore;
+using Microsoft.FeatureManagement.Telemetry.ApplicationInsights.AspNetCore;
 
 var builder = WebApplication.CreateBuilder(args);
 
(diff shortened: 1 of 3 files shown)


zhiyuanliang-ms:
Can we also update the .csproj and .sln file?

zhiyuanliang-ms:
Is there any reason why we didn't simplify Microsoft.FeatureManagement.Telemetry.ApplicationInsights.AspNetCore in this PR?

rossgrambo:
> Is there any reason why we didn't simplify Microsoft.FeatureManagement.Telemetry.ApplicationInsights.AspNetCore in this PR?

I had the same question. In the issue I asked and Jimmy had a valid point against it. https://github.com/microsoft/FeatureManagement-Dotnet/issues/392#issuecomment-2040274928

rossgrambo:
> Should we also update the .csproj and .sln file?

We only updated the namespace and not the package. So customers still install the Microsoft.FeatureManagement.Telemetry.ApplicationInsights package- but they only need `using Microsoft.FeatureManagement.Telemetry;`. Because the package name didn't change I don't think those need to be updated- but let me know if there's something I'm missing.

PR #421, opened 2024-04-08
Remove variant & telemetry support for .NET Feature Management schema

zhiyuanliang-ms:
## Why this PR?
In previous versions, the primary schema for the feature management library was the [.NET feature management schema](./schemas/FeatureManagement.Dotnet.v1.0.0.schema.json). Starting from v4.0.0, new features including variants and telemetry will not be supported for the .NET feature management schema. #408 

This PR follows up #414 

## Visible Changes

Variant and telemetry are no longer supported for the .NET Feature Management schema.

The removed fields of feature flag declaration include `Status`, `Variants`, `Allocation`, `Telemetry`

The test suite is also re-organized.


amerjusupovic:
Is there a reason to still pass in values for status, allocation, variants, telemetry to the FeatureDefinition constructor at the end of the ParseDotnet method? It looks like they should be removed, as well as the temporary variables like `Allocation allocation = null` at the beginning.

src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
```diff
@@ -622,14 +557,14 @@ private IEnumerable<IConfigurationSection> GetFeatureDefinitionSections()
...
-                                MicrosoftFeatureManagementFields.FeatureManagementSectionName : 
-                                ConfigurationFields.FeatureManagementSectionName,
+                            _microsoftFeatureManagementSchemaEnabled ?
+                                MicrosoftFeatureManagementFields.FeatureManagementSectionName :
+                                DotnetFeatureManagementFields.FeatureManagementSectionName,
                             StringComparison.OrdinalIgnoreCase));
 
             if (featureManagementConfigurationSection == null)
             {
-                if (RootConfigurationFallbackEnabled && !_microsoftFeatureManagementSchemaEnabled)
```
amerjusupovic, line 632: Can you explain why this condition is removed? I think I missed why it was added originally

src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
```diff
@@ -136,7 +142,46 @@ public async IAsyncEnumerable<FeatureDefinition> GetAllFeatureDefinitionsAsync()
...
+                }
+
+                lock (_lock)
+                {
+                    if (Interlocked.Read(ref _initialized) == 0)
+                    {
+                        _microsoftFeatureManagementSchemaEnabled = hasMicrosoftFeatureManagementSchema;
+
+                        Interlocked.Exchange(ref _initialized, 1);
+                    }
```
amerjusupovic, line 183:
```suggestion
                   if (Interlocked.Exchange(ref _initialized, 1) == 0)
                    {
                        _microsoftFeatureManagementSchemaEnabled = hasMicrosoftFeatureManagementSchema;
                    }
```

I think this achieves the same logic, and does the read/exchange at the same time

amerjusupovic: I think you might also not need the lock in this case

zhiyuanliang-ms:
> Is there a reason to still pass in values for status, allocation, variants, telemetry to the FeatureDefinition constructor at the end of the ParseDotnet method? It looks like they should be removed, as well as the temporary variables like `Allocation allocation = null` at the beginni

src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
zhiyuanliang-ms: We have this method: `IFeatureManagementBuilder AddFeatureManagement(this IServiceCollection services, IConfiguration configuration)` where you can pass a section of configuration and the configurationFeatureDefintionProvider will fall back to the root of configuration if feature management section cannot be found.

We also mention it in our [public doc](https://learn.microsoft.com/en-us/azure/azure-app-configuration/use-feature-flags-dotnet-core#set-up-feature-management)

We are promoting Microsoft schema as primary schema. I think the primary schema should be supported for this usage.

In #319, I supported this root fall back behavior for Microsoft schema. And we have to use the EnsureInit pattern for it.
At that time, the microsoft schema is still under `FeatureManagement` section.
 
In #370, we decided to put microsoft schema under `feature_management` section. I want to get rid of the EnsureInit pattern, so I talked with Jimmy, asked whether we can discard the configuration root fall back behavior for microsoft schema. Jimmy agreed that. So I removed the EnsureInit pattern and add this check: 
`if (RootConfigurationFallbackEnabled && !_microsoftFeatureManagementSchemaEnabled)`
At that time, .NET schema is the primary schema, so I think it is ok to not support this for the secondary microsoft schema.

src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
amerjusupovic: I see, I don't fully understand from the comment but I didn't know it was already agreed on. I'm fine to leave it as is.

src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
```diff
@@ -27,7 +27,9 @@ public sealed class ConfigurationFeatureDefinitionProvider : IFeatureDefinitionP
         private readonly ConcurrentDictionary<string, FeatureDefinition> _definitions;
         private IDisposable _changeSubscription;
         private int _stale = 0;
-        private readonly bool _microsoftFeatureManagementSchemaEnabled;
+        private long _initialized = 0;
+        private bool _microsoftFeatureManagementSchemaEnabled;
```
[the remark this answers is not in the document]
zhiyuanliang-ms: https://github.com/microsoft/FeatureManagement-Dotnet/pull/421#discussion_r1558782830

Now microsoft schema is the primary schema. I think the `AddFeatureManagement(IConfiguration)` method should support it. Then we cannot decide whether microsoft schema is enabled in the constructor because of the `rootConfigurationFallback`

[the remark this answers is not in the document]
jimmyca15: So you're saying it should be removed from the constructor right?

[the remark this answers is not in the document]
zhiyuanliang-ms: Oh, sorry. I forget to remove it.

jimmyca15:
> @jimmyca15 Should we update appsettings.json for all example projects to use Microsoft schema? Currently, I only updated the EvaluationDataToApplicationInsights example.

yes

src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
jimmyca15: @amerjusupovic the difference between your suggestion and what's checked in is that your suggestion can update `_microsoftFeatureManagementSchemaEnabled` even if it's already been set before.

However, there is a valid suggestion similar to what you have which is

```
                   if (Interlocked.CompareExchange(ref _initialized, 1, 0) == 0)
                    {
                        _microsoftFeatureManagementSchemaEnabled = hasMicrosoftFeatureManagementSchema;
                    }
```

It uses CompareExhange to only enter the if statement if the value of `_initialized` was `0`.

jimmyca15: The issue is that `_initialized` should not be set to `1` until `_microsoftFeatureManagementSchemaEnabled` has been assigned to.

jimmyca15 (approved):

Issue #425, opened 2024-04-11
No instance of ITargetingContextAccessor is available for variant assignment.


joacar:
Hi

In my production logs I see this warning message `No instance of ITargetingContextAccessor is available for variant assignment.` from namespace `Microsoft.FeatureManagement.FeatureManager`.

NuGet `Microsoft.FeatureManagment.AspNetCore::3.2.0`

```
// From JetBrains decompiled assembly
[assembly: Extension]
[assembly: CompilationRelaxations(8)]
[assembly: RuntimeCompatibility(WrapNonExceptionThrows = true)]
[assembly: Debuggable(DebuggableAttribute.DebuggingModes.IgnoreSymbolStoreSequencePoints)]
[assembly: TargetFramework(".NETCoreApp,Version=v8.0", FrameworkDisplayName = ".NET 8.0")]
[assembly: AssemblyCompany("Microsoft")]
[assembly: AssemblyConfiguration("Release")]
[assembly: AssemblyCopyright("© Microsoft Corporation. All rights reserved.")]
[assembly: AssemblyDescription("Microsoft.FeatureManagement.AspNetCore provides a way to develop and expose application functionality based on features. Many applications have special requirements when a new feature is developed such as when the feature should be enabled and under what conditions. This library provides a way to define these relationships, and also integrates into common ASP.NET Core code patterns to make exposing these features possible.")]
[assembly: AssemblyFileVersion("4.0.0.0")]
[assembly: AssemblyInformationalVersion("4.0.0-preview2+5c2af2c403c2f1098102a797abb83805379b7652")]
[assembly: AssemblyProduct("Microsoft.FeatureManagement.AspNetCore")]
[assembly: AssemblyTitle("Microsoft.FeatureManagement.AspNetCore")]
[assembly: AssemblyVersion("4.0.0.0")]
```

rossgrambo:
Hello @joacar!

That error is only possible in `4.0.0-preview` and above. I see `4.0.0-preview2` in your log which sounds like your app might have had the FeatureManagement package be updated to the preview version. If this wasn't intentional, you may want to go back to stable `3.2.0`.

That warning means there was no `ITargetingContextAccessor` defined. This means `TargetingContext` can't be automatically determined for the given user. This only matters if you're using TargetingFilters or the preview feature Variants.

Either way, us logging this warning here is misleading, it should happen only during variant assignment. So thank you for reporting it!

joacar:
[...]

PR #426, opened 2024-04-11
Warning log missing targeting context accessor

rossgrambo:
## Why this PR?
Adjusts warning logs identified in https://github.com/microsoft/FeatureManagement-Dotnet/issues/425

## Visible Changes
Developers will no longer see this warning when using flags that don't require TargetingContext.
```
No instance of ITargetingContextAccessor is available for variant assignment.
```
This warning will only show if they have variants defined and we've reached the place in code where allocation would happen.

### DIFF
--- a/src/Microsoft.FeatureManagement/FeatureManager.cs
+++ b/src/Microsoft.FeatureManagement/FeatureManager.cs
@@ -314,11 +314,16 @@ namespace Microsoft.FeatureManagement
                         {
                             string message;
 
-                            if (useContext) {
+                            if (useContext) 
+                            {
                                 message = $"A {nameof(TargetingContext)} required for variant assignment was not provided.";
-                            } else if (TargetingContextAccessor == null) {
+                            } 
+                            else if (TargetingContextAccessor == null) 
+                            {
                                 message = $"No instance of {nameof(ITargetingContextAccessor)} could be found for variant assignment.";
-                            } else {
+                            } 
+                            else 
+                            {
                                 message = $"No instance of {nameof(TargetingContext)} could be found using {nameof(ITargetingContextAccessor)} for variant assignment.";
                             }
 
--- a/src/Microsoft.FeatureManagement/FeatureManager.cs
+++ b/src/Microsoft.FeatureManagement/FeatureManager.cs
@@ -312,13 +312,9 @@ namespace Microsoft.FeatureManagement
                     {
                         if (targetingContext == null)
                         {
-                            if (useContext)
-                            {
-                                Logger?.LogWarning($"The {nameof(TargetingContext)} required for variant assignment was not provided.");
-                            } else
-                            {
-                                Logger?.LogWarning($"The {nameof(TargetingContext)} required for variant assignment was not retrieved from the {nameof(ITargetingContextAccessor)}.");
-                            }
+                            Logger?.LogWarning(useContext ?
+                                $"The {nameof(TargetingContext)} required for variant assignment was not provided." :
+                                $"The {nameof(TargetingContext)} required for variant assignment was not retrieved from the {nameof(ITargetingContextAccessor)}.");
                         }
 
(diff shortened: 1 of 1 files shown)


src/Microsoft.FeatureManagement/FeatureManager.cs
```diff
@@ -310,6 +310,11 @@ private async ValueTask<EvaluationEvent> EvaluateFeature<TContext>(string featur
                     }
                     else
                     {
+                        if (targetingContext == null)
+                        {
+                            Logger?.LogWarning($"No instance of {nameof(TargetingContext)} was explicitly passed or could be found using {nameof(ITargetingContextAccessor)} for variant assignment.");
+                        }
```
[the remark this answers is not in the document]
rossgrambo: Adjusted to:
```
The {nameof(TargetingContext)} required for variant assignment was not provided.
```
and
```
The {nameof(TargetingContext)} required for variant assignment was not retrieved from the {nameof(ITargetingContextAccessor)}.
```

src/Microsoft.FeatureManagement/FeatureManager.cs
```diff
@@ -310,6 +310,17 @@ private async ValueTask<EvaluationEvent> EvaluateFeature<TContext>(string featur
                     }
                     else
                     {
+                        if (targetingContext == null)
+                        {
+                            if (useContext)
```
jimmyca15, line 315:
```
Logger?.LogWarning(useContext ?
    $"The {nameof(TargetingContext)} required for variant assignment was not provided." : 
    $"The {nameof(TargetingContext)} required for variant assignment was not retrieved from the {nameof(ITargetingContextAccessor)}.");
```

src/Microsoft.FeatureManagement/FeatureManager.cs
```diff
@@ -310,6 +310,13 @@ private async ValueTask<EvaluationEvent> EvaluateFeature<TContext>(string featur
                     }
                     else
                     {
+                        if (targetingContext == null)
+                        {
+                            Logger?.LogWarning(useContext ?
+                                $"The {nameof(TargetingContext)} required for variant assignment was not provided." :
+                                $"The {nameof(TargetingContext)} required for variant assignment was not retrieved from the {nameof(ITargetingContextAccessor)}.");
```
jimmyca15, line 317: `The {nameof(TargetingContext)} required for variant assignment was not retrieved from the {nameof(ITargetingContextAccessor)}.`

What happens if the user hasn't even registered an `ITargetingContextAccessor`? Is this the only warning they'll get? If so, that would be confusing.

jimmyca15: `was not retrieved`

That wording is a bit confusing to me. It seems like it could mean all three of these things:

* We didn't even attempt to get it
* The accessor wasn't available
* We used a provided accessor, but it returned no targeting context.

src/Microsoft.FeatureManagement/FeatureManager.cs
```diff
@@ -312,9 +312,17 @@ private async ValueTask<EvaluationEvent> EvaluateFeature<TContext>(string featur
...
                         if (targetingContext == null)
                         {
-                            Logger?.LogWarning(useContext ?
-                                $"The {nameof(TargetingContext)} required for variant assignment was not provided." :
-                                $"The {nameof(TargetingContext)} required for variant assignment was not retrieved from the {nameof(ITargetingContextAccessor)}.");
+                            string message;
+
+                            if (useContext) {
+                                message = $"A {nameof(TargetingContext)} required for variant assignment was not provided.";
+                            } else if (TargetingContextAccessor == null) {
```
jimmyca15, line 319: I don't believe we use `else if` on the same line as a bracket anywhere else in the code base.

Should be like

```
if
{
}
else if ()
{
}
else
{
}

```

jimmyca15 (approved):

Issue #424, opened 2024-04-10
Consider moving `TargetingHttpContextMiddleware` to the App Insights package


rossgrambo:
We could move `TargetingHttpContextMiddleware` to the same package as `TargetingTelemetryInitializer`. The `TargetingHttpContextMiddleware` is not intended to be used without the `TargetingTelemetryInitializer`.

jimmyca15:
If we do that, I'd like to see the library expose an 

`AddXyz`

and 

`UseXyz` pattern.

jimmyca15:
I imagine that taking this change would result in changes in https://github.com/microsoft/FeatureManagement-Dotnet/pull/416

rossgrambo:
The middleware has been reworked to use Activity, which is no longer specific to App Insights. Leaving in the AspNetCore package now makes sense.

Issue #396, opened 2024-03-11
Suggest to have app.UseFeatureManagement


jimmyca15:
Expected

```
app.UseFeatureManagement();
```

Actual
```
app.UseMiddleware<TargetingHttpContextMiddleware>();
```

See [here](https://github.com/microsoft/FeatureManagement-Dotnet/blob/d0c50061a926a0d0548e9e7e0887e79f169a899d/examples/EvaluationDataToApplicationInsights/Program.cs#L60).

zhenlan:
I like it. I was thinking about something similar too.

The `TargetingHttpContextMiddleware` is only useful for ASP.NET targeting telemetry. The name of `UseFeatureManagement` is quite generic. We may want to reserve that name for a more base library though.

jimmyca15:
The middleware is defined in our base ASP.NET Core library, given the extension depends on ASP.NET core, I don't think it could go into any more base of a library.

zhenlan:
You are right about the base library, Jimmy.

As we discussed offline, since `TargetingHttpContextMiddleware` is only useful for 1. targeting scenarios 2. AppInsights telemetry. To be a good citizen, ideally, we can add middleware only when we detect those conditions. Then I have no concerns to wrap the middleware in the generic method `UseFeatureManagement`.

rossgrambo:
I think this middleware can be obsoleted by a TargetingContextAccessor doing the same thing. I spoke a little with Jimmy about the idea. I plan to make a PR and we can discuss further on that. Happy to keep this issue open until that's decided on.

zhenlan:
If I understand your proposal correctly, the problem with it is that the request that calls TrackEvent may never trigger a feature evaluation at all. The heart-vote button is an example. If I misunderstand your proposal, please elaborate.

On the other hand, I'm happy to ship an in-box HttpContext-based TargetingContextAccessor. Please open a separate issue with details of the proposal.

rossgrambo:
Hey @zhenlan & @jimmyca15 , looks like we were hoping to add ASP.NET stuff to the `.AddFeatureManagement()`. Do we still feel that way about the out-of-the-box HttpTargetingContextAccessor?

jimmyca15:
>add ASP.NET stuff to the .AddFeatureManagement()

Adding anything ASP.NET related to the `AddFeatureManagement` method isn't doable since `AddFeatureManagement` is in the base feature management library.

jimmyca15:
[...]

PR #427, opened 2024-04-11
TelemetryInitializer, Middleware, and TargetingContext rework

rossgrambo:
## Why this PR?

Primarily a refactor. 
Also resolves https://github.com/microsoft/FeatureManagement-Dotnet/issues/424. 
Related to- but explicitly not resolving here since the middleware was moved out of the main AspNetCore package https://github.com/microsoft/FeatureManagement-Dotnet/issues/396.

`ITargetingContextAccessor` is async and could not used reached by `ITelemetryInitializer`. We currently get around this by having our middleware add `TargetingContext` to `HttpContext` so it can be read synchronously.

This PR adjusts the middleware to instead place `TargetingId` on the `RequestTelemetry` object instead. This object is specific to Application Insights and we were already adding `TargetingId` to it with our existing initializer. By adding this earlier, we allow the initializer to depend on this field rather than the key on `HttpContext`. It also allows us to simply use the "TargetingId" key, rather than having a custom one for `HttpContext`.

This also allows `ITargetingContextAccessor` to cache however it wants and we're not caching in multiple places.

In short: Middleware stores on RequestTelemetry instead of HttpContext. Initializer now simply moves fields from RequestTelemetry to other Telemetry (like custom events).

## Visible Changes
`HttpContext` will no longer have the key "Microsoft.FeatureManagement.TargetingId" added to `HttpContext.Items`.

### DIFF
--- a/src/Microsoft.FeatureManagement.Telemetry.ApplicationInsights.AspNetCore/Constants.cs
+++ b/src/Microsoft.FeatureManagement.Telemetry.ApplicationInsights.AspNetCore/Constants.cs
@@ -1,0 +1,11 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Microsoft.FeatureManagement.Telemetry.ApplicationInsights.AspNetCore
+{
+    internal class Constants
+    {
+        public const string TargetingIdKey = "TargetingId";
+    }
+}
--- a/src/Microsoft.FeatureManagement.Telemetry.ApplicationInsights.AspNetCore/TargetingHttpContextMiddleware.cs
+++ b/src/Microsoft.FeatureManagement.Telemetry.ApplicationInsights.AspNetCore/TargetingHttpContextMiddleware.cs
@@ -28,14 +28,14 @@ namespace Microsoft.FeatureManagement.Telemetry.ApplicationInsights.AspNetCore
         /// <summary>
         /// Adds targeting information to the HTTP context.
         /// </summary>
-        /// <param name="context">The <see cref="HttpContext"/> to add the targeting information to.</param>
+        /// <param name="httpContext">The <see cref="HttpContext"/> to add the targeting information to.</param>
         /// <param name="targetingContextAccessor">The <see cref="ITargetingContextAccessor"/> to retrieve the targeting information from.</param>
         /// <exception cref="ArgumentNullException">Thrown if the provided context or targetingContextAccessor is null.</exception>
-        public async Task InvokeAsync(HttpContext context, ITargetingContextAccessor targetingContextAccessor)
+        public async Task InvokeAsync(HttpContext httpContext, ITargetingContextAccessor targetingContextAccessor)
         {
-            if (context == null)
+            if (httpContext == null)
             {
-                throw new ArgumentNullException(nameof(context));
+                throw new ArgumentNullException(nameof(httpContext));
             }
 
             if (targetingContextAccessor == null)
@@ -47,11 +47,11 @@ namespace Microsoft.FeatureManagement.Telemetry.ApplicationInsights.AspNetCore
 
             if (targetingContext != null)
             {
-                var requestTelemetry = context.Features.Get<RequestTelemetry>();
+                var requestTelemetry = httpContext.Features.Get<RequestTelemetry>();
 
                 if (requestTelemetry != null)
                 {
-                    requestTelemetry.Properties.Add(Constants.TargetingIdKey, targetingContext.UserId);
+                    requestTelemetry.Properties[Constants.TargetingIdKey] = targetingContext.UserId;
                 }
             }
             else
@@ -59,7 +59,7 @@ namespace Microsoft.FeatureManagement.Telemetry.ApplicationInsights.AspNetCore
                 _logger.LogDebug("The targeting context accessor returned a null TargetingContext");
             }
 
-            await _next(context);
+            await _next(httpContext);
         }
     }
 }
--- a/src/Microsoft.FeatureManagement.Telemetry.ApplicationInsights.AspNetCore/TargetingHttpContextMiddleware.cs
+++ b/src/Microsoft.FeatureManagement.Telemetry.ApplicationInsights.AspNetCore/TargetingHttpContextMiddleware.cs
@@ -2,13 +2,12 @@
 // Licensed under the MIT license.
 //
 
+using Microsoft.ApplicationInsights.DataContracts;
 using Microsoft.AspNetCore.Http;
 using Microsoft.Extensions.Logging;
 using Microsoft.FeatureManagement.FeatureFilters;
-using System;
-using System.Threading.Tasks;
 
-namespace Microsoft.FeatureManagement
+namespace Microsoft.FeatureManagement.Telemetry.ApplicationInsights.AspNetCore
 {
     /// <summary>
     /// Used to add targeting information to HTTP context.
@@ -17,8 +16,6 @@ namespace Microsoft.FeatureManagement
     {
         private readonly RequestDelegate _next;
         private readonly ILogger _logger;
-
-        private const string TargetingIdKey = $"Microsoft.FeatureManagement.TargetingId";
 
         /// <summary>
         /// Creates an instance of the TargetingHttpContextMiddleware
@@ -50,7 +47,12 @@ namespace Microsoft.FeatureManagement
 
             if (targetingContext != null)
             {
-                context.Items[TargetingIdKey] = targetingContext.UserId;
+                var requestTelemetry = context.Features.Get<RequestTelemetry>();
+
+                if (requestTelemetry != null)
+                {
+                    requestTelemetry.Properties.Add(Constants.TargetingIdKey, targetingContext.UserId);
+                }
             }
             else
             {
--- a/src/Microsoft.FeatureManagement.Telemetry.ApplicationInsights.AspNetCore/TargetingTelemetryInitializer.cs
+++ b/src/Microsoft.FeatureManagement.Telemetry.ApplicationInsights.AspNetCore/TargetingTelemetryInitializer.cs
@@ -14,8 +14,6 @@ namespace Microsoft.FeatureManagement.Telemetry.ApplicationInsights.AspNetCore
     /// </summary>
     public class TargetingTelemetryInitializer : TelemetryInitializerBase
     {
-        private const string TargetingIdKey = $"Microsoft.FeatureManagement.TargetingId";
-
         /// <summary>
         /// Creates an instance of the TargetingTelemetryInitializer
         /// </summary>
@@ -42,20 +40,13 @@ namespace Microsoft.FeatureManagement.Telemetry.ApplicationInsights.AspNetCore
                 throw new ArgumentNullException(nameof(httpContext));
             }
 
-            // Extract the targeting id from the http context
-            string targetingId = null;
-
-            if (httpContext.Items.TryGetValue(TargetingIdKey, out object value))
-            {
-                targetingId = value?.ToString();
-            }
-
-            if (!string.IsNullOrEmpty(targetingId))
+            // Extract the targeting id from the request telemetry
+            if (requestTelemetry.Properties.ContainsKey(Constants.TargetingIdKey))
             {
                 // Telemetry.Properties is deprecated in favor of ISupportProperties
                 if (telemetry is ISupportProperties telemetryWithSupportProperties)
                 {
-                    telemetryWithSupportProperties.Properties["TargetingId"] = targetingId;
+                    telemetryWithSupportProperties.Properties[Constants.TargetingIdKey] = requestTelemetry.Properties[Constants.TargetingIdKey];
                 }
             }
         }
--- a/src/Microsoft.FeatureManagement.Telemetry.ApplicationInsights.AspNetCore/TargetingTelemetryInitializer.cs
+++ b/src/Microsoft.FeatureManagement.Telemetry.ApplicationInsights.AspNetCore/TargetingTelemetryInitializer.cs
@@ -35,9 +35,9 @@ namespace Microsoft.FeatureManagement.Telemetry.ApplicationInsights.AspNetCore
                 throw new ArgumentNullException(nameof(telemetry));
             }
 
-            if (httpContext == null)
+            if (requestTelemetry == null)
             {
-                throw new ArgumentNullException(nameof(httpContext));
+                throw new ArgumentNullException(nameof(requestTelemetry));
             }
 
             // Extract the targeting id from the request telemetry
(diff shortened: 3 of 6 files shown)


src/Microsoft.FeatureManagement.Telemetry.ApplicationInsights.AspNetCore/Constants.cs
```diff
@@ -0,0 +1,11 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Microsoft.FeatureManagement.Telemetry.ApplicationInsights.AspNetCore
+{
+    internal class Constants
+    {
+        public const string TargetingIdKey = "TargetingId";
```
zhiyuanliang-ms, line 9: Is there any reason why we discard the prefix "Microsoft.FeatureManagement" we used before?

src/Microsoft.FeatureManagement.Telemetry.ApplicationInsights.AspNetCore/TargetingHttpContextMiddleware.cs
```diff
@@ -50,7 +47,12 @@ public async Task InvokeAsync(HttpContext context, ITargetingContextAccessor tar
 
             if (targetingContext != null)
             {
-                context.Items[TargetingIdKey] = targetingContext.UserId;
+                var requestTelemetry = context.Features.Get<RequestTelemetry>();
+
+                if (requestTelemetry != null)
+                {
+                    requestTelemetry.Properties.Add(Constants.TargetingIdKey, targetingContext.UserId);
+                }
```
zhiyuanliang-ms, line 55: Should we also log something when requestTelemetry is missing?

src/Microsoft.FeatureManagement.Telemetry.ApplicationInsights.AspNetCore/TargetingTelemetryInitializer.cs
```diff
@@ -42,20 +40,13 @@ protected override void OnInitializeTelemetry(HttpContext httpContext, RequestTe
...
-
-            if (!string.IsNullOrEmpty(targetingId))
+            // Extract the targeting id from the request telemetry
+            if (requestTelemetry.Properties.ContainsKey(Constants.TargetingIdKey))
             {
                 // Telemetry.Properties is deprecated in favor of ISupportProperties
                 if (telemetry is ISupportProperties telemetryWithSupportProperties)
                 {
-                    telemetryWithSupportProperties.Properties["TargetingId"] = targetingId;
+                    telemetryWithSupportProperties.Properties[Constants.TargetingIdKey] = requestTelemetry.Properties[Constants.TargetingIdKey];
```
zhiyuanliang-ms, line 49: I think we can use `Add` method here.

This is a code snippet from `AspNetCoreEnvironmentTelemetryInitializer` from `Microsoft.ApplicationInsights.AspNetCore.TelemetryInitializers`
``` C#
public void Initialize(ITelemetry telemetry)
{
    if (environment != null && telemetry is ISupportProperties supportProperties && !supportProperties.Properties.ContainsKey("AspNetCoreEnvironment"))
    {
        supportProperties.Properties.Add("AspNetCoreEnvironment", environment.EnvironmentName);
    }
}
```
Let's follow it.

src/Microsoft.FeatureManagement.Telemetry.ApplicationInsights.AspNetCore/TargetingHttpContextMiddleware.cs
```diff
@@ -50,7 +47,12 @@ public async Task InvokeAsync(HttpContext context, ITargetingContextAccessor tar
 
             if (targetingContext != null)
             {
-                context.Items[TargetingIdKey] = targetingContext.UserId;
+                var requestTelemetry = context.Features.Get<RequestTelemetry>();
+
+                if (requestTelemetry != null)
+                {
+                    requestTelemetry.Properties.Add(Constants.TargetingIdKey, targetingContext.UserId);
```
zhiyuanliang-ms, line 54: Before calling `Add` method, we need to do `ContainsKey` check.

src/Microsoft.FeatureManagement.Telemetry.ApplicationInsights.AspNetCore/Constants.cs
rossgrambo: Before, we were placing it on `HttpContext.Items`. Many different services look at this and store their own things there- which is why it needed the full "Microsoft.FeatureManagement" prefix. 

Now, we're placing it in `RequestTelemetry`. Not only is the scope of this specific to App Insights, but more importantly "TargetingId" was already being placed on it, just later in the code (so this PR isn't introducing a new key). 

It's a reasonable argument to say we should add the prefix to "TargetingId" on all of the outgoing telemetry, which we might do at some point. But that decision should be a separate issue, as it wouldn't just live on "TargetingId" but all of the fields we emit (like "Variant" and "VariantAssignmentReason")

src/Microsoft.FeatureManagement.Telemetry.ApplicationInsights.AspNetCore/TargetingHttpContextMiddleware.cs
rossgrambo: Thats a good idea- will do

src/Microsoft.FeatureManagement.Telemetry.ApplicationInsights.AspNetCore/TargetingTelemetryInitializer.cs
rossgrambo: I think we have the correct usage right now. `.Add` will throw an exception if the key already exists. For our middleware, we really don't expect the key to already exist / it would be weird if it did. For this initializer, it runs on **all** telemetry, meaning it even runs on the RequestTelemetry itself. This means the key will already exist during that call.

src/Microsoft.FeatureManagement.Telemetry.ApplicationInsights.AspNetCore/TargetingHttpContextMiddleware.cs
rossgrambo: I suppose the question to me is- if the targeting id already exists, what should we do? I think there's 3 options:
1. Throw
2. Do nothing
3. Override

Depending on what we want- I think the right functions are:
1. `.Add`
2. `.TryAdd` or `ContainsKey`
3. `[...] = ...`

I could see an argument for all of them. Although I think Override might actually make the most sense to ensure our TargetingId tagging is reliable and if people add the middleware twice it's not breaking.

src/Microsoft.FeatureManagement.Telemetry.ApplicationInsights.AspNetCore/TargetingTelemetryInitializer.cs
```diff
@@ -42,20 +40,13 @@ protected override void OnInitializeTelemetry(HttpContext httpContext, RequestTe
                 throw new ArgumentNullException(nameof(httpContext));
             }
```
zhenlan, line 41: Looks like `httpContext` is not needed anymore.

src/Microsoft.FeatureManagement.Telemetry.ApplicationInsights.AspNetCore/TargetingHttpContextMiddleware.cs
```diff
@@ -50,7 +47,12 @@ public async Task InvokeAsync(HttpContext context, ITargetingContextAccessor tar
 
             if (targetingContext != null)
             {
-                context.Items[TargetingIdKey] = targetingContext.UserId;
+                var requestTelemetry = context.Features.Get<RequestTelemetry>();
```
zhenlan, line 50: NIT: can we rename `context` to `httpContext`? There are so many "contexts". Just make it a little bit easier to read the code :)

src/Microsoft.FeatureManagement.Telemetry.ApplicationInsights.AspNetCore/TargetingHttpContextMiddleware.cs
zhenlan: I don't think the middleware should do anything if the TargetingId is already there. We shouldn't just blindly override it. With that, I wouldn't even call `targetingContextAccessor.GetContextAsync()` if the TargetingId already exists.

src/Microsoft.FeatureManagement.Telemetry.ApplicationInsights.AspNetCore/TargetingTelemetryInitializer.cs
```diff
@@ -42,20 +40,13 @@ protected override void OnInitializeTelemetry(HttpContext httpContext, RequestTe
...
-            string targetingId = null;
-
-            if (httpContext.Items.TryGetValue(TargetingIdKey, out object value))
-            {
-                targetingId = value?.ToString();
-            }
-
-            if (!string.IsNullOrEmpty(targetingId))
+            // Extract the targeting id from the request telemetry
+            if (requestTelemetry.Properties.ContainsKey(Constants.TargetingIdKey))
```
zhenlan, line 44: Should we do null check for `requestTelemetry` too?

src/Microsoft.FeatureManagement.Telemetry.ApplicationInsights.AspNetCore/Constants.cs
zhenlan: I suppose this is the code you refer to
```cs
telemetryWithSupportProperties.Properties[Constants.TargetingIdKey] = requestTelemetry.Properties[Constants.TargetingIdKey];
```

The key on the left doesn't need to be the same key on the right. I agree the key on the left is a bigger discussion, but the key on the right can be just anything. Now that we put the middleware and initializer in the same package, the key on the right can even be something like `static object TargetingIdCacheKey = new object();`.

src/Microsoft.FeatureManagement.Telemetry.ApplicationInsights.AspNetCore/TargetingTelemetryInitializer.cs
zhenlan: We shouldn't override it if someone already set it.

src/Microsoft.FeatureManagement.Telemetry.ApplicationInsights.AspNetCore/TargetingHttpContextMiddleware.cs
rossgrambo: Good call, yes.

src/Microsoft.FeatureManagement.Telemetry.ApplicationInsights.AspNetCore/TargetingTelemetryInitializer.cs
rossgrambo: Adjusted the `httpContext` null check to be a check on `requestTelemetry` instead.

src/Microsoft.FeatureManagement.Telemetry.ApplicationInsights.AspNetCore/TargetingHttpContextMiddleware.cs
rossgrambo: Went ahead and adjusted this to `[...]=...` for now.

src/Microsoft.FeatureManagement.Telemetry.ApplicationInsights.AspNetCore/TargetingTelemetryInitializer.cs
rossgrambo: @zhenlan  If someone already set it, what do you prefer happens in these cases: 
1. Developer is setting TargetingId correctly (I think throw, skip, or override are all valid)
2. Developer is setting TargetingId incorrectly. ie: to a different value than TargetingContext.UserId (I think throw or override).
3. Developer is not setting TargetingId (We add/override)

To me, override is the safest option. If they don't like it they can stop using the initializer.

rossgrambo:
Met with Zhenlan offline. My takeaway is:

Should we be setting TargetingId on RequestTelemetry? 

If the answer is no- then this PR is not the correct way to solve this problem and we should use a cache. 
If the answer is yes- this PR is a valid way to persist the TargetingId to custom events without needing a cache. 

Our current preview release is placing TargetingId on the RequestTelemetry. However Split is not reading RequestTelemetry today, so it's not currently helpful to experimentation to be on there. 

Currently I'm on the fence if we should add it. It's not helpful to split yet- but it could be helpful in App Insights today. If we go with "no" and pursue a caching solution instead, would adding TargetingId to RequestTelemetry 6 months from now make the cache obsolete? If so, our caching solution better not be a breaking change to take away.

zhiyuanliang-ms:
> Should we be setting TargetingId on RequestTelemetry at all?

Comparing to the current way (putting it in httpcontext.items), I will vote for setting TargetingId on RequestTelemetry. This is the recommanded way in App Insights [docs[(https://learn.microsoft.com/en-us/azure/azure-monitor/app/asp-net-core?tabs=netcorenew#enrich-data-through-http).

zhenlan:
I am bit on the fence regarding whether we should add TargetingId to RequestTelemetry because we don't have any out-of-box usage scenario for RequestTelemetry today, but I'm not opposed to it. It could be handy for customers if they want to correlate feature evaluation with request telemetry. Of course, it could be handy for customers if we add TargetingId to all other ASP.NET telemetry (e.g. performance counters). I almost feel we should give customers an option to opt-in/out.

Regardless of whether we add TargetingId to RequestTelemetry (and other ASP.NET telemetry), it doesn't feel right to use RequestTelemetry as a caching mechanism in the middleware. We already have an initializer, and we must add TargetingId to the custom events. Even if we want to add TargetingId to RequestTelemetry (and any other ASP.NET telemetry), we should add it in the initializer.

rossgrambo:
I agree, customers should probably be able to opt out of it. And if that's something we might offer- we would still need the custom events to be populated with it. I think that's a good argument for decoupling these. Going to close this PR.

jimmyca15:
> I agree, customers should probably be able to opt out of it. And if that's something we might offer- we would still need the custom events to be populated with it. I think that's a good argument for decoupling these. Going to close this PR.

This PR moves the middleware into the asp.net core project though. Will you be opening a new PR for that?

rossgrambo:
Yes I will. Going to solidify the shape of the caching first- since it has implications on what projects the middleware depends on. 

For example, with the changes in this PR- it made a lot of sense to place the middleware under App Insights. I want to make sure that's still the case after changes.

PR #434, opened 2024-04-23
Modified FeatureManager to accept optional parameters for FeatureFilters and SessionManagers.

wenkepaul:
I'm working on a project which doesn't use IServiceCollection or C# 9, but would benefit from using this library.  In order to do so, I need to be able to set the FeatureFilters property without using `init`.  This PR adds`Enumerable<IFeatureFilterMetadata>` and `IEnumerable<ISessionManager>` as optional parameters to the FeatureManager constructor and uses their existing `Enumerable.Empty` as a fallback.

### DIFF
--- a/src/Microsoft.FeatureManagement/FeatureManager.cs
+++ b/src/Microsoft.FeatureManagement/FeatureManager.cs
@@ -39,17 +39,21 @@ namespace Microsoft.FeatureManagement
         /// </summary>
         /// <param name="featureDefinitionProvider">The provider of feature flag definitions.</param>
         /// <param name="options">Options controlling the behavior of the feature manager.</param>
+        /// <param name="featureFilters">The collection of feature filter metadata.</param>
+        /// <param name="sessionManagers">The collection of session managers.</param>
         /// <exception cref="ArgumentNullException">Thrown if <paramref name="featureDefinitionProvider"/> is null.</exception>
         public FeatureManager(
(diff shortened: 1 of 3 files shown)


wenkepaul:

jimmyca15:
@wenkepaul can you open up an issue so the problem can be discussed?

wenkepaul:
@jimmyca15 issue [435](https://github.com/microsoft/FeatureManagement-Dotnet/issues/435) has been opened.

zhiyuanliang-ms:
> I need to be able to set the FeatureFilters property without using init.

Do you mean the `IsExternalInit` pattern will not resolve the init accessor issue for .NET Framework?
https://github.com/microsoft/FeatureManagement-Dotnet/blob/main/src/Microsoft.FeatureManagement/IsExternalInit.cs

alexaka1:
@zhiyuanliang-ms For me in a `4.7.2` project that doesn't use the SDK style project, the `IsExternalInit` does not solve the build error.

jimmyca15:
@zhiyuanliang-ms was this superseded by https://github.com/microsoft/FeatureManagement-Dotnet/pull/450 ?

zhiyuanliang-ms:
> @zhiyuanliang-ms was this superseded by #450 ?

@jimmyca15 I think, yes.

@wenkepaul I will close this PR. If you think the current solution still cannot meet your need, feel free to open an issue to track it.

PR #436, opened 2024-04-23
Adds an .editorconfig file, formats, and resolves analyzer issues

rossgrambo:
## Why this PR?

What was done?
1. Added a .editorconfig file to the solution 
  a. This is modified from the default .editorconfig generated by Visual Studio.
3. Ran `dotnet format`
4. Ran Visual Studio .NET code analyzer & manually resolved issues
5. Updated `FeatureManagement` and `FeatureManagement.AspNetCore` with `EnforceCodeStyleInBuild: true`. This means warnings will show in a build when the code style is not being followed.

Adding an `.editorconfig` enables code formatting and code analyzing tools to work. Simply by adding this file, Visual Studio and other IDEs will now respect the style choices for this project when making changes within it. Additionally, this unlocks GitHub actions that can evaluate code style and formatting and catch small mistakes for us.

## Visible Changes

No visible changes

## Philosophy
Code styles should never feel like a hurdle or a headache that we need to adhere to- so at no point should these changes block or slow development. However I believe adding this file allows for a clear definition of rules, a clear place to change style choices as we see fit, and enables IDEs & tooling to quickly follow suit.

This also means all other PRs will be cleaner- as any code style issues will already be caught and handled by tooling.


zhiyuanliang-ms:
This PR is great! This repo has been missing lint checks for a long time. The format check should be added to the azure pipeline.

src/Microsoft.FeatureManagement/AssemblyInfo.cs
```diff
@@ -1,4 +1,4 @@
-﻿﻿// Copyright (c) Microsoft Corporation.
+﻿// Copyright (c) Microsoft Corporation.
```
zhiyuanliang-ms, line 1: I am curious about what is the difference of this change? Which rule triggered it?

.editorconfig
```diff
@@ -0,0 +1,234 @@
...
+
+#### .NET Coding Conventions ####
+
+# Ignored analyzer rules
+dotnet_diagnostic.CA1725.severity = none # Interface implementations should use same parameter names
+
+# Organize usings
+dotnet_separate_import_directive_groups = false
+dotnet_sort_system_directives_first = false
+file_header_template = unset
```
zhiyuanliang-ms, line 26: Can the file header be set to our copyright header?

```
//
```

src/Microsoft.FeatureManagement.AspNetCore/InlineDisabledFeatureHandler.cs
```diff
@@ -11,7 +11,7 @@ namespace Microsoft.FeatureManagement.Mvc
     /// <summary>
     /// A disabled feature handler that wraps an inline handler.
     /// </summary>
-    class InlineDisabledFeaturesHandler : IDisabledFeaturesHandler
+    sealed class InlineDisabledFeaturesHandler : IDisabledFeaturesHandler
```
jimmyca15, line 14: Breaking change.

src/Microsoft.FeatureManagement.AspNetCore/UseForFeatureExtensions.cs
```diff
@@ -22,20 +22,14 @@ public static class UseForFeatureExtensions
         /// <returns></returns>
         public static IApplicationBuilder UseForFeature(this IApplicationBuilder app, string featureName, Action<IApplicationBuilder> configuration)
         {
-            if (app == null)
-            {
-                throw new ArgumentNullException(nameof(app));
-            }
+            ArgumentNullException.ThrowIfNull(app);
 
             if (string.IsNullOrEmpty(featureName))
```
jimmyca15, line 27: Confused why it would suggest Argument.ThrowIfNull for null checks, but not ThrowIfNullOrEmpty for string checks.

jimmyca15:
This is the type of change that is cross-repo. If we were to adopt this, all repos in our scope should adopt it.

src/Microsoft.FeatureManagement.AspNetCore/UseForFeatureExtensions.cs
rossgrambo: I had the same thought. Turns out ThrowIfNull was added in .NET 6 while ThrowIfNullOrEmpty was added in .NET 7. And we still have .NET 6 as a target.

src/Microsoft.FeatureManagement.AspNetCore/InlineDisabledFeatureHandler.cs
rossgrambo: I recalled someone saying sealing classes is a breaking change, but the warning message convinced me it was okay:

[CA1852](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1852)	Type 'InlineDisabledFeaturesHandler' can be sealed because it has no subtypes in its containing assembly and is not externally visible

If we think it's still breaking- or just don't want to risk it I can disable this rule.

.editorconfig
rossgrambo: Yes- good call

src/Microsoft.FeatureManagement.AspNetCore/InlineDisabledFeatureHandler.cs
jimmyca15: Oh it's not a public type. Then indeed it can be sealed without a breaking change.

zhiyuanliang-ms:
Adjusted the [`AnalysisLevel`](https://learn.microsoft.com/en-us/dotnet/core/project-sdk/msbuild-props#analysislevel) in csproj file to suppress CA1848 and CA2254 warnings: 
image

Now, the `AnalysisLevel` for `Microsoft.FeatureManagement` is 5.0 because `Microsoft.FeatureManagement` is targeted on netstandard and 5.0 is the lowest AnalysisLevel which was available for the .NET 5 release is used.

the `AnalysisLevel` for `Microsoft.FeatureManagement.AspNetCore` is 6.0 because `Microsoft.FeatureManagement.AspNetCore` is targeted on .NET 6.0.

zhiyuanliang-ms:
Added `dotnet_diagnostic.IDE0005.severity = warning` to automatically remove unused import.

.editorconfig
zhiyuanliang-ms: Added in [976bbe8](https://github.com/microsoft/FeatureManagement-Dotnet/pull/436/commits/976bbe85538ae39e016907ac94cff0a9fba2180b)

zhiyuanliang-ms:
A reference for all the format rules: https://www.jetbrains.com/help/resharper/EditorConfig_CSHARP_SpacesPageSchema.html#Other

zhiyuanliang-ms:
Move to #472

Issue #447, opened 2024-05-13
Nuget "Microsoft.FeatureManagement.AspNetCore" Version 3.3.0 - Multiple feature filters match the configured filter named 'Microsoft.TimeWindow'


davidstarkcab:
Hello!

After we updated the NuGet package Microsoft.FeatureManagement.AspNetCore to version 3.3.0, we started receiving the following exception: "Multiple feature filters match the configured filter named 'Microsoft.TimeWindow'" whenever we have multiple feature toggles activated with Microsoft.TimeWindow.

This is how we configure it:

```
builder.Services.AddFeatureManagement().AddFeatureFilter<ContextualTargetingFilter>().AddFeatureFilter<TimeWindowFilter>().AddFeatureFilter<PercentageFilter>();

 builder.Configuration.AddAzureAppConfiguration(options =>
 {
     options.Connect(url, credentials).UseFeatureFlags(options =>
     {
         options.Select("*", environment);
         options.CacheExpirationInterval = TimeSpan.FromSeconds(30);
     });

     configurationRefresher = options.GetRefresher();
 });
```

Here is how we encounter the error:

```
await _configurationRefresher.TryRefreshAsync();

TargetingContext targetingContext = new TargetingContext
{
    Groups = Groups,
    UserId = String.Empty
};

var features = new List<FeatureToggle>();

await foreach (var featureName in _featureManagerSnapshot.GetFeatureNamesAsync())
{
    var feature = new FeatureToggle
    {
        Feature = featureName,
        IsEnabled = await _featureManagerSnapshot.IsEnabledAsync(featureName, targetingContext)
    };
    features.Add(feature);
}
```

jimmyca15:
What version did you upgrade from, and is the same setup working with the old version?

zhiyuanliang-ms:
Hi, @davidstarkcab 

I have reproduced the bug. This bug is introduced by 3.3.0. Sorry for the inconvenience.

[...]

PR #448, opened 2024-05-14
Update README to emphasize built-in filters are add automatically and adjust wording

zhiyuanliang-ms:
## Why this PR?
#447 
https://github.com/Azure/azure-sdk-for-java/pull/40093#discussion_r1596958545

Also include a small change to remove an unused reference in this PR

## Visible changes

Update README

Remove an unused reference

### DIFF
--- a/README.md
+++ b/README.md
@@ -533,7 +533,7 @@ If all of three filters are registered:
 
 ## Built-In Feature Filters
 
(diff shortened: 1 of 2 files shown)


rossgrambo:
Should we update the release notes as well to call this a breaking change? Or adjust the AddFeatureFilter to TryAdd instead of Add.

zhenlan:
This is not intended breaking change (I suppose). If so, we should fix it ASAP. Adding it to the release notes will make it sound intentional.

zhiyuanliang-ms:
> This is not intended breaking change (I suppose). If so, we should fix it ASAP. Adding it to the release notes will make it sound intentional.

@zhenlan I have discussed with Jimmy. I will fix it ASAP.

jimmyca15 (approved):

Issue #435, opened 2024-04-23
Can't Set FeatureFilters In FeatureManager In .NET Framework Application


wenkepaul:
I'm working on a .NET Framework 4.6.1 application which doesn't use IServiceCollection and C# 9, but we'd benefit from using this library.  Because this framework targets .NET Standard 2.0 and 2.1, I was hoping to incorporate it into my application.  However I'm unable to set FeatureFilters because this library uses's C# 9's init only setters.  In order to support application which don't use C# 9, I created [PR 434](https://github.com/microsoft/FeatureManagement-Dotnet/pull/434).  In addition to those changes, FeatureManagerSnapshot would also need to made public.

zhiyuanliang-ms:
Are you experiencing a build error?
We have this [class](https://github.com/microsoft/FeatureManagement-Dotnet/blob/main/src/Microsoft.FeatureManagement/IsExternalInit.cs) which should help resolve this init accessor issue. Do you mean that it actually fails to work as expected?

wenkepaul:
Our .NET Framework application uses C# 7.3 and doesn't use IServiceCollection.  So when I try to manually register FeatureManager with our DI container, the application fails to compile with the following code:

```
var featureManager = new FeatureManager {
    FeatureFilters = new List<IFeatureFilterMetadata>()
};
```

alexaka1:
@zhiyuanliang-ms For me in a 4.7.2 project that doesn't use the SDK style project, the IsExternalInit does not solve the build error.

zhiyuanliang-ms:
@wenkepaul @alexaka1 What kind of build error do you have?

Did you have this build error: 

``` C#
error CS8370: Feature 'init-only setters' is not available in C# 7.3. Please use language version 9.0 or greater.
```

I created a .NET framwork 4.7.2 console app and added latest `Microsoft.FeatureManagement` 3.3.0 package to it.

Here are my *Program.cs*:

``` C#
using System;
using System.Threading.Tasks;
using Microsoft.Extensions.Configuration;
using Microsoft.FeatureManagement;
using Microsoft.FeatureManagement.FeatureFilters;

namespace ConsoleApp1
{
    internal class Program
    {
        static async Task Main(string[] args)
        {
            IConfiguration configuration = new ConfigurationBuilder()
                .AddJsonFile("C:\\Users\\zhiyuanliang\\OneDrive - Microsoft\\Desktop\\ConsoleApp1\\appsettings.json")
[...]

PR #450, opened 2024-05-15
Stop using init-only setter for netstandard

zhiyuanliang-ms:
## Why this PR?
#435 

## Visible change

Replace `init` setter to `set`


zhenlan:
Is it possible we can add test to make sure we are netstandard/netfx 4.x compatible?

zhiyuanliang-ms:
> Is it possible we can add test to make sure we are netstandard/netfx 4.x compatible?

Currently, the `TargetFrameworks` is netstand2.0 but the `LangVersion` is 9.0, maybe we should set the `LangVersion` to 7.3.
https://github.com/microsoft/FeatureManagement-Dotnet/blob/main/src/Microsoft.FeatureManagement/Microsoft.FeatureManagement.csproj

zhiyuanliang-ms:
@jimmyca15 @zhenlan 

Maybe we should set the `LangVersion` to 8.0. The things get quite tricky here.
We have an API: `IAsyncEnumerable<string> GetFeatureNamesAsync()` of `IFeatureManager`,

However, async stream is not available in C# 7.3 (need 8.0+), so the `GetFeatureNamesAsync` propably fails to work for .NET framework app now.

alexaka1:
The `await foreach()` syntax is exclusive to 8.0 and up, but you can still use the type manually in 7.3, getting an enumerator and calling next asynchronously in a while loop.

jimmyca15:
Using language version 8.0.0 would ensure we couldn't write init only setters. It also prevents us from taking advantage of new language features but that may be what makes sense since we are supporting netstandard.

src/Microsoft.FeatureManagement/Microsoft.FeatureManagement.csproj
```diff
@@ -15,7 +15,7 @@
     <SignAssembly>true</SignAssembly>
     <DelaySign>false</DelaySign>
     <AssemblyOriginatorKeyFile>..\..\build\Microsoft.FeatureManagement.snk</AssemblyOriginatorKeyFile>
-    <LangVersion>9.0</LangVersion>
+    <LangVersion>8.0</LangVersion>
```
jimmyca15, line 18: I feel it would be valuable to have a comment explaining why we specify v8 and not a later version.

jimmyca15 (approved):

src/Microsoft.FeatureManagement/Microsoft.FeatureManagement.csproj
```diff
@@ -15,7 +15,8 @@
     <SignAssembly>true</SignAssembly>
     <DelaySign>false</DelaySign>
     <AssemblyOriginatorKeyFile>..\..\build\Microsoft.FeatureManagement.snk</AssemblyOriginatorKeyFile>
-    <LangVersion>9.0</LangVersion>
+    <!-- Microsoft.FeatureManagement uses the feature of async streams which is not supported in versions of C# earlier than 8.0. -->
```
jimmyca15, line 18: We should mention why we don't use later than version 8.0

Issue #451, opened 2024-05-16
Consider using `TimeProvider` using Microsoft.Bcl.TimeProvider nuget package


JakeYallop:
There is a comment in `ISystemClock` here:
https://github.com/microsoft/FeatureManagement-Dotnet/blob/31b0efcd8c51d355c1a45a6ca98b483fd71f3f74/src/Microsoft.FeatureManagement/FeatureFilters/ISystemClock.cs#L9-L18

However this is not correct. The runtime provides a down-level package https://www.nuget.org/packages/Microsoft.Bcl.TimeProvider/ that allows `TimeProvider` to be used code using older versions of the runtime, or code that needs to support .NET standard.

zhiyuanliang-ms:
@JakeYallop Thank you very much for paying attention to the latest changes in our repo.

This was a oversight. I didn't know there is `Microsoft.Bcl.TimeProvider`.

 I have sent out a PR to update it. #452

zhiyuanliang-ms:
The PR #452 will reference a new package. We plan to adopt this change with the next major version bump because introducing a new package might lead to potential breaking changes.

PR #452, opened 2024-05-19
Use TimeProvider instead of ISystemClock

zhiyuanliang-ms:
## Why this PR?
#451 

## Visible Changes

Updated `Microsoft.Bcl.AsyncInterfaces` version to the latest one

Used `Microsoft.Bcl.TimeProvider` package. More info can be found [here](https://www.nuget.org/packages/Microsoft.Bcl.TimeProvider/9.0.0-preview.3.24172.9#readme-body-tab)


zhiyuanliang-ms:
This PR will be pending until major version bump.

jimmyca15 (approved):
This project follows semantic versioning, which published a 2.0.0 version describing how to handle updating dependencies [here](https://semver.org/#what-should-i-do-if-i-update-my-own-dependencies-without-changing-the-public-api).

The consensus is that a major version bump is not necessary and, in this case, only a patch version update would be necessary.

Issue #412, opened 2024-03-29
The approach to telemetry (using custom `ITelemetryPublisher`) seems overengineered and counter to modern OpenTelemetry abstractions


julealgon:
I just saw the latest release notes, and with that, I got to know a bit more on how this project is raising telemetry information on feature toggle changes through the use of a custom `ITelemetryPublisher` abstraction.

I'm raising this issue here to discuss on what I believe to be a poor approach to the problem. The OpenTelemetry spec already defines how feature toggles should be emitted in telemetry for both traces and logs here:
- [Semantic Conventions for Feature Flags](https://opentelemetry.io/docs/specs/semconv/feature-flags/)

The way `ITelemetryPublisher` was devised appears to ignore well-known observability abstractions such as `Activity` in favor of a completely bespoke channel that requires custom implementers to communicate with external observability platforms (like the AppInsights implementation that was added).

Instead, what I believe should be happening is that, when telemetry is enabled in the library, a span event is registered in the current activity (as per the OTEL spec). This would naturally flow the information outwards if any `Activity` is in scope, and subsequently using OpenTelemetry, an exporter would pick it up (AppInsights would come into play at this point).

I strongly dislike the current design because it creates a completely separate custom flow for telemetry while we already have a very well established mechanism for capturing it with OpenTelemetry.

Please keep me honest if I'm missing something completely obvious here.

jimmyca15:
@julealgon

There can be an open telemetry implementation of ITelemetryPublisher.

julealgon:
> @julealgon
> 
> There can be an open telemetry implementation of ITelemetryPublisher. I think that should alleviate these concerns as that implementation could use `Activity` as you mentioned.
> 
> Having `ITelemetryPublisher` allows anyone using the feature management library to have an extension point into feature management telemetry without having to migrate to open telemetry.

@jimmyca15 `Activity` is not OpenTelemetry-specific, it's a native .NET type that _can_ integrate with OpenTelemetry.

My argument here is more for using the native `Activity` as the telemetry abstraction instead of devising a completely custom abstraction, than it is for OpenTelemetry per se.

If you take all other major frameworks (from Microsoft itself), they use `Activity` and `Meter`s for telemetry internally, and then a separate OpenTelemetry instrumentation module is created to listen to those and expose them via OTEL. You are just introducing an additional (IMHO completely redundant and dangerous in the sense that it leads people away from the common abstraction) mechanism in this library.

[...]

PR #455, opened 2024-05-30
Telemetry Publishing via Activity & Activity Event

rossgrambo:
## Why this PR?
#412

We now have our own `ActivitySource` and add an `ActivityEvent` on it. The event is essentially the `EvaluationEvent` that was previously passed to the Publisher. 

### Why our own `ActivitySource` instead of using the parent `Activity`?
We could emit the ActivityEvent onto whatever parent Activity is running. Sadly this approach has a few drawbacks:
1.	Listeners who want this event now have to listen to all activities.
2.	The parent code may not have an activity
3.	If a customer implements slow async logic (like reading targeting context from a db), it’s no longer grouped within FeatureManagement.

## Visible Changes
1. Adding publishing uses a different method:
```csharp
    .WithTargeting<HttpContextTargetingContextAccessor>()
    .AddTelemetryPublisher<ApplicationInsightsTelemetryPublisher>();
```

Becomes 

```csharp
    .WithTargeting<HttpContextTargetingContextAccessor>()
    .AddAppInsightsTelemetryPublisher();
```

2. Public types `ApplicationInsightsTelemetryPublisher` & `ITelemetryPublisher` have been removed entirely
3. An Activity and Activity Event are available to listen to. Here's the schemas at a glance:
```javascript
Activity: {
  Name: "Microsoft.FeatureManagement",
  Version: "1.0.0"
}
```

```javascript
ActivityEvent: {
  Tags: {
    "TargetingId", evaluationEvent.TargetingContext?.UserId,
    "FeatureName", evaluationEvent.FeatureDefinition.Name,
    "Variant", evaluationEvent.Variant?.Name,
    "Enabled", evaluationEvent.Enabled,
    "VariantAssignmentReason", evaluationEvent.VariantAssignmentReason,
    "Version", ActivitySource.Version,
    ... // (Any items added to TelemetryMetadata)
  }
}
```


src/Microsoft.FeatureManagement/FeatureManager.cs
```diff
@@ -371,16 +368,53 @@ private async ValueTask<EvaluationEvent> EvaluateFeature<TContext>(string featur
...
+                { "Version", ActivitySource.Version }
+            };
+
+            if (evaluationEvent.FeatureDefinition.Telemetry.Metadata != null)
+            {
+                foreach (KeyValuePair<string, string> kvp in evaluationEvent.FeatureDefinition.Telemetry.Metadata)
+                {
+                    try
+                    {
+                        tags.Add(kvp.Key, kvp.Value);
```
julealgon, line 405: I'd recommend using 'ContainsKey` explicitly instead of relying on the exception from `Add` to log the warning, so we avoid relying on exceptions as control flow even if it is slightly more verbose. Relying on the exception is also substantially slower and could become a problem in critical paths if it keeps throwing for several features.

src/Microsoft.FeatureManagement/FeatureManager.cs
```diff
@@ -371,16 +368,53 @@ private async ValueTask<EvaluationEvent> EvaluateFeature<TContext>(string featur
...
+                    } catch (InvalidOperationException)
+                    {
+                        Logger?.LogWarning($"{kvp.Key} from TelemetryMetadata will be ignored, as it would override an existing key.");
+                    }
+                }
+            }
+
+            ActivityEvent activityEvent = new ActivityEvent("feature_flag", DateTimeOffset.Now, tags);
```
julealgon, line 413: Consider evaluating the `DateTimeOffset.Now` closer to the actual flag check and passing it as an argument so it reflects the actual moment a bit more accurately. By evaluating here, any added delay from the extra processing would end up influencing the recorded time.

Maybe grab it even before the various `ISessionManager` calls (I don't know exactly what that abstraction is responsible for so take with a grain of salt).

julealgon (changes requested):

src/Microsoft.FeatureManagement.Telemetry.ApplicationInsights/FeatureManagementBuilderExtensions.cs
```diff
@@ -0,0 +1,40 @@
...
+
+            if (builder.Services == null)
+            {
+                throw new ArgumentNullException(nameof(builder.Services));
+            }
+
+            if (!builder.Services.Any((ServiceDescriptor d) => d.ServiceType == typeof(IHostedService) && d.ImplementationType == typeof(ApplicationInsightsHostedService)))
+            {
+                builder.Services.Insert(0, ServiceDescriptor.Singleton<IHostedService, ApplicationInsightsHostedService>());
+            }
```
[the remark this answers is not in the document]
zhiyuanliang-ms: Current implementation referes to the open telemetry implementation. [here](https://github.com/open-telemetry/opentelemetry-dotnet/blob/ac0d1d1fab24db657008e853e2b882096c41a974/src/OpenTelemetry.Extensions.Hosting/OpenTelemetryServicesExtensions.cs#L39)

But I am also wondering why open telemetry wants to ensure the hosted service to be the first one.

src/Microsoft.FeatureManagement.Telemetry.ApplicationInsights/Microsoft.FeatureManagement.Telemetry.ApplicationInsights.csproj
```diff
@@ -34,6 +34,7 @@
 
   <ItemGroup>
     <PackageReference Include="Microsoft.ApplicationInsights" Version="2.22.0" />
+    <PackageReference Include="Microsoft.Extensions.Hosting.Abstractions" Version="8.0.0" />
```
zhiyuanliang-ms, line 37: We refers `Microsoft.Extensions.Logging` instead of the abstraction package in `Microsoft.FeatureManagement` project.

src/Microsoft.FeatureManagement/FeatureManager.cs
```diff
@@ -371,16 +368,53 @@ private async ValueTask<EvaluationEvent> EvaluateFeature<TContext>(string featur
...
+        {
+            // Break if there is no activity, this shouldn't happen
+            if (Activity.Current == null)
+            {
+                return;
+            }
+
+            ActivityTagsCollection tags = new ActivityTagsCollection
+            {
+                { "TargetingId", evaluationEvent.TargetingContext?.UserId },
```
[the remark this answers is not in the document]
zhiyuanliang-ms: We need these fields in app insights custom event for some internal usages.

Currenly, the ApplicationInsighsEventPublisher does the following thing
``` C#
foreach (var tag in activityEvent.Tags)
{
    properties[tag.Key] = tag.Value?.ToString();
}
```

src/Microsoft.FeatureManagement.Telemetry.ApplicationInsights/ApplicationInsightsHostedService.cs
```diff
@@ -0,0 +1,48 @@
...
+        }
+
+        /// <summary>
+        /// Constructs the <see cref="ApplicationInsightsEventPublisher"/> which will start listening for events.
+        /// </summary>
+        /// <param name="cancellationToken">The cancellation token.</param>
+        /// <returns>A <see cref="Task"/> representing the asynchronous operation.</returns>
+        public Task StartAsync(CancellationToken cancellationToken)
+        {
+            appInsightsPublisher = new ApplicationInsightsEventPublisher(_telemetryClient);
```
zhiyuanliang-ms, line 30: Is there any reason why we do not follow the opentelemtry's [implementation](https://github.com/open-telemetry/opentelemetry-dotnet/blob/ac0d1d1fab24db657008e853e2b882096c41a974/src/OpenTelemetry.Extensions.Hosting/Implementation/TelemetryHostedService.cs#L44)

This hosted service is aimed to ensure that the `ApplicationInsightsEventPublisher` will be instantiated. Besides, we can take advantage of the DI system to manage the lifecycle of it.

Instead of injecting `TelemetryClient` to the HostedService, IServiceProvider is injected and serviceProvider.GetService<ApplicationInsightsEventPublisher>() is called.

`AddAppInsightsTelemetryPublisher` method should also register `ApplicationInsightsEventPublisher` as singleton.

src/Microsoft.FeatureManagement/FeatureManager.cs
```diff
@@ -262,6 +251,14 @@ private async ValueTask<EvaluationEvent> EvaluateFeature<TContext>(string featur
                 FeatureDefinition = await GetFeatureDefinition(feature).ConfigureAwait(false)
             };
 
+            bool telemetryEnabled = evaluationEvent.FeatureDefinition.Telemetry != null && evaluationEvent.FeatureDefinition.Telemetry.Enabled;
```
zhiyuanliang-ms, line 254: FeatureDefinition can be null here. This is the reason why the testcase failed. If you are referencing a feature flag which doesn't exist in the configuration, FeatureDefinition will be null.

src/Microsoft.FeatureManagement.Telemetry.ApplicationInsights/Microsoft.FeatureManagement.Telemetry.ApplicationInsights.csproj
julealgon: @zhiyuanliang-ms `Microsoft.Extensions.Hosting.Abstractions` is required here because of the hosted service usage. If the project can reference just the `Abstractions` package, that will generate less dependencies for consumers which is a good thing.

zhiyuanliang-ms: I mean, in `Microsoft.FeatureManagement.csproj` we reference`Microsoft.Extensions.Logging` instead of Microsoft.Extensions.Logging.Abstraction`. we can change it to abstraction as well.

julealgon: Ah gotcha. If possible, that would make sense indeed, for the same reason.

src/Microsoft.FeatureManagement.Telemetry.ApplicationInsights/FeatureManagementBuilderExtensions.cs
```diff
@@ -0,0 +1,40 @@
...
+        {
+            if (builder == null)
+            {
+                throw new ArgumentNullException(nameof(builder));
+            }
+
+            if (builder.Services == null)
+            {
+                throw new ArgumentNullException(nameof(builder.Services));
+            }
```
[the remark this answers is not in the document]
jimmyca15: Not sure what you mean by nested argument validation. `builder` is an argument. The method should ensure that `builder` is in a proper state.

One thing I will say is that `ArgumentNullException` is not the write exception because `builder` is not null in this case. It should be an `ArgumentException` that mentions `builder` is invalid because `services` should be non-null.

src/Microsoft.FeatureManagement/FeatureManager.cs
jimmyca15, line 413: always use UtcNow

src/Microsoft.FeatureManagement.Telemetry.ApplicationInsights/FeatureManagementBuilderExtensions.cs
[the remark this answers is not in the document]
julealgon: > Not sure what you mean by nested argument validation. `builder` is an argument. 

`builder` is, but `Services` is not a concern of the method here: it should be ensured/validated when the builder is created. It's more of an SRP thing for me than anything else.

> The method should ensure that `builder` is in a proper state.

Sure, but `builder.Services` is not the argument. If you really want to throw an `ArgumentException`-class exception, it should target `builder` directly because that's the argument.

> One thing I will say is that `ArgumentNullException` is not the write exception because `builder` is not null in this case. It should be an `ArgumentException` that mentions `builder` is invalid because `services` should be non-null.

Agreed. `ArgumentNullException` is misleading here since no argument to the method is `null`.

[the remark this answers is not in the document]
jimmyca15: Yes, the paramName argument of `ArgumentException` should be `builder`, not `Services`.

src/Microsoft.FeatureManagement.Telemetry.ApplicationInsights/ApplicationInsightsEventPublisher.cs
```diff
@@ -0,0 +1,60 @@
...
+        /// Initializes a new instance of the <see cref="ApplicationInsightsEventPublisher"/> class.
+        /// </summary>
+        /// <param name="telemetryClient">The Application Insights telemetry client.</param>
+        public ApplicationInsightsEventPublisher(TelemetryClient telemetryClient)
+        {
+            _telemetryClient = telemetryClient ?? throw new ArgumentNullException(nameof(telemetryClient));
+
+            activityListener = new ActivityListener
+            {
+                ShouldListenTo = (activitySource) => activitySource.Name == "Microsoft.FeatureManagement",
```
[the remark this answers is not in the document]
rossgrambo: Good callout. The problem to me is that they're currently in different packages. One string is in the primary `Microsoft.FeatureManagement` while the other is in `Microsoft.FeatureManagement.Telemetry.ApplicationInsights`. I could expose it as a public or InternalsVisibleToAttribute to share between the packages but it didn't feel valuable enough.

src/Microsoft.FeatureManagement.Telemetry.ApplicationInsights/ApplicationInsightsEventPublisher.cs
```diff
@@ -0,0 +1,60 @@
...
+        private void HandleActivityEvent(ActivityEvent activityEvent)
+        {
+            var properties = new Dictionary<string, string>();
+
+            foreach (var tag in activityEvent.Tags)
+            {
+                properties[tag.Key] = tag.Value?.ToString();
+            }
+
+            _telemetryClient.TrackEvent("FeatureEvaluation", properties);
```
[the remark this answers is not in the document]
rossgrambo: In this case, they're different strings. One is the name of the Activity while the other is the name of the custom event sent to app insights. They happen to be the same for now- but that's not required.

src/Microsoft.FeatureManagement.Telemetry.ApplicationInsights/FeatureManagementBuilderExtensions.cs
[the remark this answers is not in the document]
rossgrambo: I was also curious about this- but the reason is so Telemetry is setup before other IHostedServices. This doesn't guarantee it but it's better than never capturing the telemetry. Here's OpenTelemetry's reasoning: https://github.com/open-telemetry/opentelemetry-dotnet/commit/3e885c77f201daebf5a4c00109425296d0064b06#diff-1a9ddc428e89d201f06e5e2cf75ac6cc23b7a5ba35732cccdeb6c6f5ffc2284cR5

src/Microsoft.FeatureManagement/FeatureManager.cs
```diff
@@ -32,9 +32,13 @@ public sealed class FeatureManager : IFeatureManager, IVariantFeatureManager
         private readonly ConcurrentDictionary<string, ContextualFeatureFilterEvaluator> _contextualFeatureFilterCache;
         private readonly IEnumerable<IFeatureFilterMetadata> _featureFilters;
         private readonly IEnumerable<ISessionManager> _sessionManagers;
-        private readonly IEnumerable<ITelemetryPublisher> _telemetryPublishers;
         private readonly TargetingEvaluationOptions _assignerOptions;
 
+        /// <summary>
+        /// The activity source for feature management.
+        /// </summary>
+        public static readonly ActivitySource ActivitySource = new ActivitySource("Microsoft.FeatureManagement", "1.0.0");
```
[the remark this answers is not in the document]
rossgrambo: Great point, I don't see why it should be exposed. Updated.

src/Microsoft.FeatureManagement/FeatureManager.cs
rossgrambo: Good point- updated.

src/Microsoft.FeatureManagement/FeatureManager.cs
rossgrambo: That would be more accurate for when the actual check started, but I think we'd rather be as close as possible to the return of the final result. Reason being so metrics emitted at or after this timestamp can be expected to use the new result. If we did it at the beginning and their implementation of `ISessionManager` was slow, the timestamps would be less reliable to correlate.

rossgrambo: Although- I could use the end of the activity to get the timestamp I'm referring to, so not sure if that would be as reliable though. But in that same line of thinking, the start of the activity could be the considered the start.

src/Microsoft.FeatureManagement.Telemetry.ApplicationInsights/ApplicationInsightsEventPublisher.cs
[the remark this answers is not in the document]
julealgon: `internal` + `InternalsVisibleTo` would be the best in this case indeed to avoid overexposing it. I see it is a bit convoluted though still, probably not worth it as you said.

src/Microsoft.FeatureManagement.Telemetry.ApplicationInsights/ApplicationInsightsEventPublisher.cs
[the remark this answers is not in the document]
julealgon: Ah fair enough, thanks for clarifying!

src/Microsoft.FeatureManagement.Telemetry.ApplicationInsights/FeatureManagementBuilderExtensions.cs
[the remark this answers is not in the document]
julealgon: If this starts to become such a common concern, we should consider asking for a dedicated ordering API for hosted services in the future.

Forcing a position feels like a hack to me right now. I don't think the `IServiceCollection` API was made to be inspected and manipulated like this.

src/Microsoft.FeatureManagement/FeatureManager.cs
rossgrambo: Thank you- I pulled that out to happen earlier and broke that test in the process. Updated.

julealgon (approved):

src/Microsoft.FeatureManagement.Telemetry.ApplicationInsights/ApplicationInsightsHostedService.cs
rossgrambo: There's a few things to talk about here- one, is that linked OpenTelemetry implementation is actually setting up their `ActivitySources`, not `ActivityListeners`. Their call to setup the listeners is a part of `.WithTracing()`

`WithTracing` creates a few things, but it eventually constructs `TracerProviderSdk` which creates the listeners during construction and explicitly has a shutdown function and a dispose function to handle disposing the listener: https://github.com/open-telemetry/opentelemetry-dotnet/blob/c38cc273bcd03949897d1585e9bc083962cb5f22/src/OpenTelemetry/Trace/TracerProviderSdk.cs#L352

However, we are instead using the `IHostedService` to handle the "shutdown" (OnStop) and dispose options for our ActivityListeners. It appears I forgot this- because at some point after my design I ripped out `IDisposable` on the `IHostedService`- which means in its current state, this would not dispose correctly in a bad exit where OnStop isn't called. 😅

The main question is, who *should* dispose the publisher? The options to me are:
1.  IHostedService. If so- it should dispose it appropriately on either shutdown or when sp disposes the IHostedService. 
2. ServiceProvider. If so- the disposal will happen automatically which is clean. However- this means we also have an IHostedService that stays around in the background for no reason. Stopping the hosted service would do nothing and disposing it does nothing but remove it. There's also no clear way to undo the IHostedService start without doing a full disposal of the entire ServiceProvider.

So in short- it came down to preference. I ended up landing on 1, but there's been a bit of pushback about it, so I'll likely switch over to 2.

julealgon: If we just need some hook to create the instance, could an option 3 using a `IStartupFilter` implementation also work? That interface provides `IServiceProvider`, is executed in the beginning of the pipeline, and doesn't "linger" like the hosted service would.

Contrary to HostedService, it also executes before the application actually starts listening to anything, which might actually be a positive here.

rossgrambo: `IStartupFilter` would be a great fit here- the issue is that it's only available in ASP.NET Core. Currently this telemetry publishing is possible without any ASP.NET dependency.

julealgon: Ah you are absolutely right @rossgrambo , I honestly completely forgot about that important detail when suggesting it here 😅.

julealgon: While searching for alternatives to this, I stumbled upon `IHostedLifecycleService`, which appears to provide something similar to what the original code on the OpenTelemetry repo tried to do.
- https://www.stevejgordon.co.uk/introducing-the-new-ihostedlifecycleservice-interface-in-dotnet-8

> The StartingAsync method for all registered hosted services that implement this interface will run very early in the application lifecycle before StartAsync (from the IHostedService) is called on any registered hosted services. This may be used to perform some very early validation checks before an application is started, such as checking for critical requirements or dependencies being available. This allows an application to potentially fail startup before any hosted services have begun executing their primary workload. **Other uses include “pre-heating” and initialising singletons and other state the application uses.**

(emphasis mine) 

I imagine that by leveraging the `StartingAsync` method on this more specific interface would have the same effect as this hack:
```csharp
builder.Services.Insert(0, ServiceDescriptor.Singleton<IHostedService, ApplicationInsightsHostedService>());
```

Since `StartingAsync` actually fires before every `StartAsync` call on every other hosted service registered in the application.

This doesn't answer who exactly should perform the disposal, but might still make sense for this particular initialization requirement here since we can more cleanly guarantee it will run before anything else.

src/Microsoft.FeatureManagement/FeatureManager.cs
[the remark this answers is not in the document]
rossgrambo: Turns out that Application insights is ignoring null or empty fields- so we actually have no reason to emit this. Adjusted to optionally emit the field.

src/Microsoft.FeatureManagement/FeatureManager.cs
```diff
@@ -371,16 +368,53 @@ private async ValueTask<EvaluationEvent> EvaluateFeature<TContext>(string featur
...
+            if (Activity.Current == null)
+            {
+                return;
+            }
+
+            ActivityTagsCollection tags = new ActivityTagsCollection
+            {
+                { "TargetingId", evaluationEvent.TargetingContext?.UserId },
+                { "FeatureName", evaluationEvent.FeatureDefinition.Name },
+                { "Variant", evaluationEvent.Variant?.Name },
```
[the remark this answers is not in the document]
rossgrambo: Adjusted to be optional.

src/Microsoft.FeatureManagement/FeatureManager.cs
```diff
@@ -388,14 +388,22 @@ private void AddEvaluationActivityEvent(EvaluationEvent evaluationEvent)
...
+            if (evaluationEvent.TargetingContext?.UserId != null)
+            {
+                tags["TargetingId"] = evaluationEvent.TargetingContext.UserId;
+            }
+
+            if (evaluationEvent.Variant?.Name != null)
+            {
+                tags["Variant"] = evaluationEvent.Variant.Name;
+            }
+
```
julealgon, line 406: I'd suggest using pattern matching for these as well:
```suggestion
            if (evaluationEvent.TargetingContext is { UserId: string userId })
            {
                tags["TargetingId"] = userId;
            }

            if (evaluationEvent.Variant is { Name: string variantName })
            {
                tags["Variant"] = variantName;
            }

```

julealgon (changes requested):

src/Microsoft.FeatureManagement.Telemetry.ApplicationInsights/ApplicationInsightsHostedService.cs
rossgrambo: Unfourtunatly it appears the `IHostedLifecycleService` is limited to .NET 8 and above. We're still supporting down to .NET 7.

julealgon: @rossgrambo I assume you meant that it is limited to `Microsoft.Extensions.Hosting.Abstractions` `v8`+. This doesn't seem to be tied to the framework but to the library itself. Supporting it would be a matter of bumping the dependency on `Microsoft.FeatureManagement.Telemetry.ApplicationInsights`.

Regardless, I'm curious where you found that. The page for the interface says it is supported since v6:
https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.hosting.ihostedlifecycleservice?view=net-8.0#applies-to

Can you elaborate?

rossgrambo: Oh I may have been mislead by the article then. The [article linked earlier](https://www.stevejgordon.co.uk/introducing-the-new-ihostedlifecycleservice-interface-in-dotnet-8)  mentions it a couple times as a "new thing in .NET 8". Even the title says it's new in .NET 8. When searching for it- I saw the same mention of .NET 8 in [other articles too](https://www.roundthecode.com/dotnet-tutorials/hosted-service-major-update-lifecycle-events).

But I just implemented it and ran it successfully with .net 6 and net48. So I suppose it's available?

rossgrambo: Oh okay I found the issue. The "Starting" and "Stopping" methods aren't called unless the project is running >.NET8. 

I could branch the logic on which .NET we're in- but I think I'll just stick with the `HostedService` `Start` for now- unless there's a good way around this.

rossgrambo: *Updated to use service provider here as @zhiyuanliang-ms suggested

src/Microsoft.FeatureManagement/FeatureManager.cs
```diff
@@ -371,16 +368,62 @@ private async ValueTask<EvaluationEvent> EvaluateFeature<TContext>(string featur
...
+                tags["Variant"] = variantName;
+            }
+
+            if (evaluationEvent.FeatureDefinition.Telemetry.Metadata is { } metadata)
+            {
+                foreach (KeyValuePair<string, string> kvp in metadata)
+                {
+                    if (tags.ContainsKey(kvp.Key))
+                    {
+                        Logger?.LogWarning($"{kvp.Key} from TelemetryMetadata will be ignored, as it would override an existing key.");
```
zhiyuanliang-ms, line 413:
```suggestion
                        Logger?.LogWarning($"{kvp.Key} from telemetry metadata will be ignored, as it would override an existing key.");
```

src/Microsoft.FeatureManagement.Telemetry.ApplicationInsights/FeatureManagementBuilderExtensions.cs
```diff
@@ -0,0 +1,42 @@
...
+    public static class FeatureManagementBuilderExtensions
+    {
+        /// <summary>
+        /// Adds the <see cref="ApplicationInsightsEventPublisher"/> using <see cref="ApplicationInsightsHostedService"/> to the feature management builder.
+        /// </summary>
+        /// <param name="builder">The feature management builder.</param>
+        /// <returns>The feature management builder.</returns>
+        public static IFeatureManagementBuilder AddAppInsightsTelemetryPublisher(this IFeatureManagementBuilder builder)
+        {
+            if (builder == null)
```
zhiyuanliang-ms, line 22: Could you add these null check for the `WithTargeting` extension method?

tests/Tests.FeatureManagement/FeatureManagementTest.cs
```diff
@@ -1332,6 +1332,8 @@ public async Task VariantBasedInjection()
 
     public class FeatureManagementTelemetryTest
     {
+        private readonly AutoResetEvent _waitHandle = new AutoResetEvent(false);
```
zhiyuanliang-ms, line 1335: I didn't understand the reason why we force the thread to wait until the `ActivityStopped` callback is done.

Each time `featureManagemr.IsEnabledAsync` method is called, a new activity "FeatureEvaluation" will be started and only one activity event "feature_flag" will be added to the activity. Any execution of the callback is bounded to a specific activity.

src/Microsoft.FeatureManagement.Telemetry.ApplicationInsights/ApplicationInsightsEventPublisher.cs
```diff
@@ -0,0 +1,60 @@
...
+            _activityListener.Dispose();
+        }
+
+        private void HandleActivityEvent(ActivityEvent activityEvent)
+        {
+            var properties = new Dictionary<string, string>();
+
+            foreach (var tag in activityEvent.Tags)
+            {
+                properties[tag.Key] = tag.Value?.ToString();
```
zhiyuanliang-ms, line 54: Do you think it's necessary for us to add some additional checks here to ensure that certain fields we need internally (for example "VariantAssignmentReason") will definitely appear in the app insights event? Currently, if the activity event doesn't have these tags, then the property of the app insights event won't have these fields.

src/Microsoft.FeatureManagement.Telemetry.ApplicationInsights/ApplicationInsightsHostedService.cs
julealgon: > Oh okay I found the issue. The StartingAsync and StoppingAsync methods aren't called unless the project is running >= .NET8.

Ohhh that is very weird! Documentation on this interface should be updated to reflect this limitation I think!

Sorry for leading you down this path @rossgrambo , I had no idea.

src/Microsoft.FeatureManagement.Telemetry.ApplicationInsights/FeatureManagementBuilderExtensions.cs
rossgrambo: Let's do another PR- because I'm not even sure this null check makes sense. We're using an extension parameter, which I believe means this method couldn't even be called unless the instance existed.

tests/Tests.FeatureManagement/FeatureManagementTest.cs
rossgrambo: The main reason is that we need the test code to wait until all the listener handlers successfully run. Otherwise it will kick off all of the activities and finish immediately without waiting for proof that the listener succeeded. 

The quickest option would be a Wait(13) at the end to ensure we triggered this 13 (# of tests) times. Or some other waiting that would break once all tests have run.

I ended up preferring this one by one approach. It makes it easier to debug, easier to add more tests, and is pretty clear what's going on at each step. Surprisingly, this test is just as fast as our other tests so I didn't bother trying to speed it up.

src/Microsoft.FeatureManagement.Telemetry.ApplicationInsights/ApplicationInsightsEventPublisher.cs
rossgrambo: I think no. In the case where the event isn't correct, we should still emit the event because it's still telemetry and this is the publisher for telemetry. For example, if FM is used without the provider, etag won't exist which is required for experimentation, but not required for telemetry.

tests/Tests.FeatureManagement/FeatureManagementTest.cs
zhiyuanliang-ms: > Otherwise it will kick off all of the activities and finish immediately.

This doesn't sound reasonable to me. Is there any documentation about how `ActivityStopped` callback works?

zhiyuanliang-ms (approved):

tests/Tests.FeatureManagement/FeatureManagementTest.cs
rossgrambo: I'm incorrect- I thought the listeners were handled async- or at least not immediately at the end of the Activity. Turns out it is a synchronous callback that does happen right away. 

But still- I think it's fairly valuable in these tests because nothing is actually blocked unless the test fails. It also ensures the ActivityListener.OnActivityStopped is executed exactly once per call. Without doing a check like this, a typo might mean we don't run the OnActivityStopped callback at all and the tests would still pass (because no Assert failed).

We could replace this with a bool flipping around like "wasCallbackCalled".

> Is there any documentation about how ActivityStopped callback works?

I don't see great docs about it besides a shallow API reference and some random articles.

src/Microsoft.FeatureManagement/FeatureManager.cs
```diff
@@ -371,16 +368,62 @@ private async ValueTask<EvaluationEvent> EvaluateFeature<TContext>(string featur
...
                 }
             }
 
             return evaluationEvent;
         }
 
+        private void AddEvaluationActivityEvent(EvaluationEvent evaluationEvent)
+        {
+            // Break if there is no reason to emit an event
+            if (Activity.Current == null || !Activity.Current.IsAllDataRequested)
```
jimmyca15, line 384: No need to call this method at all if this is the case. Typically better to avoid inserting immediate exit control-flow in a method that way at the call-site a method like `AddXyz` can always be expected to add something.

src/Microsoft.FeatureManagement/FeatureManager.cs
```diff
@@ -371,16 +368,62 @@ private async ValueTask<EvaluationEvent> EvaluateFeature<TContext>(string featur
...
+            {
+                return;
+            }
+
+            ActivityTagsCollection tags = new ActivityTagsCollection
```
jimmyca15, line 389: This is a good place to use `var`, the type of `tags` is clear on the other side of the assignment.


```suggestion
            var tags = new ActivityTagsCollection
```

julealgon: You might actually want to use target-typed new here instead:
```suggestion
            ActivityTagsCollection tags = new()
```

src/Microsoft.FeatureManagement/FeatureManager.cs
```diff
@@ -371,16 +368,62 @@ private async ValueTask<EvaluationEvent> EvaluateFeature<TContext>(string featur
...
+            {
+                { "FeatureName", evaluationEvent.FeatureDefinition.Name },
```
jimmyca15, line 391: Can you

```
Debug.Assert(evaluationEvent != null);
Debug.Assert(evaluationEvent.FeatureDefinition != null);
```

on entry to help bring awareness that these are known to be non-null.

src/Microsoft.FeatureManagement/FeatureManager.cs
jimmyca15: Even better.

src/Microsoft.FeatureManagement/FeatureManager.cs
julealgon: Interesting perspective on this. What do you do when the method is called multiple times, and the condition is nontrivial (like this one)?

By not including the condition in the method itself, you not only increase duplication (by replicating the check) but also arguably make SRP worse by splitting a single "thing" into 2 places.

src/Microsoft.FeatureManagement/FeatureManager.cs
```diff
@@ -371,16 +368,62 @@ private async ValueTask<EvaluationEvent> EvaluateFeature<TContext>(string featur
...
+            {
+                tags["TargetingId"] = userId;
+            }
+
+            if (evaluationEvent.Variant is { Name: string variantName })
+            {
+                tags["Variant"] = variantName;
+            }
+
+            if (evaluationEvent.FeatureDefinition.Telemetry.Metadata is { } metadata)
```
jimmyca15, line 407: Why do we need anything more than a null-check here? We already know the type of `Metadata`.

src/Microsoft.FeatureManagement/FeatureManager.cs
```diff
@@ -371,16 +368,62 @@ private async ValueTask<EvaluationEvent> EvaluateFeature<TContext>(string featur
...
+            {
+                foreach (KeyValuePair<string, string> kvp in metadata)
+                {
+                    if (tags.ContainsKey(kvp.Key))
+                    {
+                        Logger?.LogWarning($"{kvp.Key} from telemetry metadata will be ignored, as it would override an existing key.");
+                    }
+                    else
```
jimmyca15, line 415: It's common to have validation on elements in a loop. The `else` can be avoided by having a continue on elements that don't pass validation.

src/Microsoft.FeatureManagement/FeatureManager.cs
jimmyca15: In my opinion this condition is trivial. If the condition wasn't trivial an extension method may be helpful to check state.

I don't worry about duplication, I worry more about clarity.

src/Microsoft.FeatureManagement/FeatureManager.cs
```diff
@@ -371,16 +368,62 @@ private async ValueTask<EvaluationEvent> EvaluateFeature<TContext>(string featur
...
+
+            ActivityTagsCollection tags = new ActivityTagsCollection
+            {
+                { "FeatureName", evaluationEvent.FeatureDefinition.Name },
+                { "Enabled", evaluationEvent.Enabled },
+                { "VariantAssignmentReason", evaluationEvent.VariantAssignmentReason },
+                { "Version", ActivitySource.Version }
+            };
+
+            if (evaluationEvent.TargetingContext is { UserId: string userId })
```
jimmyca15, line 397: Type checking is unnecessary, we already know the type.

src/Microsoft.FeatureManagement.Telemetry.ApplicationInsights/ApplicationInsightsEventPublisher.cs
```diff
@@ -0,0 +1,60 @@
...
+            {
+                ShouldListenTo = (activitySource) => activitySource.Name == "Microsoft.FeatureManagement",
+                Sample = (ref ActivityCreationOptions<ActivityContext> options) => ActivitySamplingResult.AllData,
+                ActivityStopped = (activity) =>
+                {
+                    ActivityEvent? evaluationEvent = activity.Events.FirstOrDefault((activityEvent) => activityEvent.Name == "feature_flag");
+
+                    if (evaluationEvent != null && evaluationEvent.Value.Tags.Any())
+                    {
+                        HandleActivityEvent(evaluationEvent.Value);
```
jimmyca15, line 32:
```suggestion
                        HandleFeatureFlagEvent(evaluationEvent.Value);
```

src/Microsoft.FeatureManagement.Telemetry.ApplicationInsights/ApplicationInsightsEventPublisher.cs
```diff
@@ -0,0 +1,60 @@
...
+
+            _activityListener = new ActivityListener
+            {
+                ShouldListenTo = (activitySource) => activitySource.Name == "Microsoft.FeatureManagement",
+                Sample = (ref ActivityCreationOptions<ActivityContext> options) => ActivitySamplingResult.AllData,
+                ActivityStopped = (activity) =>
+                {
+                    ActivityEvent? evaluationEvent = activity.Events.FirstOrDefault((activityEvent) => activityEvent.Name == "feature_flag");
+
+                    if (evaluationEvent != null && evaluationEvent.Value.Tags.Any())
```
jimmyca15, line 30:
```suggestion
                    if (evaluationEvent.HasValue && evaluationEvent.Value.Tags.Any())
```

It's functionally the same, that's just the model of nullables.

src/Microsoft.FeatureManagement.Telemetry.ApplicationInsights/FeatureManagementBuilderExtensions.cs
```diff
@@ -0,0 +1,42 @@
...
+    /// Extensions used to add feature management functionality.
+    /// </summary>
+    public static class FeatureManagementBuilderExtensions
+    {
+        /// <summary>
+        /// Adds the <see cref="ApplicationInsightsEventPublisher"/> using <see cref="ApplicationInsightsHostedService"/> to the feature management builder.
+        /// </summary>
+        /// <param name="builder">The feature management builder.</param>
+        /// <returns>The feature management builder.</returns>
+        public static IFeatureManagementBuilder AddAppInsightsTelemetryPublisher(this IFeatureManagementBuilder builder)
```
jimmyca15, line 20:
```suggestion
        public static IFeatureManagementBuilder AddApplicationInsightsTelemetryPublisher(this IFeatureManagementBuilder builder)
```

tests/Tests.FeatureManagement/FeatureManagementTest.cs
```diff
@@ -1342,137 +1344,168 @@ public async Task TelemetryPublishing()
...
-            // Test a feature with telemetry disabled
-            bool result = await featureManager.IsEnabledAsync(Features.OnTestFeature, cancellationToken);
-
-            Assert.True(result);
-            Assert.Null(testPublisher.evaluationEventCache);
-
-            // Test telemetry cases
-            result = await featureManager.IsEnabledAsync(Features.OnTelemetryTestFeature, cancellationToken);
+            // Start listener
+            ActivityListener activityListener = new ActivityListener
```
jimmyca15, line 1355: Is that disposable?

tests/Tests.FeatureManagement/FeatureManagementTest.cs
jimmyca15: > a typo (like the wrong activity name) might mean we don't run the OnActivityStopped callback at all and the tests would still pass (because no Assert failed).

Define a few booleans (e.g. `bool ranTestOne = false;`) and make sure in the callback to set them to true, at the end of the test assert they were flipped to assert that the callbacks were executed.

Shouldn't be any reason to have the synchronization primitive.

src/Microsoft.FeatureManagement/FeatureManager.cs
julealgon: > Why do we need anything more than a null-check here? 

We don't. That line _is_ an inverted null check, but it extracts the value into the `metatada` variable so you don't need to repeat the entire thing again inside the block.

> We already know the type of `Metadata`.

Yes, which is why we use `is { } metadata` instead of `is TheTypeOfTheMetadata metadata`. Using `{ }` is a way to avoid needing to specify the type.

It would basically be equivalent to:
```csharp
if (evaluationEvent.FeatureDefinition.Telemetry.Metadata is not null and var metadata)
```

jimmyca15: `is` is good for type checking. I'd avoid introducing usage of `is` here now to replace basic null checks.

src/Microsoft.FeatureManagement/FeatureManager.cs
julealgon: I believe you are missing the point of pattern matching extractions.

tests/Tests.FeatureManagement/FeatureManagementTest.cs
rossgrambo: After the change, I noticed there was a small bug here where the ActivityListener would be triggered from other tests running async (and using the same features). Added a thread check to mitigate that.

src/Microsoft.FeatureManagement/FeatureManager.cs
jimmyca15: We have no need to introduce new patterns for null checking at this point. It will be inconsistent with all other null checks in the library.

src/Microsoft.FeatureManagement/FeatureManager.cs
jimmyca15: sounds good

src/Microsoft.FeatureManagement/FeatureManager.cs
rossgrambo: Works for me. We should probably let a lint rule decide which pattern to use for the library anyway.

tests/Tests.FeatureManagement/FeatureManagementTest.cs
jimmyca15: Parent activity id should be available for this.

jimmyca15 (approved):

jimmyca15:
Thanks for the thorough review on this @julealgon !

tests/Tests.FeatureManagement/FeatureManagementTest.cs
rossgrambo: Updated to use parent id.

src/Microsoft.FeatureManagement/FeatureManager.cs
```diff
@@ -371,16 +368,61 @@ private async ValueTask<EvaluationEvent> EvaluateFeature<TContext>(string featur
...
+                        Logger?.LogWarning($"{kvp.Key} from telemetry metadata will be ignored, as it would override an existing key.");
+
+                        continue;
+                    }
+
+                    tags[kvp.Key] = kvp.Value;
+                }
+            }
+
+            var activityEvent = new ActivityEvent("feature_flag", DateTimeOffset.UtcNow, tags);
```
jimmyca15, line 421: Why does this use snake casing instead of pascal like all the other fields?

rossgrambo: This is following OpenTelemetry's guide for the naming and I didn't feel strongly to deviate. It made more sense when we were emitting both OpenTelemetry and FeatureManagement fields, but I still don't feel we need to deviate.

jimmyca15: This would be the only place where any naming convention is used, and the naming deviates from the casing of all of our tags. Seems that it should be `FeatureFlag`

rossgrambo: Makes sense. PR: https://github.com/microsoft/FeatureManagement-Dotnet/pull/464

PR #466, opened 2024-06-20
Adds Default Targeting Accessor and extension method WithTargeting

rossgrambo:
## Why this PR?

1. Addresses issues 
  a. https://github.com/microsoft/FeatureManagement-Dotnet/issues/415
2. Updates FeatureFlagDemo example to use the new extension

## Visible Changes

Developers using ASP.NET will now have a new extension method available to them, `WithTargeting()`. Previously, they could only use `WithTargeting<ITargetingContextAccessor>()`

```csharp
    .WithTargeting()
```

This default accessor will extract the targeting info from `HttpContext.User`. `Groups` will be extracted from claims of type `Role`. `UserId` is taken from the `Identity.Name` field.


src/Microsoft.FeatureManagement.AspNetCore/AspNetCoreFeatureManagementBuilderExtensions.cs
```diff
@@ -44,5 +47,28 @@ public static IFeatureManagementBuilder UseDisabledFeaturesHandler(this IFeature
 
             return builder;
         }
+
+        /// <summary>
+        /// Adds the <see cref="DefaultHttpTargetingContextAccessor"/> to be used for targeting and registers the targeting filter to the feature management system.
+        /// </summary>
+        /// <param name="builder">The <see cref="IFeatureManagementBuilder"/> used to customize feature management functionality.</param>
+        /// <returns>A <see cref="IFeatureManagementBuilder"/> that can be used to customize feature management functionality.</returns>
+        public static IFeatureManagementBuilder WithTargeting(this IFeatureManagementBuilder builder)
```
zhiyuanliang-ms, line 56: Should we call `AddHttpContextAccessor` in this method since our default targeting accessor depends on it.

zhiyuanliang-ms:
This is a new feature. After this is released, we need to update our public doc to introduce it.

src/Microsoft.FeatureManagement.AspNetCore/AspNetCoreFeatureManagementBuilderExtensions.cs
rossgrambo: That's a great question. I originally thought no- so we're in parity with what we have today. However it would remove another hurdle to get started so I think it's a good idea.

src/Microsoft.FeatureManagement.AspNetCore/DefaultHttpTargetingContextAccessor.cs
```diff
@@ -0,0 +1,73 @@
...
+            // Treat user identity name as user id
+            ClaimsPrincipal user = httpContext.User;
+
+            string userId = user?.Identity?.Name;
+
+            //
+            // Treat claims of type Role as groups
+            IEnumerable<string> groups = httpContext.User.Claims
+                .Where(c => c.Type == ClaimTypes.Role)
+                .Select(c => c.Value);
```
zhenlan, line 58: Should we force the query evaluation here?

rossgrambo: Ideally no- do you have an idea for how to avoid it?

rossgrambo: I misunderstood this comment. I think you're right, and we should force evaluation. Adjusting this to:

```csharp
IEnumerable<string> groups = httpContext.User.Claims
    .Where(c => c.Type == ClaimTypes.Role)
    .Select(c => c.Value)
    .ToList();
```

src/Microsoft.FeatureManagement.AspNetCore/AspNetCoreFeatureManagementBuilderExtensions.cs
jimmyca15: I'd like to decouple that.

By the way, is this method name the one that we discussed so many options and couldn't settle on one?

src/Microsoft.FeatureManagement.AspNetCore/DefaultHttpTargetingContextAccessor.cs
zhenlan: Sounds good.

src/Microsoft.FeatureManagement.AspNetCore/AspNetCoreFeatureManagementBuilderExtensions.cs
rossgrambo: Okay, let's keep it decoupled. Could always do that in a separate PR anyway.

> By the way, is this method name the one that we discussed so many options and couldn't settle on one?

Yes it is. The last design review it was discussed up was the caching meeting last month. We lightly aligned on `.WithTelemetry<DefaultTargetingContextAccessor>()`, but some smaller discussion afterwards lead to `.WithTelemetry()` so we didn't have to make `DefaultTargetingContextAccessor` public.

Although I'm having trouble finding a paper trail to confirm that we did commit to this.

rossgrambo: @zhenlan do you have thoughts on the `.WithTelemetry()` over `.WithTelemetry<DefaultTargetingContextAccessor>()`

zhenlan: Do you mean WithTargeting?

jimmyca15: Yes, this would have been referring to `WithTargeting`, not `WithTelemetry`

jimmyca15: Would a blazor user take a dependency on this package and see this method? My thought is that if we go with WithTargeting, any application that would depend on this package should most likely use this method.

rossgrambo: My fault- yes `.WithTargeting`. 

Blazor Webassembly is just a client side thing- so it would work fine with an ASP.NET backend requiring this.

Blazor Server would not work as expected- and I wouldn't expect the dev to use the ASP.NET package. Although- Blazor Server are built on top of ASP.NET Core so it's not unreasonable a developer would try this.

zhenlan: We could add it by default if 1) it will not break any existing applications and 2) users can easily override it if they want to use their own targeting context accessor.

jimmyca15: Okay, having it with just `WithTargeting` sounds reasonable to me.

src/Microsoft.FeatureManagement.AspNetCore/AspNetCoreFeatureManagementBuilderExtensions.cs
```diff
@@ -44,5 +47,28 @@ public static IFeatureManagementBuilder UseDisabledFeaturesHandler(this IFeature
 
             return builder;
         }
+
+        /// <summary>
+        /// Adds the <see cref="DefaultHttpTargetingContextAccessor"/> to be used for targeting and registers the targeting filter to the feature management system.
```
jimmyca15, line 52: That's an internal type. It shouldn't show up in a public method summary. The compiler allows it?

jimmyca15: Since `DefaultHttpTargetingContextAccessor` is internal, users won't see it's method summary so some of those details would be useful here.

jimmyca15 (approved):

src/Microsoft.FeatureManagement.AspNetCore/AspNetCoreFeatureManagementBuilderExtensions.cs
rossgrambo: Good point- how about:   
```suggestion
        /// Adds the default ASP.NET <see cref="ITargetingContextAccessor"/> to be used for targeting and registers the targeting filter to the feature management system.
```
?

jimmyca15: I would say something like.

`Enables the use of targeting within the application and adds a targeting context accessor that extracts targeting details from a request's HTTP context.`

zhiyuanliang-ms:
@jimmyca15 @rossgrambo After this PR is merged, I think we can publish a new stable release. What do you think?

rossgrambo:
Sounds good to me. 

There was an issue with the build not starting. Should be fixed now, so I'll merge shortly.

PR #467, opened 2024-06-20
Activity Based TargetingId Persistance

rossgrambo:
## Why this PR?

1. Switches our Middleware and Initializer to use `Activity.Baggage` instead of `HttpContext.Items` to store Targeting information.
2. Removes the `Microsoft.FeatureManagement.ApplicationInsights.AspNetCore` package

## Visible Changes
The `Microsoft.FeatureManagement.ApplicationInsights.AspNetCore` will no longer be a nuget package deployed. The initializer has moved to `Microsoft.FeatureManagement.ApplicationInsights`.

### Migrating
Developers using the initializer will need to change their using statement from
```csharp
using Microsoft.FeatureManagement.ApplicationInsights.AspNetCore
```
to
```csharp
using Microsoft.FeatureManagement.ApplicationInsights
```

There are no changes to the line that adds the initializer or anywhere else. This line stays:
```csharp
builder.Services.AddSingleton<ITelemetryInitializer, TargetingTelemetryInitializer>();
```


jimmyca15:
> Removes the Microsoft.FeatureManagement.ApplicationInsights.AspNetCore package

I don't see any files removed in the diff

src/Microsoft.FeatureManagement.AspNetCore/TargetingHttpContextMiddleware.cs
```diff
@@ -50,14 +52,15 @@ public async Task InvokeAsync(HttpContext context, ITargetingContextAccessor tar
 
             if (targetingContext != null)
             {
-                context.Items[TargetingIdKey] = targetingContext.UserId;
+                var activityFeature = httpContext.Features.Get<IHttpActivityFeature>();
```
jimmyca15, line 55: Can it be null?

src/Microsoft.FeatureManagement.AspNetCore/TargetingHttpContextMiddleware.cs
```diff
@@ -50,14 +52,15 @@ public async Task InvokeAsync(HttpContext context, ITargetingContextAccessor tar
...
+                activityFeature.Activity.AddBaggage(TargetingIdKey, targetingContext.UserId);
```
jimmyca15, line 56: Can `Activity` be null?

src/Microsoft.FeatureManagement.Telemetry.ApplicationInsights/TargetingTelemetryInitializer.cs
```diff
@@ -0,0 +1,52 @@
...
+    /// </summary>
+    public class TargetingTelemetryInitializer : ITelemetryInitializer
+    {
+        private const string TargetingIdKey = $"Microsoft.FeatureManagement.TargetingId";
+
+        /// <summary>
+        /// Creates an instance of the TargetingTelemetryInitializer
+        /// </summary>
+        public TargetingTelemetryInitializer() { }
+
```
jimmyca15, line 23: Is there a reason to define the default constructor?

src/Microsoft.FeatureManagement.Telemetry.ApplicationInsights/TargetingTelemetryInitializer.cs
```diff
@@ -0,0 +1,52 @@
...
+            string targetingId = Activity.Current?.Baggage.FirstOrDefault(t => t.Key == TargetingIdKey).Value;
+
+            // Don't modify telemetry if there's no available targeting id
+            if (string.IsNullOrEmpty(targetingId))
+            {
+                return;
+            }
+
+            // Telemetry.Properties is deprecated in favor of ISupportProperties
+            if (telemetry is ISupportProperties telemetryWithSupportProperties)
```
jimmyca15, line 46: Is there a case that this wouldn't be true, perhaps if a user used an old version of app insights?

src/Microsoft.FeatureManagement.AspNetCore/TargetingHttpContextMiddleware.cs
rossgrambo: Seems like yes, it can be. Adjusting the following line from:
```csharp
activityFeature.Activity.AddBaggage(TargetingIdKey, targetingContext.UserId);
```
To:
```csharp
activityFeature?.Activity.AddBaggage(TargetingIdKey, targetingContext.UserId);
```

src/Microsoft.FeatureManagement.AspNetCore/TargetingHttpContextMiddleware.cs
rossgrambo: I see conflicting examples around this- so I'll add a null check to be safe. I suppose this will become:
```csharp
activityFeature?.Activity?.AddBaggage(TargetingIdKey, targetingContext.UserId);
```

src/Microsoft.FeatureManagement.Telemetry.ApplicationInsights/TargetingTelemetryInitializer.cs
rossgrambo: No, removed.

src/Microsoft.FeatureManagement.Telemetry.ApplicationInsights/TargetingTelemetryInitializer.cs
rossgrambo: My understanding is no, that they will always be available. The deprecation message for telemetry.Properties says:
> TelemetryContext.Properties is obsolete. Use GlobalProperties to set global level properties. For properties at item level, use ISupportProperties.Properties.

They were added in v2.21.0. .NET 6 and above should use at least v2.21.0, so I think we can assume this will be available. The versions before that are all deprecated (actually including v2.21.0 as well). v2.22.0 is the latest.

rossgrambo:
Updated to remove Microsoft.FeatureManagement.Telemetry.ApplicationInsights.AspNetCore

rossgrambo:
Any other concerns?

src/Microsoft.FeatureManagement.AspNetCore/TargetingHttpContextMiddleware.cs
```diff
@@ -50,14 +53,15 @@ public async Task InvokeAsync(HttpContext context, ITargetingContextAccessor tar
 
             if (targetingContext != null)
             {
-                context.Items[TargetingIdKey] = targetingContext.UserId;
+                var activityFeature = httpContext.Features.Get<IHttpActivityFeature>();
+                activityFeature?.Activity?.AddBaggage(TargetingIdKey, targetingContext.UserId);
```
jimmyca15, line 57: The null check on activity feature is a safeguard, but it's still unexpected right? I'm thinking we should log a warning if it's null.

jimmyca15 (approved):

src/Microsoft.FeatureManagement.AspNetCore/TargetingHttpContextMiddleware.cs
rossgrambo: Adjusted to:

```csharp
var activityFeature = httpContext.Features.Get<IHttpActivityFeature>();

if (activityFeature == null)
{
    _logger.LogDebug("The IHttpActivityFeature from the IFeatureCollection was null");
}
else if (activityFeature.Activity == null)
{
    _logger.LogDebug("The Activity on the IHttpActivityFeature was null");
} else
{
    activityFeature.Activity.AddBaggage(TargetingIdKey, targetingContext.UserId);
}
```

jimmyca15: That's a debug log. You don't consider it worth a warning?

rossgrambo: I was following suit on the other null check that causes the middleware to no op:
```csharp
    _logger.LogDebug("The targeting context accessor returned a null TargetingContext");
```

But I suppose these two are more unexpected than that one?

jimmyca15: > But I suppose these two are more unexpected than that one?

Yes, I would say that. And also I believe we didn't want to log warning in the middleware for not having a targeting context because the idea was to consider the middleware may be in use even if targeting is not enabled.

rossgrambo: PR: https://github.com/microsoft/FeatureManagement-Dotnet/pull/471

PR #470, opened 2024-06-27
Respect .NET and Microsoft feature management schemas in configuration

zhiyuanliang-ms:
## Why this PR?

This PR is created to work in conjunction with [Support both .NET and microsoft schema for feature flags](https://github.com/Azure/AppConfiguration-DotnetProvider/pull/566) for .NET provider.

Our .NET provider starts to embrace the [Microsoft feature management schema](https://github.com/Azure/AppConfiguration/blob/main/docs/FeatureManagement/FeatureManagement.v2.0.0.schema.json). To avoid the breaking change that old versions of feature management library (ealier than 3.2.0) cannot work with v8 .NET provider. The .NET provider will put feature flags into configuration in both [.NET feature management schema](https://github.com/microsoft/FeatureManagement-Dotnet/blob/main/schemas/FeatureManagement.Dotnet.v1.0.0.schema.json) and Microsoft schema. More specifically, all variant feature flags will be converted to the Microsoft schema, while regular feature flags will be converted to the .NET schema. This results in both schemas being present in the configuration simultaneously.

Feature management library needs to recognize and respect both of them.

## Visible changes

Previously, our `ConfigurationFeatureDefinitionProvider` will only respect one of Microsoft schema or .NET schema. If Microsoft schema can be found, any feature flag written in .NET schema will be ignored. This PR changes this behavior. All feature flags written in Microsoft schema and .NET schema will be recognized and used.

Now, when [`RootConfigurationFallbackEnabled`](https://github.com/microsoft/FeatureManagement-Dotnet/blob/preview/src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs#L53), it will only look for feature flags written in .NET schema. We don't want to break people who is using this function #308. But we also don't want to keep it for our new Microsoft schema because there will be a lot of corner cases when we support both .NET and Microsoft schema.


src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
```diff
@@ -84,16 +81,47 @@ public Task<FeatureDefinition> GetFeatureDefinitionAsync(string featureName)
...
+                    .FirstOrDefault(section =>
+                        string.Equals(section.Key, featureName, StringComparison.OrdinalIgnoreCase));
+
+                if (configuration != null)
+                {
+                    definition = _definitions.GetOrAdd(featureName, (_) => ParseDotnetFeatureDefinition(configuration));
+                }
+            }
 
             return Task.FromResult(definition);
```
jimmyca15, line 126: This line looks like it should become `return Task.FromResult(null)`;

jimmyca15: And we used to cache null in `_definitions`

src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
```diff
@@ -84,16 +81,47 @@ public Task<FeatureDefinition> GetFeatureDefinitionAsync(string featureName)
...
             }
 
-            //
-            // Query by feature name
-            FeatureDefinition definition = _definitions.GetOrAdd(featureName, (name) => ReadFeatureDefinition(name));
+            if (_definitions.TryGetValue(featureName, out FeatureDefinition definition))
+            {
+                return Task.FromResult(definition);
+            }
+
```
jimmyca15, line 93: I think to help readers follow what we have here it would be good to have something like

```
if (!TryGetMicrosoftSchemaFeatureDefinition(featureName, out definition))
{
    definition = GetDotnetSchemaFeatureDefinition(featureName);
}
```


Putting it all together
```
return Task.FromResult(
    _definitions.GetOrAdd(
        featureName,
        (name) =>
        {
            if (!TryGetMicrosoftSchemaFeatureDefinition(featureName, out definition))
            {
                definition = GetDotnetSchemaFeatureDefinition(featureName);
            }
        });
```

src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
```diff
@@ -109,97 +137,77 @@ public Task<FeatureDefinition> GetFeatureDefinitionAsync(string featureName)
...
 
-            //
-            // Iterate over all features registered in the system at initial invocation time
-            foreach (IConfigurationSection featureSection in GetFeatureDefinitionSections())
-            {
-                string featureName = GetFeatureName(featureSection);
+            IEnumerable<IConfigurationSection> microsoftFeatureManagementSection = GetFeatureDefinitionSections(MicrosoftFeatureManagementFields.FeatureManagementSectionName);
 
-                if (string.IsNullOrEmpty(featureName))
+            if (microsoftFeatureManagementSection != null)
```
jimmyca15, line 147: Usually IEnumerables don't come back as null, but rather empty. Is it really null?

src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
```diff
@@ -109,97 +137,77 @@ public Task<FeatureDefinition> GetFeatureDefinitionAsync(string featureName)
...
-                    .GetChildren()
-                    .FirstOrDefault(section =>
-                        string.Equals(
-                            section.Key,
-                            MicrosoftFeatureManagementFields.FeatureManagementSectionName,
-                            StringComparison.OrdinalIgnoreCase));
+            bool rootConfigurationFallback = microsoftFeatureManagementSection == null && RootConfigurationFallbackEnabled;
 
-                bool hasMicrosoftFeatureManagementSchema = MicrosoftFeatureManagementConfigurationSection != null;
+            IEnumerable<IConfigurationSection> dotnetFeatureManagementSection = GetDotnetFeatureManagementSection(rootConfigurationFallback);
```
jimmyca15, line 171: I think it would simplify things a bit if the root configuration fallback was calculated inside of `GetDotnetFeatureManagementSections`

src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
```diff
@@ -109,97 +137,77 @@ public Task<FeatureDefinition> GetFeatureDefinitionAsync(string featureName)
...
+            if (dotnetFeatureManagementSection == null && rootConfigurationFallback)
             {
-                return null;
-            }
-
-            return ReadFeatureDefinition(configuration);
-        }
+                if (!_configuration.GetChildren().Any())
+                {
+                    Logger?.LogDebug($"Configuration is empty.");
```
jimmyca15, line 204: Who is this message intended for?

zhiyuanliang-ms: It used to be logged when you passed an empty configuration to `AddFeatureManagement(configuration)`.

src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
zhiyuanliang-ms: How about

``` C#
return Task.FromResult(
    _definitions.GetOrAdd(
        featureName,
        (name) => GetMicrosoftSchemaFeatureDefinition(name) ?? GetDotnetSchemaFeatureDefinition(name)));
```

Both `GetMicrosoftSchemaFeatureDefinition` and `GetDotnetSchemaFeatureDefinition` method will return null if feature definition cannot be found.

jimmyca15: Sure

src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
```diff
@@ -84,18 +81,20 @@ public Task<FeatureDefinition> GetFeatureDefinitionAsync(string featureName)
...
-
             if (Interlocked.Exchange(ref _stale, 0) != 0)
             {
                 _definitions.Clear();
             }
 
-            //
-            // Query by feature name
-            FeatureDefinition definition = _definitions.GetOrAdd(featureName, (name) => ReadFeatureDefinition(name));
+            if (_definitions.TryGetValue(featureName, out FeatureDefinition definition))
```
jimmyca15, line 89: Doesn't GetOrAdd accomplish the same thing?

src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
```diff
@@ -129,80 +126,99 @@ public async IAsyncEnumerable<FeatureDefinition> GetAllFeatureDefinitionsAsync()
...
 
-            return ReadFeatureDefinition(configuration);
+            return ParseDotnetSchemaFeatureDefinition(configuration);
         }
 
-        private FeatureDefinition ReadFeatureDefinition(IConfigurationSection configurationSection)
+        private FeatureDefinition GetMicrosoftSchemaFeatureDefinition(string featureName)
         {
-            if (_microsoftFeatureManagementSchemaEnabled)
+            IEnumerable<IConfigurationSection> microsoftFeatureManagementSection = GetMicrosoftFeatureDefinitionSections();
```
jimmyca15, line 177:
```suggestion
            IEnumerable<IConfigurationSection> microsoftFeatureManagementSections = GetMicrosoftFeatureDefinitionSections();
```

jimmyca15 (approved):

PR #471, opened 2024-07-09
Switching log from debug to warning

rossgrambo:


src/Microsoft.FeatureManagement.AspNetCore/TargetingHttpContextMiddleware.cs
```diff
@@ -57,11 +57,11 @@ public async Task InvokeAsync(HttpContext httpContext, ITargetingContextAccessor
 
                 if (activityFeature == null)
                 {
-                    _logger.LogDebug("The IHttpActivityFeature from the IFeatureCollection was null");
+                    _logger.LogWarning("The IHttpActivityFeature from the IFeatureCollection was null");
```
zhenlan, line 60: If a customer came to us and asked us what they should do with this warning, what would we suggest? Can we make the message more helpful?

rossgrambo: It's a good question, and kinda no. The only circumstances I'm aware of where this can happen is if some code before ours is ripping out the IHttpActivityFeature from the ServiceCollection that's included by default. Or if we're in the context of an integration test or something where the middleware is called directly with a manually created HttpContext.

jimmyca15: Yes, the only way I can imagine it would be missing is if someone/something running earlier on in the request ripped it out.

Could probably say "A request feature required for telemetry, IHttpActivityFeature, was not found in the request's feature collection. Removing this feature from the request's feature collection will cause telemetry emission to fail."

rossgrambo: That's reasonable. I'll update to that format

rossgrambo: Updated to

```csharp
if (activityFeature == null)
{
    _logger.LogWarning("A request feature required for telemetry, IHttpActivityFeature, was not found in the request's feature collection. Removing this feature from the request's feature collection will cause telemetry emission to fail.");
}
else if (activityFeature.Activity == null)
{
    _logger.LogWarning("The IHttpActivityFeature request feature's Activity is null. This will cause telemetry emission to fail.");
}
```

zhenlan: > _logger.LogWarning("The IHttpActivityFeature request feature's Activity is null. This will cause telemetry emission to fail.");

Can we update this to provide more context like the other too?

rossgrambo: How about:

`The Activity on the IHttpActivityFeature request feature was null. The missing activity will cause telemetry emission to fail.` ?

I think the situation is if IHttpActivityFeature was manually added/overridden and Activity was not set- but having trouble turning that into an error message without assuming too much.

rossgrambo: Updated!

zhenlan (approved):

PR #472, opened 2024-07-12
Add linting rule

zhiyuanliang-ms:
## Why this PR?

#436 

## Visible changes

Linting rules are defined in .editorconfig file. 

<EnforceCodeStyleInBuild> property is added to the csproj file and then all code format issues will be reported during build process. In the editorconfig file, it sets the severity of IDE0055 and some other formating rules to error.
This will make build fail if the formating rule is not followed.


zhiyuanliang-ms:
#AzurePipeline run

zhiyuanliang-ms:
@jimmyca15 @rossgrambo Could you approve this PR so that I can merge main branch to preview branch?

rossgrambo (approved):

PR #484, opened 2024-07-26
Use ITargetingContext when calling GetVariantAsync

zhiyuanliang-ms:
## Why this PR?
#485 

## Visible Changes

1. Updated `IVariantFeatureManager.GetVariantAsync()` method to 
``` C#
ValueTask<Variant> GetVariantAsync(string feature, ITargetingContext context, CancellationToken cancellationToken = default);
```
It will accept `ITargetingContext` instead of `TargetingContext` type.

2. Added an entension method which is an overload of `GetVariantAsync`. This method will accept `TargetingContext` for discoverability.

### DIFF
--- a/src/Microsoft.FeatureManagement/FeatureManager.cs
+++ b/src/Microsoft.FeatureManagement/FeatureManager.cs
@@ -225,10 +225,10 @@ namespace Microsoft.FeatureManagement
         /// Gets the assigned variant for a specific feature.
         /// </summary>
         /// <param name="feature">The name of the feature to evaluate.</param>
-        /// <param name="context">An instance of <see cref="TargetingContext"/> used to evaluate which variant the user will be assigned.</param>
+        /// <param name="context">A context providing information that can used to evaluate which variant the user will be assigned.</param>
         /// <param name="cancellationToken">The cancellation token to cancel the operation.</param>
         /// <returns>A variant assigned to the user based on the feature's configured allocation.</returns>
-        public async ValueTask<Variant> GetVariantAsync(string feature, ITargetingContext context, CancellationToken cancellationToken = default)
+        public async ValueTask<Variant> GetVariantAsync<TContext>(string feature, TContext context, CancellationToken cancellationToken = default)
         {
             if (string.IsNullOrEmpty(feature))
             {
@@ -262,11 +262,11 @@ namespace Microsoft.FeatureManagement
 
             //
             // Determine Targeting Context
-            TargetingContext targetingContext;
+            ITargetingContext targetingContext;
 
             if (useContext)
             {
-                targetingContext = context as TargetingContext;
+                targetingContext = context as ITargetingContext;
             }
             else
             {
@@ -601,7 +601,7 @@ namespace Microsoft.FeatureManagement
             return context;
         }
 
-        private ValueTask<VariantDefinition> AssignVariantAsync(EvaluationEvent evaluationEvent, TargetingContext targetingContext, CancellationToken cancellationToken)
+        private ValueTask<VariantDefinition> AssignVariantAsync(EvaluationEvent evaluationEvent, ITargetingContext targetingContext, CancellationToken cancellationToken)
         {
             Debug.Assert(evaluationEvent != null);
 
--- a/src/Microsoft.FeatureManagement/FeatureManager.cs
+++ b/src/Microsoft.FeatureManagement/FeatureManager.cs
@@ -228,7 +228,7 @@ namespace Microsoft.FeatureManagement
         /// <param name="context">An instance of <see cref="TargetingContext"/> used to evaluate which variant the user will be assigned.</param>
         /// <param name="cancellationToken">The cancellation token to cancel the operation.</param>
         /// <returns>A variant assigned to the user based on the feature's configured allocation.</returns>
-        public async ValueTask<Variant> GetVariantAsync(string feature, TargetingContext context, CancellationToken cancellationToken = default)
+        public async ValueTask<Variant> GetVariantAsync(string feature, ITargetingContext context, CancellationToken cancellationToken = default)
         {
             if (string.IsNullOrEmpty(feature))
             {
--- a/src/Microsoft.FeatureManagement/FeatureManager.cs
+++ b/src/Microsoft.FeatureManagement/FeatureManager.cs
@@ -605,7 +605,7 @@ namespace Microsoft.FeatureManagement
             return context;
         }
 
-        private ValueTask<VariantDefinition> AssignVariantAsync(EvaluationEvent evaluationEvent, ITargetingContext targetingContext, CancellationToken cancellationToken)
+        private ValueTask<VariantDefinition> AssignVariantAsync(EvaluationEvent evaluationEvent, TargetingContext targetingContext, CancellationToken cancellationToken)
         {
             Debug.Assert(evaluationEvent != null);
 
--- a/src/Microsoft.FeatureManagement/FeatureManager.cs
+++ b/src/Microsoft.FeatureManagement/FeatureManager.cs
@@ -216,7 +216,7 @@ namespace Microsoft.FeatureManagement
                 throw new ArgumentNullException(nameof(feature));
             }
 
-            EvaluationEvent evaluationEvent = await EvaluateFeature<TargetingContext>(feature, context: null, useContext: false, cancellationToken);
+            EvaluationEvent evaluationEvent = await EvaluateFeature<object>(feature, context: null, useContext: false, cancellationToken);
 
             return evaluationEvent.Variant;
         }
@@ -262,13 +262,18 @@ namespace Microsoft.FeatureManagement
 
             //
             // Determine Targeting Context
-            ITargetingContext targetingContext;
+            TargetingContext targetingContext = null;
 
-            if (useContext)
+            if (useContext && context is ITargetingContext targetingInfo)
             {
-                targetingContext = context as ITargetingContext;
+                targetingContext = new TargetingContext
+                {
+                    UserId = targetingInfo.UserId,
+                    Groups = targetingInfo.Groups
+                };
             }
-            else
+
+            if (targetingContext == null)
             {
                 targetingContext = await ResolveTargetingContextAsync(cancellationToken).ConfigureAwait(false);
             }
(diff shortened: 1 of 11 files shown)


jimmyca15 (approved):

src/Microsoft.FeatureManagement/FeatureManager.cs
```diff
@@ -225,10 +225,10 @@ public async ValueTask<Variant> GetVariantAsync(string feature, CancellationToke
         /// Gets the assigned variant for a specific feature.
```
jimmyca15, line 225: In the `public async ValueTask<Variant> GetVariantAsync(string feature, CancellationToken cancellationToken = default)` method,

I can't see any reason that we don't call `await EvaluateFeature<object>(feature, context: null, useContext: false, cancellationToken);` instead of `await EvaluateFeature<TargetingContext>(feature, context: null, useContext: false, cancellationToken);` similarly to how we handle non-contextual IsEnabledAsync.

src/Microsoft.FeatureManagement/Telemetry/EvaluationEvent.cs
```diff
@@ -18,7 +18,7 @@ public class EvaluationEvent
         /// <summary>
         /// The targeting context used to evaluate the feature.
         /// </summary>
-        public TargetingContext TargetingContext { get; set; }
+        public ITargetingContext TargetingContext { get; set; }
```
jimmyca15, line 21: I'd like to keep this as `TargetingContext` to enable easy expansion. We should be able to create a new instance and copy over in FeatureManager.

jimmyca15 (changes requested):
TargetingContext in EvaluationEvent

src/Microsoft.FeatureManagement/FeatureManager.cs
```diff
@@ -225,10 +225,10 @@ public async ValueTask<Variant> GetVariantAsync(string feature, CancellationToke
...
         /// </summary>
         /// <param name="feature">The name of the feature to evaluate.</param>
-        /// <param name="context">An instance of <see cref="TargetingContext"/> used to evaluate which variant the user will be assigned.</param>
+        /// <param name="context">A context providing information that can used to evaluate which variant the user will be assigned.</param>
```
zhenlan, line 228: A context providing information that can be used to evaluate which variant will be assigned to the user.

src/Microsoft.FeatureManagement/FeatureManager.cs
```diff
@@ -225,10 +225,10 @@ public async ValueTask<Variant> GetVariantAsync(string feature, CancellationToke
...
         /// <param name="cancellationToken">The cancellation token to cancel the operation.</param>
         /// <returns>A variant assigned to the user based on the feature's configured allocation.</returns>
-        public async ValueTask<Variant> GetVariantAsync(string feature, TargetingContext context, CancellationToken cancellationToken = default)
+        public async ValueTask<Variant> GetVariantAsync<TContext>(string feature, TContext context, CancellationToken cancellationToken = default)
```
zhenlan, line 231: Isn't it required for `TContext` to implement `ITargetingContext` for this API to work? If so, what's the point to have `TContext`?

src/Microsoft.FeatureManagement/IVariantFeatureManager.cs
```diff
@@ -49,9 +49,9 @@ public interface IVariantFeatureManager
         /// Gets the assigned variant for a specific feature.
         /// </summary>
         /// <param name="feature">The name of the feature to evaluate.</param>
-        /// <param name="context">An instance of <see cref="TargetingContext"/> used to evaluate which variant the user will be assigned.</param>
+        /// <param name="context">A context providing information that can used to evaluate which variant the user will be assigned.</param>
```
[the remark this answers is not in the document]
zhiyuanliang-ms: I followed some previous [comments](https://github.com/microsoft/FeatureManagement-Dotnet/blob/preview/src/Microsoft.FeatureManagement/IVariantFeatureManager.cs#L35).

src/Microsoft.FeatureManagement/FeatureManager.cs
zhiyuanliang-ms: While `GetVariantAsync`, the first step is to evaluate the feature flag state according to the feature filter.

If the user uses a contextual feature filter which requires `TContext` for filter evaluation.
`TContext` is not necessarily `ITargetingContext`. The feature manager will still use it to evaluate whether feature flag is enabled or disabled. 
And he can also use two variants through `DefaultWhenEnabled` and `DefaultWhenDisabled` which don't require `TargetingContext` to do the variant allocation.

Previously, we only allow user to pass `TargetingContext` which will make it impossible to use contextual filter. 

How about this?

If the `TContext` doesn't implement `ITargetingContext`, the feature manager will still give it another chance to get targeting context from the targeting context accessor.

src/Microsoft.FeatureManagement/FeatureManager.cs
zhiyuanliang-ms: I followed the existing [comments](https://github.com/microsoft/FeatureManagement-Dotnet/blob/preview/src/Microsoft.FeatureManagement/IVariantFeatureManager.cs#L35).

src/Microsoft.FeatureManagement/FeatureManager.cs
zhenlan: You are talking about an edge case where `ITargetingContext` may not be necessary, but general speaking, `ITargetingContext` is required for variant assignment. We should enforce the context to implement `ITargetingContext`.

> If the `TContext` doesn't implement `ITargetingContext`, the feature manager will still give it another chance to get targeting context from the targeting context accessor.

No, we shouldn't use ambient context in contextual feature evaluation.

src/Microsoft.FeatureManagement/FeatureManager.cs
zhenlan: I’m not sure which comments you’re referring to, but the current one you have contains grammatical errors.

src/Microsoft.FeatureManagement/FeatureManager.cs
zhiyuanliang-ms: > ITargetingContext is required for variant assignment. We should enforce the context to implement ITargetingContext.

`ITargetingContext` is a public interface, we cannot modify it to avoid breaking changes.
If in the future, we want to add new fields in targeting context, we may create a new interface called IAttributeTargetingContext. If we enforce the context to implement ITargetingContext here, we will constrain ourselves.

`TargetingContext` should be sealed. We had an oversight. I created an issue #486 Currently, we are in a uncomfortable spot where we cannot modify anything related to targeting context to avoid potential breaking changes.

src/Microsoft.FeatureManagement/FeatureManager.cs
zhiyuanliang-ms: Line 35 in IVariantFeatureManager.cs. 

> but the current one you have contains grammatical errors.

Ok, I will change all of them.

src/Microsoft.FeatureManagement/IVariantFeatureManager.cs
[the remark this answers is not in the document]
zhenlan: Not resolved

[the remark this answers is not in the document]
zhiyuanliang-ms: Resolved in 0bdf279ed47c79903da3fdc88c16425d8249f7cb

src/Microsoft.FeatureManagement/FeatureManager.cs
```diff
@@ -144,7 +144,7 @@ public async Task<bool> IsEnabledAsync(string feature)
         /// Checks whether a given feature is enabled.
         /// </summary>
         /// <param name="feature">The name of the feature to check.</param>
-        /// <param name="appContext">A context providing information that can be used to evaluate whether a feature should be on or off.</param>
+        /// <param name="appContext">A context that provides information that can be used to evaluate whether a feature should be on or off.</param>
```
zhenlan, line 147: No need for this change. This causes nested "that" clauses.

src/Microsoft.FeatureManagement/FeatureManager.cs
```diff
@@ -170,7 +170,7 @@ public async ValueTask<bool> IsEnabledAsync(string feature, CancellationToken ca
...
```
zhenlan, line 173: No need for this change.

src/Microsoft.FeatureManagement/FeatureManager.cs
zhiyuanliang-ms: I will change to "A context that provides information to evaluate whether a feature should be on or off."

zhenlan: sure, it works.

src/Microsoft.FeatureManagement/FeatureManager.cs
rossgrambo: I also don't think Variants need to be fully tied to targeting. We don't support it yet- but it seems quite reasonable to want variants without wanting user based allocation. 

Making it TContext also aligns it with IsEnabled, which is more consistent. IsEnabled is also able to use Allocation in the same way. So whatever we decide on- the two interfaces should use the same.

jimmyca15 (approved):

src/Microsoft.FeatureManagement/VariantFeatureManagerExtensions.cs
```diff
@@ -0,0 +1,28 @@
...
+// Licensed under the MIT license.
+//
+using Microsoft.FeatureManagement.FeatureFilters;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.FeatureManagement
+{
+    /// <summary>
+    /// Extensions used to provide an overload for <see cref="IVariantFeatureManager.GetVariantAsync(string, ITargetingContext, CancellationToken)"/> that accepts a <see cref="TargetingContext"/> type directly.
```
jimmyca15, line 11: This class summary is too coupled to the method summary. It's unnecessary and would need to be updated if the method was updated or ever an other method was added.

```suggestion
    /// Extensions for <see cref="IVariantFeatureManager"/>.
```

PR #487, opened 2024-08-20
Adjusts builder extension methods to be friendlier

rossgrambo:
Preview now with app insights telemetry looks like this:

```csharp
// Add App Insights
builder.Services.AddApplicationInsightsTelemetry(); // Shouldn't be done by us

builder.Services.AddHttpContextAccessor();

// Add FeatureManagement's TargetingTelemetryInitializer
builder.Services.AddSingleton<ITelemetryInitializer, TargetingTelemetryInitializer>();

// Add FeatureManagement
    .WithTargeting() // Uses DefaultHttpTargetingAccessor, requires HttpContextAccessor
    .AddApplicationInsightsTelemetryPublisher();

var app = builder.Build();

//...

app.UseMiddleware<TargetingHttpContextMiddleware>();
```

Making a few opinionated changes- we can make it:

```csharp
// Add App Insights
builder.Services.AddApplicationInsightsTelemetry(); // Shouldn't be done by us

    .WithTargeting() // Uses DefaultHttpTargetingAccessor, adds HttpContextAccessor if it's not added already
    .AddApplicationInsightsTelemetry(); // Adds TargetingTelemetryInitializer as well

var app = builder.Build();

app.UseMiddleware<TargetingHttpContextMiddleware>();
```

### DIFF
--- a/src/Microsoft.FeatureManagement.Telemetry.ApplicationInsights/FeatureManagementBuilderExtensions.cs
+++ b/src/Microsoft.FeatureManagement.Telemetry.ApplicationInsights/FeatureManagementBuilderExtensions.cs
@@ -15,11 +15,11 @@ namespace Microsoft.FeatureManagement
     public static class FeatureManagementBuilderExtensions
     {
         /// <summary>
-        /// Adds the <see cref="ApplicationInsightsEventPublisher"/> using <see cref="ApplicationInsightsHostedService"/> to the feature management builder.
+        /// Adds the <see cref="TargetingTelemetryInitializer"/> and the <see cref="ApplicationInsightsEventPublisher"/> using <see cref="ApplicationInsightsHostedService"/> to the feature management builder.
         /// </summary>
         /// <param name="builder">The feature management builder.</param>
         /// <returns>The feature management builder.</returns>
-        public static IFeatureManagementBuilder AddApplicationInsightsTelemetryPublisher(this IFeatureManagementBuilder builder)
+        public static IFeatureManagementBuilder AddApplicationInsightsTelemetry(this IFeatureManagementBuilder builder)
         {
             if (builder == null)
             {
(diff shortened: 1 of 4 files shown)


rossgrambo:
Telemetry Publisher vs Initializer relationship:
* `AddApplicationInsightsTelemetryPublisher` publishes Impression events and doesn't directly require the Initializer. 
* The Initializer makes all outgoing Telemetry able to be connected to Impression events (by appending TargetingId).

I think it's okay to add these together- because it's very unlikely someone will want the publisher and not want the initializer.

src/Microsoft.FeatureManagement.Telemetry.ApplicationInsights/FeatureManagementBuilderExtensions.cs
```diff
@@ -29,6 +31,8 @@ public static IFeatureManagementBuilder AddApplicationInsightsTelemetryPublisher
                 throw new ArgumentException($"The provided builder's services must not be null.", nameof(builder));
             }
 
+            builder.Services.AddSingleton<ITelemetryInitializer, TargetingTelemetryInitializer>();
```
zhiyuanliang-ms, line 34: If we add telemetry initialier here. The method name should change to `AddApplicationInsightsTelemetry` or something similar. Adding the telemetry initializer will make all telemetry be tagged with targeting id. Is there any valid reason that the user won't want to have targeting id in some of telemetry?

jimmyca15: I like @zhiyuanliang-ms suggestion here. `AddApplicationInsightsTelemetry`

jimmyca15:
I see three improvements here

(1) Add httpcontext accessor in WithTargeting
(2) Add telemetry initializer when adding telemetry publisher
(3) Add syntax sugar helper for middleware

(1) Agree to have it. The default targeting context accessor doesn't make sense without http context accessor being present.
(2) Agree to have it. The scenario for feature management telemetry is to emit targeting id if present.
(3) Doesn't seem helpful enough to me at this point to have it.

rossgrambo:
Adjusted name of adding App Insights and removed middleware helper. Adjusted PR description to reflect the latest.

.AddApplicationInsightsTelemetry~Publisher~();
~app.UseFeatureManagement();~

jimmyca15 (approved):

PR #490, opened 2024-08-26
Renames and updates EvaluationDataToAppInsights example

rossgrambo:
## Why this PR?

We want to clean up examples

## Visible changes

* Renames `EvaluationDataToAppInsights` to `VariantAndTelemetryDemo`
* Removes custom TargetingContextAccessor and instead uses the default accessor
  * Updates auth to populate HttpContext.User instead of simple cookie reading


zhiyuanliang-ms:
We also have an example called `VariantServiceDemo` which uses custom targeting context accessor and cookie reading. Shall we also update it in this PR?We also have an example called `VariantServiceDemo` which uses custom targeting context accessor and cookie reading. Shall we also update it in this PR?

rossgrambo:
I considered it- but figured we also want an example showing how to declare and use a custom targeting accessor, so I didn't make changes.

examples/VariantAndTelemetryDemo/Pages/Shared/_Layout.cshtml
```diff
@@ -6,10 +6,10 @@
 <head>
     <meta charset="utf-8" />
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
-    <title>@ViewData["Title"] - EvaluationDataToApplicationInsights</title>
+    <title>@ViewData["Title"] - VariantAndTelemetryDemo</title>
     <link rel="stylesheet" href="~/lib/bootstrap/dist/css/bootstrap.min.css" />
     <link rel="stylesheet" href="~/css/site.css" asp-append-version="true" />
-    <link rel="stylesheet" href="~/EvaluationDataToApplicationInsights.styles.css" asp-append-version="true" />
+    <link rel="stylesheet" href="~/VariantAndTelemetryDemo.styles.css" asp-append-version="true" />
```
zhiyuanliang-ms, line 12: Is this a file "~/VariantAndTelemetryDemo.styles.css"?

zhiyuanliang-ms (approved):

README.md
```diff
@@ -20,7 +20,7 @@ Feature management provides a way to develop and expose application functionalit
 * [ASP.NET Core Web App (Razor Page)](./examples/RazorPages)
 * [ASP.NET Core Web App (MVC)](./examples/FeatureFlagDemo)
 * [Blazor Server App](./examples/BlazorServerApp)
-* [ASP.NET Core Web App with Feature Flag Telemetry](./examples/EvaluationDataToApplicationInsights)
+* [ASP.NET Core Web App with Variants and Telemetry](./examples/VariantAndTelemetryDemo)
```
zhiyuanliang-ms, line 23: The example is called `VariantAndTelemetryDemo`. Can we use singular here for consistency if there is no grammer issue?

examples/VariantAndTelemetryDemo/Program.cs
```diff
@@ -1,35 +1,34 @@
...
 //
-using EvaluationDataToApplicationInsights;
-using Microsoft.ApplicationInsights.Extensibility;
 using Microsoft.FeatureManagement;
-using Microsoft.FeatureManagement.Telemetry.ApplicationInsights;
 
 var builder = WebApplication.CreateBuilder(args);
 
+//
+// Uses cookie auth for simplicity and randomizing user
```
zhiyuanliang-ms, line 9:
```suggestion
// Use cookie auth for simplicity and randomizing user
```

README.md
rossgrambo: Weirdly- I think the grammar makes sense. This is an app that has variants (plural). But when describing something as a demo- it doesn't need to be plural. Like FilterDemo (instead of FiltersDemo). 

If others feel strongly to align the two- I'd prefer changing the package to `VariantsAndTelemetryDemo`

PR #491, opened 2024-08-30
For examples and tests- fixes warnings, updates packages, and updates…

rossgrambo:
## Why this PR?

Planning to release soon- it's a good opportunity to update packages, examples, and fix some linting errors.

## Visible changes

Some minor version bumps in test packages and examples.


examples/TargetingConsoleApp/Program.cs
```diff
@@ -17,7 +17,7 @@
 services.AddSingleton(configuration)
         .AddFeatureManagement();
 
-IUserRepository userRepository = new InMemoryUserRepository();
+InMemoryUserRepository userRepository = new InMemoryUserRepository();
```
zhiyuanliang-ms, line 20:
```suggestion
var userRepository = new InMemoryUserRepository();
```

examples/VariantAndTelemetryDemo/Pages/Index.cshtml.cs
```diff
@@ -5,21 +5,15 @@
...
 namespace VariantAndTelemetryDemo.Pages
 {
-    public class IndexModel : PageModel
+    public class IndexModel(
+        IVariantFeatureManager featureManager,
+        TelemetryClient telemetry) : PageModel
     {
-        private readonly IVariantFeatureManager _featureManager;
-        private readonly TelemetryClient _telemetry;
+        private readonly IVariantFeatureManager _featureManager = featureManager ?? throw new ArgumentNullException(nameof(featureManager));
```
zhiyuanliang-ms, line 12: Is this change triggered for any linting rule? I prefer the previous pattern.

src/Microsoft.FeatureManagement/FeatureManager.cs
```diff
@@ -699,9 +699,7 @@ private ValueTask<VariantDefinition> AssignVariantAsync(EvaluationEvent evaluati
 
         private void BindSettings(IFeatureFilterMetadata filter, FeatureFilterEvaluationContext context, int filterIndex)
         {
-            IFilterParametersBinder binder = filter as IFilterParametersBinder;
-
-            if (binder == null)
+            if (!(filter is IFilterParametersBinder binder))
```
zhiyuanliang-ms, line 702: Is this change caused by any linting rule?

zhiyuanliang-ms (approved):

examples/TargetingConsoleApp/Program.cs
rossgrambo: Did we commit to this format with our linter? I believe others prefer this way, but personally, I don't like it. Haha.

jimmyca15: Both of these options are fine.

The only thing that's not fine is using var when the type isn't obvious in the assignment.

jimmyca15: For the sake of taking a direction and rolling with it, our code base heavily prefers the option of

`var userRepository = new InMemoryUserRepository();`

As the type repetition is unnecessary.

In the future I expect we'll shift to the latest language syntax for this scenario

`InMemoryUserRepository userRepository = new();`

src/Microsoft.FeatureManagement/FeatureManager.cs
rossgrambo: Yeah this was shown to me by code analysis. I can't find the rule- even if I undo the change, it's no longer flagging it... 

The check right below this uses this format anyway, so we should be consistent anyway.

examples/TargetingConsoleApp/Program.cs
rossgrambo: Sounds good, I'll do with `var` for now. I like the latest when we're able to make the switch.

examples/VariantAndTelemetryDemo/Pages/Index.cshtml.cs
rossgrambo: Yeah this was triggered by primary constructor: https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/ide0290

Which I do think looks nicer- although it's a little less nice when we have ArgumentNullException checks

zhiyuanliang-ms: > although it's a little less nice when we have ArgumentNullException checks

Yes. And I do think the previous one looks nicer to me. hahaha 

I don't remember our current linting rule set which only followed default settings of IDE0055 will enforce this. If not, I prefer to not this rule in the example projects.

src/Microsoft.FeatureManagement/FeatureManager.cs
zhiyuanliang-ms: > so we should be consistent anyway

Do you mean to revert this change or to keep it?

rossgrambo: Either keep this change- or change the following if statement to use the same style as this one.

examples/VariantAndTelemetryDemo/Pages/Index.cshtml.cs
rossgrambo: I don't feel strongly either way, so I'll go ahead and revert it.

jimmyca15 (approved):

PR #495, opened 2024-09-11
Adds more detailed telemetry fields

rossgrambo:
## Why this PR?

We want reader of telemetry to have more insight into the state of the feature flag. 

* Adds additional telemetry fields.
* Added the same BytesExtension we have on the provider for Base64Url.
* Adds and adjusts unit tests

## Visible Changes

* FeatureEvaluation events will have new fields:
  * **VariantAssignmentPercentage** - The percent of users we expect to be assigned the given variant for the given reason
  * **DefaultWhenEnabled** - The DefaultWhenEnabled field
  * **AllocationId** - An id that uniquely reflects the state of allocation. Two events with the same AllocationId can expect a given TargetingContext will always give the same result.  **Note- This is instead computed in App Configuration Provider


rossgrambo:
@zhiyuanliang-ms if you have time could you review this?

src/Microsoft.FeatureManagement/Telemetry/TelemetryEventHandler.cs
```diff
@@ -0,0 +1,92 @@
...
+                foreach (KeyValuePair<string, string> kvp in evaluationEvent.FeatureDefinition.Telemetry.Metadata)
+                {
+                    if (tags.ContainsKey(kvp.Key))
+                    {
+                        logger?.LogWarning($"{kvp.Key} from telemetry metadata will be ignored, as it would override an existing key.");
+
+                        continue;
+                    }
+
+                    tags[kvp.Key] = kvp.Value;
```
samsadsam, line 52: Can we instead emit all this metadata as an object with key `Metadata` ?

src/Microsoft.FeatureManagement/Telemetry/TelemetryEventHandler.cs
```diff
@@ -0,0 +1,92 @@
...
+            if (evaluationEvent.VariantAssignmentReason == VariantAssignmentReason.DefaultWhenEnabled)
+            {
+                // If the variant was assigned due to DefaultWhenEnabled, the percentage reflects the unallocated percentiles
+                double allocatedPercentage = 0;
+
+                if (evaluationEvent.FeatureDefinition.Allocation?.Percentile != null)
+                {
+                    allocatedPercentage += evaluationEvent.FeatureDefinition.Allocation.Percentile
+                        .Sum(p => p.To - p.From);
+                }
```
samsadsam, line 66:
```suggestion
double allocatedPercentage = evaluationEvent.FeatureDefinition.Allocation?.Percentile?.Sum(p => p.To - p.From) ?? 0;    
```

src/Microsoft.FeatureManagement/FeatureManager.cs
```diff
@@ -373,56 +373,13 @@ private async ValueTask<EvaluationEvent> EvaluateFeature<TContext>(string featur
                     Activity.Current != null &&
                     Activity.Current.IsAllDataRequested)
                 {
-                    AddEvaluationActivityEvent(evaluationEvent);
+                    TelemetryEventHandler.HandleEvaluationEvent(evaluationEvent, Logger);
```
samsadsam, line 376:
```suggestion
     FeatureEvaluationTelemetry.Publish(evaluationEvent, Logger);
```
How does this one sound?

tests/Tests.FeatureManagement/appsettings.json
```diff
@@ -209,7 +209,7 @@
               "to": 50
             }
           ],
-          "seed": 1234
+          "seed": "1234"
```
samsadsam, line 212: Why did the seed change from a number to string?

src/Microsoft.FeatureManagement/FeatureManager.cs
rossgrambo: I like it. Updated

src/Microsoft.FeatureManagement/Telemetry/TelemetryEventHandler.cs
rossgrambo: We could, but I don't think it adds any value. This is the outgoing event that will go to App Insights. As far as their concerned- this should just flattened data. These fields are only considered `metadata` because they were given to FM from an outside source (like the provider).

tests/Tests.FeatureManagement/appsettings.json
rossgrambo: Schema dictates it's a string. In .net this happened to work because everything is turned into a string when placed in config.

samsadsam (approved):

PR #501, opened 2024-09-16
Last one wins flag providers

rossgrambo:
## Why this PR?

Updates the flag provider to handle duplicate flags by using the last flag instead of the first. 

## Visible Changes

* For the [Microsoft Feature Flag Schema](https://github.com/microsoft/FeatureManagement/blob/main/Schema/FeatureManagement.v2.0.0.schema.json) - When there are duplicate feature flag ids- the last flag will be used. This allows multiple flag providers or multiple sources to load flags in a consistent way across languages. 
  * This also aligns with the existing functionality in stable. 
  * This change only affects the built-in `ConfigurationFeatureDefinitionProvider`, custom providers are still in charge of their ordering.

### DIFF
--- a/src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
+++ b/src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
@@ -156,7 +156,7 @@ namespace Microsoft.FeatureManagement
             IEnumerable<IConfigurationSection> dotnetFeatureDefinitionSections = GetDotnetFeatureDefinitionSections();
 
             IConfigurationSection configuration = dotnetFeatureDefinitionSections
-                .LastOrDefault(section =>
+                .FirstOrDefault(section =>
                     string.Equals(section.Key, featureName, StringComparison.OrdinalIgnoreCase));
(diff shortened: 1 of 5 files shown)


src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
```diff
@@ -156,7 +156,7 @@ private FeatureDefinition GetDotnetSchemaFeatureDefinition(string featureName)
             IEnumerable<IConfigurationSection> dotnetFeatureDefinitionSections = GetDotnetFeatureDefinitionSections();
 
             IConfigurationSection configuration = dotnetFeatureDefinitionSections
-                .FirstOrDefault(section =>
+                .LastOrDefault(section =>
```
jimmyca15, line 159: No change should be necessary for .NET featre definition schema.

rossgrambo: Reverted back to FirstOrDefault

jimmyca15 (approved):

Issue #117, opened 2021-03-25
CancellationToken Support


matthawley:
Curious if there's a reason why all the Async methods on IFeatureManager do not take a cancellation token? Can overloads be added that take this in and appropriately use them within the stack? If things aren't truly asynchronous, can synchronous methods be exposed?

jimmyca15:
@matthawley 

It's an oversight. Since IFeatureManager is an interface we need a breaking change release to address it. But it is known and we plan to address it when we have the opportunity to make a breaking change. Should be upcoming based off what we have in the works.

zhiyuanliang-ms:
Hi, @matthawley 
This issue will be solved in our next release.

brental:
@zhenlan @zhiyuanliang-ms In #376 we were discussing cancellation tokens and the fact that the FeatureManager class does not support them came up. I was going to raise a new issue for that to address it in v4 (as part of adding cancellation token support) but found this pre-existing issue. It would be good to ensure that support gets added as part of v4.

zhiyuanliang-ms:
@jimmyca15 @rossgrambo ISessionManager should take cancellation token too.

brental:
@zhenlan @zhiyuanliang-ms Are these changes still planned for v4.0.0? I ask because as per #500 the v4 code has been merged into main in prep for release but the changes to address this issue have not been implemented. It would be good to have these changes in as part of v4 given they would be breaking changes.

jimmyca15:
@brental v4 adds `IVariantFeatureManager` ([link](https://github.com/microsoft/FeatureManagement-Dotnet/blob/7050fa0c5138ad44860236f7496958b511ea8f6d/src/Microsoft.FeatureManagement/IVariantFeatureManager.cs#L38)). This interface is a superset of `IFeatureManager`, and it also supports cancellation token. 

Since this can be used for proper cancellation, we avoided any breaking change on `IFeatureManager`.

jimmyca15:
@brental

This issue should have been updated. The intention of the v4 release is to go without breaking changes. 

> Consumers should not be forced to use IVariantFeatureManager simply to have cancellation token support. So, I think the opportunity should be taken to make the breaking change so they also have proper cancellation token support.

I don't think it's so clear cut to say we should take a breaking change to modify `IFeatureManager` when `IVariantFeatureManager` will offer this functionality.

[...]

PR #505, opened 2024-09-25
Resolves snapshot breaking change and adjusts feature tag helper as well

rossgrambo:
## Why this PR?

* Resolves breaking change with `IFeatureManagerSnapshot` brought up by https://github.com/microsoft/FeatureManagement-Dotnet/issues/117#issuecomment-2369011536
* More unit tests to catch similar issues in the future. 

## Visible Changes

* `IFeatureManagerSnapshot` and `FeatureTagHelper` now require both an `IFeatureManager` and an `IVariantFeatureManager`. For the common case, these will refer to the same instance. By separating them, we preserve backwards compatability for custom `IFeatureManager` implementations.

### DIFF
--- a/src/Microsoft.FeatureManagement.AspNetCore/TagHelpers/FeatureTagHelper.cs
+++ b/src/Microsoft.FeatureManagement.AspNetCore/TagHelpers/FeatureTagHelper.cs
@@ -14,7 +14,8 @@ namespace Microsoft.FeatureManagement.Mvc.TagHelpers
     /// </summary>
     public class FeatureTagHelper : TagHelper
     {
-        private readonly IVariantFeatureManager _featureManager;
+        private readonly IFeatureManager _featureManager;
+        private readonly IVariantFeatureManager _variantFeatureManager;
 
         /// <summary>
         /// A feature name, or comma separated list of feature names, for which the content should be rendered. By default, all specified features must be enabled to render the content, but this requirement can be controlled by the <see cref="Requirement"/> property.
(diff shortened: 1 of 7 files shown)


src/Microsoft.FeatureManagement/FeatureManagerSnapshot.cs
```diff
@@ -16,19 +16,38 @@ namespace Microsoft.FeatureManagement
     /// </summary>
     class FeatureManagerSnapshot : IFeatureManagerSnapshot, IVariantFeatureManagerSnapshot
     {
-        private readonly IVariantFeatureManager _featureManager;
-        private readonly ConcurrentDictionary<string, ValueTask<bool>> _flagCache = new ConcurrentDictionary<string, ValueTask<bool>>();
+        private readonly IFeatureManager _featureManager;
+        private readonly IVariantFeatureManager _variantFeatureManager;
+        private readonly ConcurrentDictionary<string, Task<bool>> _flagCache = new ConcurrentDictionary<string, Task<bool>>();
```
jimmyca15, line 21: I don't see any reason to have two flag caches.

src/Microsoft.FeatureManagement.AspNetCore/TagHelpers/FeatureTagHelper.cs
```diff
@@ -38,12 +39,14 @@ public class FeatureTagHelper : TagHelper
         public string Variant { get; set; }
 
         /// <summary>
-        /// Creates a feature tag helper.
+        /// Creates a feature tag helper. Takes both a feature manager and a variant feature manager for backwards compatibility.
```
jimmyca15, line 42: I don't think that needs to be in the class summary. Might be helpful as a code comment.

src/Microsoft.FeatureManagement/FeatureManagerSnapshot.cs
rossgrambo: Reason I did two caches is when the `IFeatureManager` and `IVariantFeatureManager` are different implementations (someone today has a custom IFeatureManager). If the `IFeatureManager` returns true for a feature while the `IVariantFeatureManager` would return false- it feels incorrect to have one take priority over the other.

We could have one cache and use a prefix or something but this felt cleaner to me.

zhiyuanliang-ms (approved):

jimmyca15 (approved):

src/Microsoft.FeatureManagement/FeatureManagerSnapshot.cs
rossgrambo: We aligned on a single cache to better fit the common case.

PR #508, opened 2024-10-17
Allocation cleanup

rossgrambo:
## Why this PR?

We noticed we were allocating a lot of memory in v4 compared to v3. This PR resolves many instances of extra allocations or memory and resolves some issues that existed in v3 as well. 

Now- the allocations are just the EvaluationEvent object (which could be improved more for the boolean and missing defintion case), async closures, and enumerators only when needed. 

Additionally-
* Resolves some spelling issues
* Adds a pull request template
* Bumps vulnerable packages in examples and test projects

## Visible Changes

No changes are visible to developers using the library. They may notice allocation wins.


src/Microsoft.FeatureManagement/FeatureManager.cs
```diff
@@ -130,9 +144,7 @@ public TargetingEvaluationOptions AssignerOptions
         /// <returns>True if the feature is enabled, otherwise false.</returns>
         public async Task<bool> IsEnabledAsync(string feature)
         {
-            EvaluationEvent evaluationEvent = await EvaluateFeature<object>(feature, context: null, useContext: false, CancellationToken.None);
-
-            return evaluationEvent.Enabled;
+            return (await EvaluateFeature<object>(feature, context: null, useContext: false, CancellationToken.None).ConfigureAwait(false)).Enabled;
```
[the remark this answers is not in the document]
rossgrambo: Yes it was not there. Although it felt safe to me to add it- because the function it calls (EvaluateFeature) uses `.ConfigureAwait(false)` when reading configs and when evaluating true or false. So I don't think anyone could have utilized the default (aka `.ConfigureAwait(true)`) behavior.

rossgrambo:
@avanigupta @zhiyuanliang-ms @zhenlan This is blocking v4 release If you have time to review!

src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
```diff
@@ -86,10 +86,13 @@ public Task<FeatureDefinition> GetFeatureDefinitionAsync(string featureName)
                 _definitions.Clear();
             }
 
-            return Task.FromResult(
-                _definitions.GetOrAdd(
-                    featureName,
-                    (_) => GetMicrosoftSchemaFeatureDefinition(featureName) ?? GetDotnetSchemaFeatureDefinition(featureName)));
+            if (!_definitions.ContainsKey(featureName))
```
zhiyuanliang-ms, line 89: What's the reason for this change?

zhiyuanliang-ms, line 89: What's the reason for this change?

tests/Tests.FeatureManagement/FeatureManagementTest.cs
```diff
@@ -1976,4 +1976,51 @@ public async Task CustomIFeatureDefinitionProvider()
             Assert.True(called);
         }
     }
+
+    public class PerformanceTests
```
zhiyuanliang-ms, line 1980: Can we use the tool like [benchmark .net](https://github.com/dotnet/BenchmarkDotNet) to verifiy this change can improve the performance?

zhiyuanliang-ms:
@rossgrambo Could please explain why storing task in the dictionary can save memory?

rossgrambo:
The single line that caused a developer to write in was:

```
return Task.FromResult(
                _definitions.GetOrAdd(
                    featureName,
                    (_) => GetMicrosoftSchemaFeatureDefinition(featureName) ?? GetDotnetSchemaFeatureDefinition(featureName)));
```

The problem with this function is that the lambda/delegate/func syntax of:

```
(_) => ...
```

Creates a Func. Although it's not called in the case we hit the cache- it still allocates memory for the func every time this line is hit. This is because it happens before the `GetOrAdd` method is called actually called, as the parameters are resolved first before the method can be called.

This line also uses `Task.FromResult`. This creates a unique `Task` object on each request- even when the item is found in cache. By adjusting the cache to include the `Task`, we're able to keep a 1:1 relationship between `Task` and `FeatureDefinition`- rather than a 1:1 between `Task` and number of called to `IsEnabledAsync()`.

src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
```diff
@@ -24,9 +24,10 @@ public sealed class ConfigurationFeatureDefinitionProvider : IFeatureDefinitionP
         // provider to be marked for caching as well.
 
         private readonly IConfiguration _configuration;
-        private readonly ConcurrentDictionary<string, FeatureDefinition> _definitions;
+        private readonly ConcurrentDictionary<string, Task<FeatureDefinition>> _definitions;
         private IDisposable _changeSubscription;
         private int _stale = 0;
+        public Func<string, Task<FeatureDefinition>> schemaReadFunc;
```
zhiyuanliang-ms, line 30:
```suggestion
        public Func<string, Task<FeatureDefinition>> getFeatureDefinitionFunc;
```

src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
```diff
@@ -151,41 +155,41 @@ public async IAsyncEnumerable<FeatureDefinition> GetAllFeatureDefinitionsAsync()
             }
         }
 
-        private FeatureDefinition GetDotnetSchemaFeatureDefinition(string featureName)
+        private static FeatureDefinition GetDotnetSchemaFeatureDefinition(string featureName, IConfiguration configuration, bool rootConfigurationFallbackEnabled)
```
zhiyuanliang-ms, line 158: The current implementation (i.e. setting schemaReadFunc in the constructor) don't require these methods to be static right? Doing this is for performance, is that true?

zhiyuanliang-ms (approved):

zhiyuanliang-ms:
Discussed offline with @rossgrambo, this PR looks good to me.

src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
avanigupta, line 30: should be private, and start with _

src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
rossgrambo: True. I don't think there's even a performance gain with static here since we already have an instance of the class. Going to remove static from these.

rossgrambo:
Most benchmarking was done in VS- but it doesn't have a great way to compare. BenchmarkDotNet does! Here's the result (4.0.0/4.9999.9999 includes these changes):

image

tests/Tests.FeatureManagement/FeatureManagementTest.cs
avanigupta: Can we create an issue to track this? It doesn't need to block v4 release.

rossgrambo:
Ran test again using .net8 (previous screenshot was .net6). This version is actually the most performant when using .net 8.
image

avanigupta (approved):

PR #513, opened 2024-10-30
Drops .net7 as a target

rossgrambo:
## Why this PR?

* Removes .net7 as it's out of support. Keeping .net6 for now to help adoption- although it will be out of support soon.
* Adds .net9 as a target for tests and asp.net
* Adds a missing readonly

## Visible Changes

* No visible changes- although we will no longer run tests directly on .net 7. Once .net 6 is removed- customers using .net6 or .net7 will be unable to use the library.


src/Microsoft.FeatureManagement.AspNetCore/Microsoft.FeatureManagement.AspNetCore.csproj
```diff
@@ -12,7 +12,7 @@
   <Import Project="..\..\build\Versioning.props" />
 
   <PropertyGroup>
-    <TargetFrameworks>net6.0;net7.0;net8.0</TargetFrameworks>
+    <TargetFrameworks>net6.0;net8.0;net9.0</TargetFrameworks>
```
avanigupta, line 15: We might need to add .net 9 SDK for the CICD pipeline

avanigupta: https://github.com/microsoft/FeatureManagement-Dotnet/blob/main/build/install-dotnet.ps1

build/install-dotnet.ps1
```diff
@@ -7,4 +7,6 @@
 
 &([scriptblock]::Create((Invoke-WebRequest -UseBasicParsing 'https://dot.net/v1/dotnet-install.ps1'))) -Channel 7.0
```
WeihanLi, line 8: Should `7.0` be removed since we drop `net7` target

rossgrambo: Tracked in https://github.com/microsoft/FeatureManagement-Dotnet/issues/517

PR #522, opened 2024-12-09
Small fix for recurring time window testcase

zhiyuanliang-ms:
## Why this PR?

The validation of recurrence configuration has been decoupled from recurrence evaluation. I forgot to update the related testcase. Also, update the implementation to make it more readable, not a bug fix.

related PR: #266 

## Visible Changes

-changes that are visible to developers using this library-

### DIFF
--- a/src/Microsoft.FeatureManagement/FeatureFilters/Recurrence/RecurrenceValidator.cs
+++ b/src/Microsoft.FeatureManagement/FeatureFilters/Recurrence/RecurrenceValidator.cs
@@ -381,7 +381,7 @@ namespace Microsoft.FeatureManagement.FeatureFilters
                 DateTime firstDayOfNextWeek = firstDayOfThisWeek.AddDays(DaysPerWeek);
 
                 DateTime firstOccurrenceInNextWeek = firstDayOfNextWeek.AddDays(
-                    CalculateWeeklyDayOffset(sortedDaysOfWeek.First(), firstDayOfWeek));
+                    CalculateWeeklyDayOffset(sortedDaysOfWeek.First(), firstDayOfWeek)); // firstDayOfWeek may not in the sortedDaysOfWeek
 
                 TimeSpan gap = firstOccurrenceInNextWeek - prev;
 
--- a/src/Microsoft.FeatureManagement/FeatureFilters/Recurrence/RecurrenceValidator.cs
+++ b/src/Microsoft.FeatureManagement/FeatureFilters/Recurrence/RecurrenceValidator.cs
@@ -341,6 +341,7 @@ namespace Microsoft.FeatureManagement.FeatureFilters
         private static bool IsDurationCompliantWithDaysOfWeek(TimeSpan duration, int interval, IEnumerable<DayOfWeek> daysOfWeek, DayOfWeek firstDayOfWeek)
         {
             Debug.Assert(interval > 0);
+            Debug.Assert(daysOfWeek.Count() > 0);
 
             if (daysOfWeek.Count() == 1)
             {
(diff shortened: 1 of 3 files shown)


src/Microsoft.FeatureManagement/FeatureFilters/Recurrence/RecurrenceValidator.cs
```diff
@@ -341,49 +341,38 @@ private static bool TryValidateNumberOfOccurrences(TimeWindowFilterSettings sett
...
 
-            DateTime prev = DateTime.MinValue;
+            DayOfWeek firstDay = sortedDaysOfWeek.First(); // the closest occurrence day to the first day of week
+
+            DayOfWeek prev = firstDay;
 
             TimeSpan minGap = TimeSpan.FromDays(DaysPerWeek);
 
-            foreach (DayOfWeek dayOfWeek in sortedDaysOfWeek)
+            foreach (DayOfWeek dayOfWeek in sortedDaysOfWeek.Skip(1))
```
rossgrambo, line 359: I don't like this Skip syntax- it seems less readable, could you add a comment why we skip the first? 

I'd prefer we don't do a skip and explicitly check `if prev == firstDay` or `if prev != null` instead.

src/Microsoft.FeatureManagement/FeatureFilters/Recurrence/RecurrenceValidator.cs
```diff
@@ -413,7 +402,7 @@ private static int CalculateWeeklyDayOffset(DayOfWeek day1, DayOfWeek day2)
         /// </summary>
         private static List<DayOfWeek> SortDaysOfWeek(IEnumerable<DayOfWeek> daysOfWeek, DayOfWeek firstDayOfWeek)
         {
-            List<DayOfWeek> result = daysOfWeek.ToList();
+            List<DayOfWeek> result = daysOfWeek.Distinct().ToList(); // dedup
```
rossgrambo, line 405: Could we take a Set instead so we don't need to Distinct it?

rossgrambo (approved):

src/Microsoft.FeatureManagement/FeatureFilters/Recurrence/RecurrenceValidator.cs
zhiyuanliang-ms: Change to 

``` C#
DayOfWeek firstDay = sortedDaysOfWeek.First(); // the closest occurrence day to the first day of week

DayOfWeek prev = firstDay;

TimeSpan minGap = TimeSpan.FromDays(DaysPerWeek);

foreach (DayOfWeek dayOfWeek in sortedDaysOfWeek)
{
    if (dayOfWeek != prev)
    {
        TimeSpan gap = TimeSpan.FromDays(CalculateWeeklyDayOffset(dayOfWeek, prev));

        if (gap < minGap)
        {
            minGap = gap;
        }

        prev = dayOfWeek;
    }
}
```

src/Microsoft.FeatureManagement/FeatureFilters/Recurrence/RecurrenceValidator.cs
zhiyuanliang-ms: I personally like the `Distinct` one, it is more readable. If we use a set here, we will either need to write two lines of code (1 line for new set and 1 line for ToList) or we will write `List<DayOfWeek> result = new HashSet<DayOfWeek>(daysOfWeek).ToList();` I feel this single line ugly. I believe the background implementation of `Distinct` is based on hash set. Do you have any other concern about using `Distinct`

zhiyuanliang-ms:
@rossgrambo I am going to merge this pr. If you have any further concern about what we've discussed in the comments, please let me know.

PR #525, opened 2025-01-15
Avoid using boolean.ToString and enum.ToString

zhiyuanliang-ms:
## Why this PR?

Currently, we put `VariantAssignementReason` enum in the activity tags. [ref](https://github.com/microsoft/FeatureManagement-Dotnet/blob/main/src/Microsoft.FeatureManagement/FeatureManager.cs#L387)

When putting it into app insights telemetry, we call `ToString` on Enum type, this is not a good practice. Because, code refactoring may cause the value in telemetry changed.

We also call `ToString` on boolean, which will produce "True" instead of "true". Since all of our released packages have the same behavior, we should keep it. But it should be hard-coded. Like JS FM does [ref](https://github.com/microsoft/FeatureManagement-JavaScript/blob/main/src/feature-management/src/telemetry/featureEvaluationEvent.ts#L22) and [ref](https://github.com/microsoft/FeatureManagement-JavaScript/blob/main/src/feature-management/src/featureManager.ts#L222)
Related PR: https://github.com/microsoft/FeatureManagement/pull/26

## Visible Changes

No behavior change

### DIFF
--- a/src/Microsoft.FeatureManagement.Telemetry.ApplicationInsights/ApplicationInsightsEventPublisher.cs
+++ b/src/Microsoft.FeatureManagement.Telemetry.ApplicationInsights/ApplicationInsightsEventPublisher.cs
@@ -51,7 +51,41 @@ namespace Microsoft.FeatureManagement.Telemetry.ApplicationInsights
 
             foreach (var tag in activityEvent.Tags)
             {
-                properties[tag.Key] = tag.Value?.ToString();
+                // FeatureEvaluation event schema: https://github.com/microsoft/FeatureManagement/blob/main/Schema/FeatureEvaluationEvent/FeatureEvaluationEvent.v1.0.0.schema.json
+                if (tag.Key == "VariantAssignmentReason" && tag.Value is VariantAssignmentReason reason)
+                {
+                    switch (reason)
+                    {
+                        case VariantAssignmentReason.None:
+                            properties[tag.Key] = "None";
+                            break;
+                        case VariantAssignmentReason.DefaultWhenDisabled:
+                            properties[tag.Key] = "DefaultWhenDisabled";
+                            break;
+                        case VariantAssignmentReason.DefaultWhenEnabled:
+                            properties[tag.Key] = "DefaultWhenEnabled";
+                            break;
+                        case VariantAssignmentReason.User:
+                            properties[tag.Key] = "User";
+                            break;
+                        case VariantAssignmentReason.Group:
+                            properties[tag.Key] = "Group";
+                            break;
+                        case VariantAssignmentReason.Percentile:
+                            properties[tag.Key] = "Percentile";
+                            break;
+                        default:
+                            throw new ArgumentOutOfRangeException(nameof(activityEvent), "The variant assignment reason is unrecognizable.");
+                    }
+                }
+                else if (tag.Key == "Enabled" && tag.Value is bool enabled)
+                {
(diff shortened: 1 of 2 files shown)


src/Microsoft.FeatureManagement.Telemetry.ApplicationInsights/ApplicationInsightsEventPublisher.cs
```diff
@@ -51,7 +51,41 @@ private void HandleFeatureFlagEvent(ActivityEvent activityEvent)
...
+                        case VariantAssignmentReason.Percentile:
+                            properties[tag.Key] = "Percentile";
+                            break;
+                        default:
+                            throw new ArgumentOutOfRangeException(nameof(activityEvent), "The variant assignment reason is unrecognizable.");
+                    }
+                }
+                else if (tag.Key == "Enabled" && tag.Value is bool enabled)
+                {
+                    properties[tag.Key] = enabled ? "True" : "False";
```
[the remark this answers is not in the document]
zhiyuanliang-ms: This is on purpose. See the PR description

src/Microsoft.FeatureManagement.Telemetry.ApplicationInsights/ApplicationInsightsEventPublisher.cs
```diff
@@ -51,7 +51,41 @@ private void HandleFeatureFlagEvent(ActivityEvent activityEvent)
 
             foreach (var tag in activityEvent.Tags)
             {
-                properties[tag.Key] = tag.Value?.ToString();
+                // FeatureEvaluation event schema: https://github.com/microsoft/FeatureManagement/blob/main/Schema/FeatureEvaluationEvent/FeatureEvaluationEvent.v1.0.0.schema.json
+                if (tag.Key == "VariantAssignmentReason" && tag.Value is VariantAssignmentReason reason)
```
jimmyca15, line 55:
```suggestion
                if (tag.Value is VariantAssignmentReason reason)
```

Doesn't seem necessary. to check for the key's value.

src/Microsoft.FeatureManagement.Telemetry.ApplicationInsights/ApplicationInsightsEventPublisher.cs
```diff
@@ -51,7 +51,41 @@ private void HandleFeatureFlagEvent(ActivityEvent activityEvent)
...
+                            properties[tag.Key] = "Group";
+                            break;
+                        case VariantAssignmentReason.Percentile:
+                            properties[tag.Key] = "Percentile";
+                            break;
+                        default:
+                            throw new ArgumentOutOfRangeException(nameof(activityEvent), "The variant assignment reason is unrecognizable.");
+                    }
+                }
+                else if (tag.Key == "Enabled" && tag.Value is bool enabled)
```
jimmyca15, line 81:
```suggestion
                else if (tag.Value is bool enabled)
```

src/Microsoft.FeatureManagement.Telemetry.ApplicationInsights/ApplicationInsightsEventPublisher.cs
jimmyca15: I don't think we should have different serialization patterns for different VariantAssignmentReasons, or booleans.

jimmyca15 (approved):

jimmyca15 (approved):

PR #528, opened 2025-02-07
Do not use DI for console app example

zhiyuanliang-ms:
## Why this PR?

Showcase the non-DI usage of feature management

## Visible Changes

Update the console app examples

### DIFF
--- a/examples/ConsoleApp/Program.cs
+++ b/examples/ConsoleApp/Program.cs
@@ -2,7 +2,6 @@
 // Licensed under the MIT license.
 //
 using Microsoft.Extensions.Configuration;
-using Microsoft.Extensions.DependencyInjection;
 using Microsoft.FeatureManagement;
 
 //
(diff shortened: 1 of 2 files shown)


examples/ConsoleApp/Program.cs
```diff
@@ -11,45 +10,37 @@
     .AddJsonFile("appsettings.json")
     .Build();
 
-//
-// Setup application services + feature management
-IServiceCollection services = new ServiceCollection();
+IFeatureDefinitionProvider featureDefinitionProvider = new ConfigurationFeatureDefinitionProvider(configuration);
```
jimmyca15, line 13: I think inlining this would be even better. 

```
var featureManager = new FeatureManager(new ConfigurationFeatureDefinitionProvider(configuration))
```

Shortens the sample even more.

jimmyca15 (approved):

PR #535, opened 2025-05-08
Dedupe when returning all feature names

### LINKED ISSUE #534: GetFeatureNames can return duplicates
In experimenting with transitioning to the MS feat mgmt schema, I added configuration for it under `feature_management` (alongside existing configuration under `FeatureManagement`). 

In doing so, I noticed that `GetFeatureNamesAsync` returned the name of that feature multiple times. That actually broke this code we had in one of our legacy apps:
```csharp
            var featureNamesEnumerable = _featureManager.GetFeatureNamesAsync();
            await featureNamesEnumerable.ForEachAwaitAsync(async featureName =>
            {
                var isEnabled = await _featureManager.IsEnabledAsync(featureName);
                featureFlags.Add(featureName, isEnabled);
            });
```
(note, needs System.Linq.Async nuget package).

Given that `IsEnabledAsync` digs into the provider and checks for MS Feature Mgmt schema config first, before falling back to the .NET Feature Mgmt schema config after, i.e. it doesn't matter that there's dupe config, it feels a bit unexpected to see the feature name appear twice. 

```csharp
            var featureNamesEnumerable = _featureManager.GetFeatureNamesAsync().Distinct();
            // ...
```

Is it worth doing that the `GetFeatureNamesAsync` implementation? Happy to submit a PR if that's approved, as I also need to submit my other PR for the MVC tag helper and can knock both PRs out in one go.

benmccallum:
## Why this PR?

Closes #534

## Visible Changes

For developers using a `IFeatureDefinitionProvider` that can return multiple `FeatureDefinition` items with the same `Name`, as is the case with the built-in `Microsoft.FeatureManagement.ConfigurationFeatureDefinitionProvider` (which sources from the MS schema and the DotNet schema), `FeatureManagement::GetFeatureNamesAsync` will no longer return duplicates if a feature is configured multiple times (i.e. in each of those schema). 

This will help folks wanting to transition over gracefully, but _could_ be considered a breaking change to current behaviour.

Note: Duplicate feature names in the one schema is already de-duped, as a test I added proved.

### DIFF
--- a/src/Microsoft.FeatureManagement/FeatureManager.cs
+++ b/src/Microsoft.FeatureManagement/FeatureManager.cs
@@ -196,11 +196,17 @@ namespace Microsoft.FeatureManagement
         /// <returns>An enumerator which provides asynchronous iteration over the feature names registered in the feature manager.</returns>
         public async IAsyncEnumerable<string> GetFeatureNamesAsync([EnumeratorCancellation] CancellationToken cancellationToken = default)
         {
+            var featureNamesReturned = new HashSet<string>();
             await foreach (FeatureDefinition featureDefinition in _featureDefinitionProvider.GetAllFeatureDefinitionsAsync().ConfigureAwait(false))
             {
                 cancellationToken.ThrowIfCancellationRequested();
 
-                yield return featureDefinition.Name;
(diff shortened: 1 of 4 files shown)


benmccallum:

Issue #546, opened 2025-06-30
AddFeatureManagement doesn't Register a Custom Configuration


daniell-reeco:
Nuget version affected: **4.2.0** (worked in 4.0.0 until nuget updated)

In my Program.cs:
`builder.Services.AddFeatureManagement(builder.Configuration.GetSection("FeatureFlags"));`

`GetSection` confirmed to retrieve the configuration from appsettings.json which is:
`"FeatureFlags": {
    "UseReecoUsers": true
  }`

but when calling IsEnabledAsync with the manager I received the following log:
"The feature definition for the feature 'UseReecoUsers' was not found"

When changing the feature flag section to:
`"FeatureManagement": {
    "UseReecoUsers": true
  }`

and changing the line in Program.cs to:
`builder.Services.AddFeatureManagement(/*builder.Configuration.GetSection("FeatureFlags")*/);`

it identifed the feature flag.

micheltol:
I can confirm this issue. It worked with version 4.1.0

zhiyuanliang-ms:
Hi, @micheltol @daniell-reeco 
I am sorry about this issue. This is an oversight. It is a bug introduced in 4.2.0.

In 4.2.0, we added an enhancement #536 that we will merge [Microsoft feature flags](https://learn.microsoft.com/en-us/azure/azure-app-configuration/feature-management-dotnet-reference#feature-flag-declaration) from different configuration sources.

This enhancement will make the operation to [GetMicrosoftFeatureDefinitionSections](https://github.com/microsoft/FeatureManagement-Dotnet/blob/4.2.0/src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs#L217) becomes more time-consuming. 
Before 4.2.0, we will get all configuration sections for feature definitions and find the feature flags among them each time when there is no cache.  [code](https://github.com/microsoft/FeatureManagement-Dotnet/blob/4.1.0/src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs#L95)
[...]

PR #547, opened 2025-06-30
Bug fix - Respect root configuration fallback

zhiyuanliang-ms:
## Why this PR?

Fixed a bug introduced in 4.2.0: #546

### DIFF
--- a/src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
+++ b/src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
@@ -23,11 +23,12 @@ namespace Microsoft.FeatureManagement
         // IFeatureDefinitionProviderCacheable interface is only used to mark this provider as cacheable. This allows our test suite's
         // provider to be marked for caching as well.
         private readonly IConfiguration _configuration;
-        private Lazy<IEnumerable<IConfigurationSection>> _dotnetFeatureDefinitionSections;
-        private Lazy<IEnumerable<IConfigurationSection>> _microsoftFeatureDefinitionSections;
+        private IEnumerable<IConfigurationSection> _dotnetFeatureDefinitionSections;
+        private IEnumerable<IConfigurationSection> _microsoftFeatureDefinitionSections;
         private readonly ConcurrentDictionary<string, Task<FeatureDefinition>> _definitions;
         private IDisposable _changeSubscription;
         private int _stale = 0;
+        private int _initialized = 0;
         private readonly Func<string, Task<FeatureDefinition>> _getFeatureDefinitionFunc;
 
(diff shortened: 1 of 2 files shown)


src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
```diff
@@ -163,6 +165,28 @@ public async IAsyncEnumerable<FeatureDefinition> GetAllFeatureDefinitionsAsync()
...
+                        _dotnetFeatureDefinitionSections = dotnetFeatureDefinitionSections;
+
+                        _microsoftFeatureDefinitionSections = microsoftFeatureDefinitionSections;
+
+                        Interlocked.Exchange(ref _initialized, 1);
+                    }
+                }
+            }
+        }
+
```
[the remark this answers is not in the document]
zhiyuanliang-ms: Educate me what is Lazy<>

zhiyuanliang-ms:
Add a testcaste that fails before the fix and succeed after the fix

tests/Tests.FeatureManagement/FeatureManagementTest.cs
```diff
@@ -90,7 +90,7 @@ public async Task ReadsTopLevelConfiguration()
 
             IFeatureManager featureManager = serviceProvider.GetRequiredService<IFeatureManager>();
 
-            //Assert.True(await featureManager.IsEnabledAsync("FeatureX"));
+            Assert.True(await featureManager.IsEnabledAsync("FeatureX"));
```
zhiyuanliang-ms, line 93: This is a mistake we made in #470

src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
```diff
@@ -96,9 +96,9 @@ public Task<FeatureDefinition> GetFeatureDefinitionAsync(string featureName)
             {
                 _definitions.Clear();
 
-                _dotnetFeatureDefinitionSections = GetDotnetFeatureDefinitionSections();
+                _dotnetFeatureDefinitionSections = new Lazy<IEnumerable<IConfigurationSection>>(GetDotnetFeatureDefinitionSections);
```
jimmyca15, line 99: Lazy uses locks. Do we need to lock? Locking avoids multi-threaded scenarios from all executing the factory for finding the definition sections, but lock also holds up threads. If we don't need the lock I'd opt to not have it. And if we don't need a lock we can either use Lazy with ThreadSafetyMode set to PublicationOnly, or we could use EnsureInit, but a simplified version that doesn't use any locks.

zhiyuanliang-ms: Update to use "EnsureInit" pattern without lock.

Issue #507, opened 2024-09-30
Difference in behavior between .NET and Microsoft schemas with configuration merging


CuddleBunny:
This isn't necessarily an issue, but it should probably be called out somewhere: when switching from the .NET schema to the Microsoft schema you lose the ability to spread definitions across multiple configuration files (ie. appsettings.json and appsettings.Prod.json) because `feature_management:feature_flags` is an array which causes additional configuration sources to overwrite the original. This isn't the case with the .NET schema because the collection is an object and merges fine.

rossgrambo:
Hey @CuddleBunny, thank you for calling this out and following up on it.
 
It's unfortunate that this change to unify our libraries caused this to work differently. 

The issue with Configuration merging is discussed [here](https://github.com/dotnet/runtime/issues/36569). Sadly, it's upstream of FeatureManagement- as this library reads whatever is made available by IConfiguration. 

Theres two workarounds I'm aware of:
### Define your own `IFeatureDefinitionProvider`

I know this could solve the issue if you manually read the files and merge them how you see appropriate. Our default implementation reads from `IConfiguration` but you don't have to in your version.

It might be possible to have your implementation call each `JsonConfigurationProvider` within the `Configuration` separately so you can handle the merge yourself and don't have to re-read these files. If this is possible and not too heavy it could be something we consider adding to the default `IFeatureDefintionProvider`. But I think that's unlikely because there's not one way to merge that works for everyone.

### `:#` syntax (Not recommended)
Using `:#` to override arrays appropriately. 
  When defining an array in json to be used in IConfiguration, you can define it like:
  
  ```json
  "feature_management": {
    "feature_flags:0": {
      "id": "OnTestFeature",
      "enabled": true
    }
  }
  ```

In your example, you could define your `appsettings.prod.json` as
```json
{
  "feature_management": {
    "feature_flags:2": {
      "id": "ProdOnlyThing",
      "enabled": true
    }
  }
}
```

Unfortunately, this is finicky to line things up when you're trying to override correctly. 

If your only goal is to merge I believe you start at a large number (like `:1000`) in all of your prod flags and they would always append to the end of the other flags. This is not an intended feature so there's no guarantee that it'll continue to work but it seems to work for me at first glance.

CuddleBunny:
[...]

PR #551, opened 2025-08-06
Merge Microsoft feature flags

zhiyuanliang-ms:
## Why this PR?

#507 

There is a well-known [pitfall](https://rimdev.io/avoiding-aspnet-core-configuration-pitfalls-with-array-values) when merging array value in configuration system in .NET.

We presented a [solution](https://github.com/microsoft/FeatureManagement-Dotnet/pull/536) in v4.2.1: concatenate feature flag arrays and let the last feature definition win.

However, a [customer](https://github.com/microsoft/FeatureManagement-Dotnet/issues/507#issuecomment-3111322555) raised a concern that this behavior change will break their current usage. The customer’s team distributed  a base appsettings file in the docker image. Developers just want to toggle the feature flag and what they do is to use another configuration source and update the “enabled” field for certain feature flags.

Base appsettings.json:
```js
{
  "feature_management": {
    "feature_flags": [
      {
        "id": "FeatureA",
        "enabled": true,
        "conditions": {
          "client_filters": [ ... ]
        }
      },
      {
        "id": "FeatureB",
        "enabled": false,
        "variants": [ ... ],
        "allocation": { ... }
      }
    ]
  }
}
```

People may want to use the following appsetting.development.json file to toggle the “FeatureB” flag:
```js
{
  "feature_management": {
    "feature_flags": [
      {
        "id": "FeatureA",
        "enabled": true
      },
      {
        "id": "FeatureB",
        "enabled": true // toggle "FeatureB"
      }
    ]
  }
}
```

However, in 4.2.1, our concat ff arrays behavior will produce the following configuration:

```js
{
  "feature_management": {
    "feature_flags": [
      {
        "id": "FeatureA",
        "enabled": true,
        "conditions": {
          "client_filters": [ ... ]
        }
      },
      {
        "id": "FeatureB",
        "enabled": false,
        "variants": [ ... ],
        "allocation": { ... }
      },
      {
        "id": "FeatureA",
        "enabled": true
      },
      {
        "id": "FeatureB",
        "enabled": true
      }
    ]
  }
}
```

The merge result that customer wanted
```js
{
  "feature_management": {
    "feature_flags": [
      {
        "id": "FeatureA",
        "enabled": true,
        "conditions": {
          "client_filters": [ ... ]
        }
      },
      {
        "id": "FeatureB",
        "enabled": true, // toggled
        "variants": [ ... ],
        "allocation": { ... }
      }
    ]
  }
}
```

------------------------------------------------------------------------------

In addition, some customers may have the following usage:
appsettings.json
```js
{
    "feature_management": {
        "feature_flags": [
            {
                "id": "FeatureA",
                "enabled": true,
                "variants": [ ... ],
                "allocation": { ... }
            }
        ]
    }
}
```
appsettings.Production.json
```js
{
    "feature_management": {
        "feature_flags": [
            {
                "id": "FeatureA",
                 // enabled/variant/allocation field is not specified
                "conditions": {  // add condition for production
                    "client_filters": [
                        {
                            "name": "FeatureFilterA"
                        }
                    ]
                }
            }
        ]
    }
}
```
In short, what people is a layering-manner merge behavior. 

## Visible Changes

- This PR is built on #536 and #547. The `ConfigurationFeatureDefinitionProvider` will not only read `feature_flags` array from different `ConfigurationProvider`, but also it will maintain a dictionary where key is feature name and value is `List<IConfigurationSection>` that contains feature definition from different configuration source. In this case, `ConfigurationFeatureDefinitionProvider` can merge them in an intuitive manner:

- The later feature definition will override the previous feature defintion. The granularity of the override behavior is at the property level within the FeatureDefinition class. For example:
  ```js
  "feature_flags": [
      {
          "id": "FeatureA",
          "enabled": false，
          "variants": [
              {
                  "name": "Variant1",
                  "configuration_value": "Value1"
              },
              {
                  "name": "Variant2",
                  "configuration_value": "Value2"
              }
          ]
      },
      {
          "id": "FeatureA",
          "enabled": true，
          "variants": [
              {
                  "name": "Variant2",
                  "configuration_value": "Value2-updated"
              },
              {
                  "name": "Variant3",
                  "configuration_value": "Value3"
              },
          ],
          "conditions": {
              "client_filters": [...]
          }
      }
  ]
  ```
  The above configuration will produce such a feature definition after merging:

  ```js
  {
      "id": "FeatureA",
      "enabled": true，
      "variants": [
          {
              "name": "Variant1",
              "configuration_value": "Value1"
          },
          {
              "name": "Variant2",
              "configuration_value": "Value2-updated"
          },
          {
              "name": "Variant3",
              "configuration_value": "Value3"
          }
      ],
      "conditions": {
          "client_filters": [...]
      }
  }
  ```
  
  For more information, please go to testcases.

- A lock is used in the `EnsureInit` method.

### DIFF
--- a/src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
+++ b/src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
@@ -400,6 +400,96 @@ namespace Microsoft.FeatureManagement
                     }
                 }
 
+                if (additionalDefinition.Allocation != null)
+                {
+                    if (mergedAllocation == null)
+                    {
+                        mergedAllocation = additionalDefinition.Allocation;
+                    }
+                    else
+                    {
+                        // Merge allocation properties, only override if not null
+                        if (!string.IsNullOrEmpty(additionalDefinition.Allocation.DefaultWhenEnabled))
+                        {
+                            mergedAllocation.DefaultWhenEnabled = additionalDefinition.Allocation.DefaultWhenEnabled;
+                        }
+
+                        if (!string.IsNullOrEmpty(additionalDefinition.Allocation.DefaultWhenDisabled))
+                        {
+                            mergedAllocation.DefaultWhenDisabled = additionalDefinition.Allocation.DefaultWhenDisabled;
+                        }
+
+                        if (!string.IsNullOrEmpty(additionalDefinition.Allocation.Seed))
+                        {
+                            mergedAllocation.Seed = additionalDefinition.Allocation.Seed;
+                        }
+
+                        if (additionalDefinition.Allocation.User != null)
+                        {
+                            var mergedUserAllocations = new List<UserAllocation>(mergedAllocation.User ?? Enumerable.Empty<UserAllocation>());
+
+                            foreach (UserAllocation userAllocation in additionalDefinition.Allocation.User)
+                            {
+                                int existingIndex = mergedUserAllocations.FindIndex(u => string.Equals(u.Variant, userAllocation.Variant, StringComparison.OrdinalIgnoreCase));
+
+                                if (existingIndex >= 0)
+                                {
+                                    mergedUserAllocations[existingIndex] = userAllocation;
+                                }
+                                else
+                                {
+                                    mergedUserAllocations.Add(userAllocation);
+                                }
+                            }
+
+                            mergedAllocation.User = mergedUserAllocations;
+                        }
+
+                        if (additionalDefinition.Allocation.Group != null)
+                        {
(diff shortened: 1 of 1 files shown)


juniwang:
If the scenarios described in the samples(add condition for production, new variant Variant3 added after merging) are reasonable, it also makes sense to remove condition/variant after merging, which is not supported by the new proposal, but supported by the current impl of overriding the whole FF.

zhiyuanliang-ms:
> it also makes sense to remove condition/variant after merging

This is against the normal pattern of native .net configuration system. What our customers want is the native .net configuration behavior of layering configurations.

The solution we offered in 4.2.1 is more like that we proposed a new way to tackle the issue of merging array values. But before we present the solution, some customers they have their own solutions which utilized the native .net configuration layering behavior. So we want to improve our solution based on real customers' usage.

Issue #550, opened 2025-07-17
The latest version 4.2.1 break the possibility of enabling/disabling a feature via environment variables.


Quentin-IMA:
Hello,

I recently tried the new version 4.2.1 but it breaks the way I manage whether or not to enable features in my container (k8s) via an environment variable.

I dived into the code and found the problem [here](https://github.com/microsoft/FeatureManagement-Dotnet/blob/3b2e5510c73e03c94970843d6232c35f3b981823/src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs#L197).

When I add an env variable like this: “feature_management:feature_flags:0:enabled”, this entry is ignored.

sample app : 

```csharp
using Microsoft.FeatureManagement;

var builder = WebApplication.CreateBuilder(args);

builder.Configuration
    .AddEnvironmentVariables()
    .AddInMemoryCollection(new Dictionary<string, string?>(StringComparer.OrdinalIgnoreCase)
    {
        ["feature_management:feature_flags:0:enabled"] = bool.FalseString
    });

builder.Services.AddFeatureManagement();

var app = builder.Build();

var fm = app.Services.GetRequiredService<IFeatureManager>();

var isEnabled = await fm.IsEnabledAsync("EnablePersistence");
    
if (isEnabled)
    throw new InvalidOperationException("Feature 'EnablePersistence' should not be enabled.");

```

appsettings :

```json
"feature_management": {
  "feature_flags": [
    {
      "id": "EnablePersistence",
      "enabled": true
    }
  ]
},
```

jimmyca15:
> but was this behavior expected, or is it more of an unintended side effect?

It's somewhere in the middle. It was a side-effect, but our expectation was that matching flags across different configuration providers by matching array indices would not have been an approach that was in use due to the complexities of having to make sure that the array items are in the exact same position. 

However, it appears that it may have affected your application as well as [this issue](https://github.com/microsoft/FeatureManagement-Dotnet/issues/507#issuecomment-3070683675). That issue does suggest possibly adding a switch to merge flags by id, but I'm wondering if in your view the new method of specifying the id might actually be more durable?

Quentin-IMA:
I need to “duplicate” the environment variables on each of my projects using feature flags in order to add the id. It's not very complicated, but it changes the way you usually overload appsetting variables.

sander1095:
I just ran into this issue and had to spend a good few hours to figure this out.
[...]

PR #552, opened 2025-08-15
Add ConfigurationFeatureDefinitionProviderOptions

zhiyuanliang-ms:
## Why this PR?

#507 #550 

### DI

```
services.Configure<ConfigurationFeatureDefinitionProviderOptions>(o =>
{
    o.CustomConfigurationMergingEnabled = true;
});
```

### Direct

```
var FeatureManager = new FeatureManager(
        configuration,
        new ConfigurationFeatureDefinitionProviderOptions
        {
            CustomConfigurationMergingEnabled = true
        });
```

### New type

FeatureManagement options shouldn't be used because this is for configuration based feature definition provider only. We need a new type to customize the ConfigurationFeatureDefinitionProvider

```
public class ConfigurationFeatureDefinitionProviderOptions
{
    bool CustomConfigurationMergingEnabled { get; set; }
}
```

### DIFF
--- a/tests/Tests.FeatureManagement/FeatureManagementTest.cs
+++ b/tests/Tests.FeatureManagement/FeatureManagementTest.cs
@@ -467,14 +467,13 @@ namespace Tests.FeatureManagement
             //
             // DI usage
             var services1 = new ServiceCollection();
+            services1
+                .AddSingleton(configuration2)
+                .AddFeatureManagement();
             services1.Configure<ConfigurationFeatureDefinitionProviderOptions>(o =>
             {
                 o.CustomConfigurationMergingEnabled = true;
             });
-
-            services1
-                .AddSingleton(configuration2)
-                .AddFeatureManagement();
             ServiceProvider serviceProvider1 = services1.BuildServiceProvider();
             IFeatureManager featureManager5 = serviceProvider1.GetRequiredService<IFeatureManager>();
 
@@ -485,11 +484,6 @@ namespace Tests.FeatureManagement
             Assert.False(await featureManager5.IsEnabledAsync("Feature2"));
 
             var services2 = new ServiceCollection();
-            services2.Configure<ConfigurationFeatureDefinitionProviderOptions>(o =>
-            {
-                o.CustomConfigurationMergingEnabled = false;
-            });
-
             services2
                 .AddSingleton(configuration2)
                 .AddFeatureManagement();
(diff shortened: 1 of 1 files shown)


jimmyca15:
I get your point about expectations of a .NET developer. With respect to that I agree that making this opt-out will be a better option. We still need a major version because despite 4.2.1 being delisted, it was still released.

Regarding the API, due to the extensibility model of feature definition providers i would not advise that we special case a single provider by putting switches in the main feature management options. Instead the proposal i had limits the exposure to those using the Configuration based feature definition provider, which happens to be the default built-in provider. This is a similar pattern we've already with targeting options.

src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProviderOptions.cs
```diff
@@ -0,0 +1,64 @@
...
+    ///     "feature_flags": [
+    ///       {
+    ///         "id": "feature2",
+    ///         "enabled": true
+    ///       }
+    ///     ]
+    ///   }
+    /// }
+    /// 
+    /// With custom merging (default behavior):
```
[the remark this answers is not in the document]
zhiyuanliang-ms: It is a mistake. Updated

sander1095:
It seems like I would still not be able to configure this with `services.AddFeatureManagement(x => x...)`, which makes this feature very undiscoverable to me. Outside of that, LGTM.

src/Microsoft.FeatureManagement/ServiceCollectionExtensions.cs
```diff
@@ -145,7 +151,11 @@ private static void AddCommonFeatureManagementServices(IServiceCollection servic
 
             services.AddMemoryCache();
 
-            services.TryAddSingleton<IFeatureDefinitionProvider, ConfigurationFeatureDefinitionProvider>();
+            services.TryAddSingleton<IFeatureDefinitionProvider>(sp =>
+                new ConfigurationFeatureDefinitionProvider(
+                    sp.GetRequiredService<IConfiguration>(),
+                    sp.GetRequiredService<IOptions<ConfigurationFeatureDefinitionProviderOptions>>().Value)
```
jimmyca15, line 157: I don't think it's required. `GetService` should suffice.

jimmyca15 (approved):

src/Microsoft.FeatureManagement/ServiceCollectionExtensions.cs
zhiyuanliang-ms: We call `GetRequiredService` for `FeatureManagementOptions` and `TargetingEvaluationOptions`. 
image

I think it is fine to use GetRequiredService here. We called `services.AddLogging()` in advance where Options infrastructure is registered.

image

jimmyca15 (approved):

PR #578, opened 2026-01-05
OpenTelemetry and Azure Monitor Exporter support

rossgrambo:
## Why this PR?

This change adds support for using OpenTelemetry with Azure Monitor Exporter instead of Application Insights.

## Visible Changes

1. There is a new package, `Microsoft.FeatureManagement.Telemetry.AzureMonitor`. This package looks very similar to `Microsoft.FeatureManagement.Telemetry.ApplicationInsights` and works similarly.
1. There is a new example. Named `VariantAndAzureMonitorDemo`. It is a clone of VariantAndTelemetry demo but uses OTel and Azure Monitor.

## Migrating from .ApplicationInsights to .AzureMonitor 
Package
```csharp
Microsoft.FeatureManagement.Telemetry.AzureMonitor
// instead of
Microsoft.FeatureManagement.Telemetry.ApplicationInsights
```

Builder extension
```csharp
    .AddAzureMonitorTelemetry();
// instead of
    .AddApplicationInsightsTelemetry();
```

Emitting events
```csharp
_logger.LogInformation("{microsoft.custom_event.name} {ImageRating}", "Vote", rating);
// instead of
_telemetry.TrackEvent(
  "Vote",
  properties: null,
  metrics: new Dictionary<string, double>
  {
    { "ImageRating", rating }
  });
```
*Note: Even better, use High-performance logging, like the `LoggerExtension.cs` in the example.


examples/VariantAndAzureMonitorDemo/wwwroot/lib/jquery-validation-unobtrusive/jquery.validate.unobtrusive.js
```diff
@@ -0,0 +1,432 @@
...
+                    .add($selector.find("form"))
+                    .has("[data-val=true]");
+
+            $selector.find("[data-val=true]").each(function () {
+                $jQval.unobtrusive.parseElement(this, true);
+            });
+
+            $forms.each(function () {
+                var info = validationInfo(this);
+                if (info) {
```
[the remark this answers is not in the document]
rossgrambo: Ignoring since this is default Visual Studio jquery templating code- not ours.

src/Microsoft.FeatureManagement.Telemetry.AzureMonitor/TargetingActivityProcessor.cs
```diff
@@ -0,0 +1,38 @@
...
+        /// </summary>
+        /// <param name="activity">The activity that ended.</param>
+        public override void OnEnd(Activity activity)
+        {
+            if (activity == null)
+            {
+                return;
+            }
+
+            string targetingId = activity.Baggage.FirstOrDefault(t => t.Key == TargetingIdKey).Value;
```
thompson-tomo, line 30: Based on https://opentelemetry.io/docs/specs/semconv/feature-flags/feature-flags-events/#evaluation-event shouldn't this value already be a tag hence u could use that instead of baggage

src/Microsoft.FeatureManagement.Telemetry.AzureMonitor/AzureMonitorEventPublisher.cs
```diff
@@ -0,0 +1,125 @@
...
+//
+using Microsoft.Extensions.Logging;
+using System.Diagnostics;
+
+namespace Microsoft.FeatureManagement.Telemetry.AzureMonitor
+{
+    /// <summary>
+    /// Listens to <see cref="Activity"/> events from feature management and sends them to Azure Monitor via structured logging.
+    /// </summary>
+    internal sealed class AzureMonitorEventPublisher : IDisposable
```
thompson-tomo, line 12: Wouldn't it be easier to implement this as otel processor/exporter?

src/Microsoft.FeatureManagement.Telemetry.AzureMonitor/TargetingLogProcessor.cs
```diff
@@ -0,0 +1,43 @@
...
+        /// </summary>
+        /// <param name="logRecord">The log record that ended.</param>
+        public override void OnEnd(LogRecord logRecord)
+        {
+            if (logRecord == null)
+            {
+                return;
+            }
+
+            string targetingId = Activity.Current?.Baggage.FirstOrDefault(t => t.Key == TargetingIdKey).Value;
```
thompson-tomo, line 31: Based on https://opentelemetry.io/docs/specs/semconv/feature-flags/feature-flags-events/#evaluation-event shouldn't this value already be a tag hence u could use that instead of baggage

src/Microsoft.FeatureManagement.Telemetry.AzureMonitor/TargetingActivityProcessor.cs
rossgrambo: This code is solving a different problem. The idea is that we need to get the TargetingId onto other activities besides the feature evaluation event. For example

Incoming API Request (Activity 1 - TargetingHttpContextMiddleware sets TargetingId as baggage)
| Feature Evaluation event happens (TargetingId added as ActivityEvent tag)
| Request to other API (Activity 2)
| Response from other API (Activity 2 end - processor promotes baggage to tag)
| misc code
Returned result (Activity 1 end - processor promotes baggage to tag)

Now the Azure Monitor exporter will emit Activity 2 to the "dependencies" table (since it was an outgoing request) and include the custom dimension "TargetingId". Activity 1 is emitted to the "requests" table (with it's targeting id as well)

src/Microsoft.FeatureManagement.Telemetry.AzureMonitor/AzureMonitorEventPublisher.cs
rossgrambo: Good question. 

The primary reason I went this route is that this pattern is already established and working in the ApplicationInsights event publisher. Additionally- we have better control over sampling here since the feature flag evaluation events can't be sampled for reliable results.

But yes, we could rewrite this as a processor in the future.

rossgrambo:
I'm no longer working on this so I'm going to close this for now. I imagine someone on the team will pick up the work later.

Issue #580, opened 2026-02-12
SessionManager does not work when FeatureDefinition is null


apetukhov:
We use a ISessionManager to maintain a consistent flag state across child requests. If somewhere is no FeatureDefinition, the result of checking the flag in one service may differ from the result in another within the same logical operation.

The behavior changed in version 4 of the package.

zhiyuanliang-ms:
Hey, @apetukhov 

I found that we unintentionally introduced a behavior change in v4.

In v3, even if `FeatureDefinition` is null, the [session manager will still be used for evaluation result](https://github.com/microsoft/FeatureManagement-Dotnet/blob/3.0.0/src/Microsoft.FeatureManagement/FeatureManager.cs#L82).

In v4, the [session manager will be skipped](https://github.com/microsoft/FeatureManagement-Dotnet/blob/main/src/Microsoft.FeatureManagement/FeatureManager.cs#L264) if `FeatureDefinition` is null.

This is an oversight. Sorry for the inconvenience.

zhiyuanliang-ms:
@jimmyca15 @rossgrambo 
I think `ISessionManager` should be also able to cache variant.

jimmyca15:
> [@jimmyca15](https://github.com/jimmyca15) [@rossgrambo](https://github.com/rossgrambo) I think `ISessionManager` should be also able to cache variant.

It's a breaking change to add to the existing interface, but we could add a new one. But it's a separate issue.

PR #581, opened 2026-02-13
Fix session manager bug

zhiyuanliang-ms:
## Why this PR?

#580

### DIFF
--- a/src/Microsoft.FeatureManagement/FeatureManager.cs
+++ b/src/Microsoft.FeatureManagement/FeatureManager.cs
@@ -371,6 +371,20 @@ namespace Microsoft.FeatureManagement
                     FeatureEvaluationTelemetry.Publish(evaluationEvent, Logger);
                 }
             }
+            else if (_sessionManagers != null)
+            {
+                foreach (ISessionManager sessionManager in _sessionManagers)
+                {
+                    bool? readSessionResult = await sessionManager.GetAsync(feature).ConfigureAwait(false);
+
+                    if (readSessionResult.HasValue)
+                    {
+                        evaluationEvent.Enabled = readSessionResult.Value;
+
+                        break;
+                    }
+                }
+            }
 
             return evaluationEvent;
         }
--- a/src/Microsoft.FeatureManagement/FeatureManager.cs
+++ b/src/Microsoft.FeatureManagement/FeatureManager.cs
@@ -355,14 +355,6 @@ namespace Microsoft.FeatureManagement
                     }
                 }
 
-                if (_sessionManagers != null)
-                {
-                    foreach (ISessionManager sessionManager in _sessionManagers)
-                    {
-                        await sessionManager.SetAsync(evaluationEvent.FeatureDefinition.Name, evaluationEvent.Enabled).ConfigureAwait(false);
-                    }
-                }
-
                 // Only add an activity event if telemetry is enabled for the feature and the activity is valid
                 if (telemetryEnabled &&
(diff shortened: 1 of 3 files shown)


src/Microsoft.FeatureManagement/FeatureManager.cs
```diff
@@ -371,6 +371,20 @@ private async ValueTask<EvaluationEvent> EvaluateFeature<TContext>(string featur
...
+                    bool? readSessionResult = await sessionManager.GetAsync(feature).ConfigureAwait(false);
+
+                    if (readSessionResult.HasValue)
+                    {
+                        evaluationEvent.Enabled = readSessionResult.Value;
+
+                        break;
+                    }
+                }
+            }
```
apetukhov, line 379: I also think that if there is no value in the session manager, then it should be written to what the feature manager returns, i.e. the default value.

https://github.com/microsoft/FeatureManagement-Dotnet/blob/57f5f9363a84ca4a1f870484c82a38cb4285fdf8/src/Microsoft.FeatureManagement/FeatureManager.cs#L275-L278

zhiyuanliang-ms: The current logic is that if feature definition is not null, then feature manager will do the feature evaluation and the result will be set to session manager.

Did I miss something?

apetukhov: It doesn't matter whether a feature definition exists; if the feature manager returns a result and doesn't throw an exception, the same result must be set in the session manager.

Test case
1. Check the boolean feature x in the feature manager.
2. There is no definition for feature x.
3. There is no previously saved result in the session manager.
4. The default result "false" stored in the session manager
5. The feature manager returns the default result "false"

Step 4 is currently missing.

zhiyuanliang-ms: > It doesn't matter whether a feature definition exists; if the feature manager returns a result and doesn't throw an exception, the same result must be set in the session manager.

I believe this is the [current behavior](https://github.com/microsoft/FeatureManagement-Dotnet/blob/main/src/Microsoft.FeatureManagement/FeatureManager.cs#L362) (when feature definition is not null). 

We introduced the bug in v4. The bug is when feature definition is null, the result won't be set to session manage, I explained it in this [comment](https://github.com/microsoft/FeatureManagement-Dotnet/issues/580#issuecomment-3894827690)

apetukhov:
```cs
[Fact]
        public async Task SetResultInSessionManagerWhenFeatureDefinitionIsNull()
        {
            IConfiguration config = new ConfigurationBuilder().AddJsonFile("appsettings.json").Build();

            var services = new ServiceCollection();

            ISessionManager sessionManager = new TestSessionManager();

            services
                .AddSingleton(config)
                .AddSingleton<ISessionManager>(sessionManager)
                .AddFeatureManagement();

            ServiceProvider serviceProvider = services.BuildServiceProvider();

           // session manager is clean here
            var result = await featureManager.IsEnabledAsync("UnexistedFeature");

           // session manager should store the result after evaluation
             Assert.AreEqual(result, await sessionManager.GetAsync("UnexistedFeature"));
        }
```

zhiyuanliang-ms: I see. Updated. Thanks!

jimmyca15 (approved):

PR #582, opened 2026-03-02
Fix binary incompatibility introduced by 4.4.0

richardszalay:
## Why this PR?

4.4.0 [introduced](https://github.com/microsoft/FeatureManagement-Dotnet/commit/cb49942ef3fc17e839eda0aaefc79853b21cfa99) a change to `ConfigurationFeatureDefinitionProvider`, adding a nullable argument to the existing constructor. This was _source_ compatible, but not binary incompatible and so breaks (via `MissingMethodException`) any libraries that depend on it directly. 

## Visible Changes

The `ConfigurationFeatureDefinitionProviderOptions` constructor argument is separated into an overloaded constructor, restoring the < 4.4.0 constructor signature.

### DIFF
--- a/src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
+++ b/src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
@@ -38,10 +38,18 @@ namespace Microsoft.FeatureManagement
         /// Creates a configuration feature definition provider.
         /// </summary>
         /// <param name="configuration">The configuration of feature definitions.</param>
+        public ConfigurationFeatureDefinitionProvider(IConfiguration configuration) : this(configuration, null)
+        {
+        }
+
+        /// <summary>
+        /// Creates a configuration feature definition provider.
+        /// </summary>
+        /// <param name="configuration">The configuration of feature definitions.</param>
         /// <param name="options">The options for the configuration feature definition provider.</param>
         public ConfigurationFeatureDefinitionProvider(
             IConfiguration configuration,
-            ConfigurationFeatureDefinitionProviderOptions options = null)
+            ConfigurationFeatureDefinitionProviderOptions options)
         {
             _configuration = configuration ?? throw new ArgumentNullException(nameof(configuration));
             _options = options ?? new ConfigurationFeatureDefinitionProviderOptions();


zhiyuanliang-ms:
Hey, @richardszalay Thanks for pointing this out. This is an oversight. Sorry for that.

zhiyuanliang-ms (approved):

src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
```diff
@@ -34,14 +34,22 @@ public sealed class ConfigurationFeatureDefinitionProvider : IFeatureDefinitionP
...
+
         /// <summary>
         /// Creates a configuration feature definition provider.
         /// </summary>
         /// <param name="configuration">The configuration of feature definitions.</param>
         /// <param name="options">The options for the configuration feature definition provider.</param>
         public ConfigurationFeatureDefinitionProvider(
             IConfiguration configuration,
-            ConfigurationFeatureDefinitionProviderOptions options = null)
+            ConfigurationFeatureDefinitionProviderOptions options)
```
zhiyuanliang-ms, line 52: Could you please add a test to verify that DI will pick the ctor with most parameters?

richardszalay: I don't think that's necessary, for two reasons:

1. The library doesn't rely on DI ([ref](https://github.com/microsoft/FeatureManagement-Dotnet/blob/main/src/Microsoft.FeatureManagement/ServiceCollectionExtensions.cs)) to construct ConfigurationFeatureDefinitionProviderOptions - it always new's it up inline
2. The MS DI [Constructor selection rules](https://learn.microsoft.com/en-us/dotnet/core/extensions/dependency-injection/overview#constructor-selection-rules) are quite clear: the constructor with the most arguments is used

zhiyuanliang-ms: Make sense!

zhiyuanliang-ms:
Thank you for your contribution! @richardszalay

src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
jimmyca15: Referencing implementation shouldn't be used to justify tests. Otherwise, we wouldn't need tests, we could just say look at the code, it works.

The thing that needs coverage is if `ConfigurationFeatureDefinitionProviderOptions` are registered, then they are respected.

@zhiyuanliang-ms , I believe that's covered [here](https://github.com/microsoft/FeatureManagement-Dotnet/blob/ed9860d3162bbc5e16cc63fa381b10f7b3c14a82/tests/Tests.FeatureManagement/FeatureManagementTest.cs#L530), can you confirm?

zhiyuanliang-ms: confirmed

PR #589, opened 2026-04-02
Feature filter parameter object

linglingye001:
## Overview

This PR introduces a new `ParameterObject` property to `FeatureFilterEvaluationContext`, enabling custom `IFeatureDefinitionProvider` implementations to supply feature filter settings directly without requiring `IConfiguration`.

## Motivation

Currently, feature filters receive their configuration through `IConfiguration`, which works well for JSON/config file-based feature definitions. However, custom `IFeatureDefinitionProvider` implementations that source feature definitions from alternative backends (e.g., databases, REST APIs, gRPC services) are forced to:

1. Construct an `IConfiguration` object from their native data format
2. Have filters bind that configuration back to strongly-typed settings objects

This round-trip is unnecessary overhead when the provider already has the ability to create strongly-typed settings objects like `TargetingFilterSettings` or `TimeWindowFilterSettings` directly.

### Before: Using `IConfiguration` with `InMemoryCollection`

When implementing a custom `IFeatureDefinitionProvider` that fetches feature definitions from a database, you previously had to construct an `IConfiguration` object to pass filter parameters:

```csharp
// Fetch from database
var dbFeature = await _database.GetFeatureAsync(featureName);

// Must convert to IConfiguration even though we already have the data
var configData = new Dictionary<string, string>
{
    ["Audience:Users:0"] = dbFeature.TargetedUsers[0],
    ["Audience:Users:1"] = dbFeature.TargetedUsers[1],
    ["Audience:Groups:0:Name"] = dbFeature.TargetedGroups[0].Name,
    ["Audience:Groups:0:RolloutPercentage"] = dbFeature.TargetedGroups[0].RolloutPercentage.ToString(),
    ["Audience:DefaultRolloutPercentage"] = dbFeature.DefaultRolloutPercentage.ToString()
};

return new FeatureDefinition
{
    Name = featureName,
    EnabledFor = new[]
    {
        new FeatureFilterConfiguration
        {
            Name = "Targeting",
            Parameters = new ConfigurationBuilder()
                .AddInMemoryCollection(configData)
                .Build()
        }
    }
};
```

This approach has several drawbacks:
- Verbose dictionary key construction with magic strings
- Error-prone index management for arrays
- Unnecessary serialization to key-value pairs just to deserialize back to objects

### After: Using `ParameterObject` Directly

With the new `ParameterObject` property, custom providers can supply strongly-typed settings directly:

```csharp
// Fetch from database
var dbFeature = await _database.GetFeatureAsync(featureName);

// Directly create strongly-typed settings
var targetingSettings = new TargetingFilterSettings
{
    Audience = new TargetingFilterAudience
    {
        Users = dbFeature.TargetedUsers,
        Groups = dbFeature.TargetedGroups.Select(g => new GroupRollout
        {
            Name = g.Name,
            RolloutPercentage = g.RolloutPercentage
        }).ToList(),
        DefaultRolloutPercentage = dbFeature.DefaultRolloutPercentage
    }
};

return new FeatureDefinition
{
    Name = featureName,
    EnabledFor = new[]
    {
        new FeatureFilterConfiguration
        {
            Name = "Targeting",
            ParameterObject = targetingSettings  // Direct assignment!
        }
    }
};
```

## Changes

- **`FeatureFilterEvaluationContext`**: Added `ParameterObject` property to hold strongly-typed settings provided by custom providers
- **Built-in filters updated** (`PercentageFilter`, `TimeWindowFilter`, `ContextualTargetingFilter`): Now check `ParameterObject` first before falling back to `Settings` or binding from `Parameters`

## Precedence Order

When evaluating filter settings:

1. `ParameterObject` (if set) — Direct strongly-typed object from provider
2. `Settings` — Pre-bound settings via `IFilterParametersBinder`
3. `Parameters` — Bind from `IConfiguration`

The expectation is that systems using `ParameterObject` will not set `Parameters`/`IConfiguration`.

### DIFF
--- a/src/Microsoft.FeatureManagement/FeatureFilterConfiguration.cs
+++ b/src/Microsoft.FeatureManagement/FeatureFilterConfiguration.cs
@@ -20,5 +20,13 @@ namespace Microsoft.FeatureManagement
         /// Configurable parameters that can change across instances of a feature filter.
         /// </summary>
         public IConfiguration Parameters { get; set; } = new ConfigurationRoot(new List<IConfigurationProvider>());
+
+        /// <summary>
+        /// A strongly-typed parameter object that can be used as an alternative to <see cref="Parameters"/>.
+        /// Custom <see cref="IFeatureDefinitionProvider"/> implementations can populate this property directly
+        /// instead of constructing an <see cref="IConfiguration"/> instance.
+        /// When set, feature filters should prefer this over <see cref="Parameters"/>.
+        /// </summary>
+        public object ParameterObject { get; set; }
     }
 }
--- a/src/Microsoft.FeatureManagement/FeatureFilterConfiguration.cs
+++ b/src/Microsoft.FeatureManagement/FeatureFilterConfiguration.cs
@@ -22,7 +22,7 @@ namespace Microsoft.FeatureManagement
         public IConfiguration Parameters { get; set; } = new ConfigurationRoot(new List<IConfigurationProvider>());
 
         /// <summary>
-        /// A strongly-typed parameter object that can be used as an alternative to <see cref="Parameters"/>.
+        /// A parameter object that can be used as an alternative to <see cref="Parameters"/>.
         /// Custom <see cref="IFeatureDefinitionProvider"/> implementations can populate this property directly
         /// instead of constructing an <see cref="IConfiguration"/> instance.
         /// When set, feature filters should prefer this over <see cref="Parameters"/>.
--- a/src/Microsoft.FeatureManagement/FeatureFilterConfiguration.cs
+++ b/src/Microsoft.FeatureManagement/FeatureFilterConfiguration.cs
@@ -27,6 +27,6 @@ namespace Microsoft.FeatureManagement
         /// instead of constructing an <see cref="IConfiguration"/> instance.
         /// When set, feature filters should prefer this over <see cref="Parameters"/>.
         /// </summary>
-        public object ParameterObject { get; set; }
+        public object ParametersObject { get; set; }
     }
 }
--- a/src/Microsoft.FeatureManagement/FeatureFilterEvaluationContext.cs
+++ b/src/Microsoft.FeatureManagement/FeatureFilterEvaluationContext.cs
@@ -22,10 +22,9 @@ namespace Microsoft.FeatureManagement
         public IConfiguration Parameters { get; set; }
 
         /// <summary>
-        /// A strongly-typed parameter object, if any, provided by a custom <see cref="IFeatureDefinitionProvider"/>.
-        /// When set, feature filters should prefer this over <see cref="Parameters"/>.
+        /// The settings provided for the feature filter to use when evaluating whether the feature should be enabled. This property takes precedence over Parameters if both are provided.
         /// </summary>
-        public object ParameterObject { get; set; }
+        public object ParametersObject { get; set; }
 
         /// <summary>
         /// A settings object, if any, that has been pre-bound from <see cref="Parameters"/>.
--- a/src/Microsoft.FeatureManagement/FeatureFilterEvaluationContext.cs
+++ b/src/Microsoft.FeatureManagement/FeatureFilterEvaluationContext.cs
@@ -22,6 +22,12 @@ namespace Microsoft.FeatureManagement
         public IConfiguration Parameters { get; set; }
 
         /// <summary>
+        /// A strongly-typed parameter object, if any, provided by a custom <see cref="IFeatureDefinitionProvider"/>.
+        /// When set, feature filters should prefer this over <see cref="Parameters"/>.
+        /// </summary>
+        public object ParameterObject { get; set; }
+
+        /// <summary>
         /// A settings object, if any, that has been pre-bound from <see cref="Parameters"/>.
         /// The settings are made available for <see cref="IFeatureFilter"/>s that implement <see cref="IFilterParametersBinder"/>.
         /// </summary>
--- a/src/Microsoft.FeatureManagement/FeatureFilterEvaluationContext.cs
+++ b/src/Microsoft.FeatureManagement/FeatureFilterEvaluationContext.cs
@@ -22,7 +22,7 @@ namespace Microsoft.FeatureManagement
         public IConfiguration Parameters { get; set; }
 
         /// <summary>
-        /// The settings provided for the feature filter to use when evaluating whether the feature should be enabled. This property takes precedence over Parameters if both are provided.
+        /// The settings provided for the feature filter to use when evaluating whether the feature should be enabled. This property takes precedence over <see cref="Settings"/> if both are provided.
         /// </summary>
         public object ParametersObject { get; set; }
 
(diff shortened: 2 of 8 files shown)


src/Microsoft.FeatureManagement/FeatureFilterConfiguration.cs
```diff
@@ -20,5 +20,13 @@ public class FeatureFilterConfiguration
...
         /// </summary>
         public IConfiguration Parameters { get; set; } = new ConfigurationRoot(new List<IConfigurationProvider>());
+
+        /// <summary>
+        /// A strongly-typed parameter object that can be used as an alternative to <see cref="Parameters"/>.
+        /// Custom <see cref="IFeatureDefinitionProvider"/> implementations can populate this property directly
+        /// instead of constructing an <see cref="IConfiguration"/> instance.
+        /// When set, feature filters should prefer this over <see cref="Parameters"/>.
+        /// </summary>
+        public object ParameterObject { get; set; }
```
jimmyca15:
```suggestion
        public object ParametersObject { get; set; }
```

src/Microsoft.FeatureManagement/FeatureFilterEvaluationContext.cs
```diff
@@ -21,6 +21,12 @@ public class FeatureFilterEvaluationContext
         /// </summary>
         public IConfiguration Parameters { get; set; }
 
+        /// <summary>
+        /// A strongly-typed parameter object, if any, provided by a custom <see cref="IFeatureDefinitionProvider"/>.
```
jimmyca15:
```suggestion
        /// The settings provided for the feature filter to use when evaluating whether the feature should be enabled. This property takes precedence over Parameters if both are provided.
```

src/Microsoft.FeatureManagement/Targeting/ContextualTargetingFilter.cs
```diff
@@ -68,8 +68,10 @@ public Task<bool> EvaluateAsync(FeatureFilterEvaluationContext context, ITargeti
             }
 
             //
-            // Check if prebound settings available, otherwise bind from parameters.
-            TargetingFilterSettings settings = (TargetingFilterSettings)context.Settings ?? (TargetingFilterSettings)BindParameters(context.Parameters);
+            // Check if ParameterObject available (takes precedence), then prebound settings, otherwise bind from parameters.
+            TargetingFilterSettings settings = context.ParameterObject != null
+                ? (TargetingFilterSettings)context.ParameterObject
```
jimmyca15: Consider the exception that will be raised by the system if someone provides the incorrect type here. Ideally we'd like a `FeatureManagementException` if someone populates ParametersObject, but it's not `TargetingFilterSettings`

linglingye001: Got it. Added `FeatureManagementError.InvalidParametersObject` for this scenario.

src/Microsoft.FeatureManagement/FeatureFilterEvaluationContext.cs
```diff
@@ -21,6 +21,11 @@ public class FeatureFilterEvaluationContext
         /// </summary>
         public IConfiguration Parameters { get; set; }
 
+        /// <summary>
+        /// The settings provided for the feature filter to use when evaluating whether the feature should be enabled. This property takes precedence over Parameters if both are provided.
```
linglingye001: over Settings, add see ref

src/Microsoft.FeatureManagement/FeatureManagementError.cs
```diff
@@ -31,6 +31,11 @@ public enum FeatureManagementError
         /// <summary>
         /// The given configuration setting was invalid.
         /// </summary>
-        InvalidConfigurationSetting
+        InvalidConfigurationSetting,
+
+        /// <summary>
+        /// The <see cref="FeatureFilterConfiguration.ParametersObject"/> provided for a feature filter is not of the expected type.
+        /// </summary>
+        InvalidParametersObject
```
linglingye001: we can throw argument exception

src/Microsoft.FeatureManagement/Targeting/ContextualTargetingFilter.cs
linglingye001: Discussed offline, throw `ArgumentException`  for incorrect type.

src/Microsoft.FeatureManagement/FeatureFilterEvaluationContext.cs
```diff
@@ -21,6 +21,11 @@ public class FeatureFilterEvaluationContext
         /// </summary>
         public IConfiguration Parameters { get; set; }
 
+        /// <summary>
+        /// The settings provided for the feature filter to use when evaluating whether the feature should be enabled. This property takes precedence over <see cref="Settings"/> if both are provided.
```
jimmyca15:
```suggestion
        /// The settings provided for the feature filter to use when evaluating whether the feature should be enabled. This property takes precedence over <see cref="Settings"/> and <see cref="Parameters"/> if both are provided.
```

jimmyca15:
Copilot's findings of "This assignment to parametersObject is useless, since its value is never read." are valid and should be addressed before merging.

jimmyca15 (approved):

jimmyca15 (approved):

src/Microsoft.FeatureManagement/FeatureManager.cs
```diff
@@ -498,6 +498,7 @@ private async ValueTask<bool> IsEnabledAsync<TContext>(FeatureDefinition feature
                     {
                         FeatureName = featureDefinition.Name,
                         Parameters = featureFilterConfiguration.Parameters,
+                        ParametersObject = featureFilterConfiguration.ParametersObject,
```
zhiyuanliang-ms, line 501: Can we set `ParametersObject` to `FeatureFilterEvaluationContext.Settings` instead of adding a new `ParametersObject` property.

@jimmyca15

src/Microsoft.FeatureManagement/FeatureFilters/TimeWindowFilter.cs
```diff
@@ -71,8 +71,17 @@ public Task<bool> EvaluateAsync(FeatureFilterEvaluationContext context)
             }
 
             //
-            // Check if prebound settings available, otherwise bind from parameters.
-            TimeWindowFilterSettings settings = (TimeWindowFilterSettings)context.Settings ?? (TimeWindowFilterSettings)BindParameters(context.Parameters);
```
zhiyuanliang-ms, line 75: If we set the new parametersObject to context.settings, the code path will be more unified.

src/Microsoft.FeatureManagement/FeatureManager.cs
jimmyca15: I think that's a good idea. We'll need an update on the description of `Settings`. Something like

```
/// <summary>
/// A settings object, if any, provided for the feature filter to use when evaluating whether the feature should be enabled.
/// This property is populated in two cases:
/// For features that provide parameters as an object.
/// For <see cref="IFeatureFilter"/>s that implement <see cref="IFilterParametersBinder"/>.
/// </summary>
```

jimmyca15: @linglingye001

PR #590, opened 2026-04-07
Update missing filter handling to treat it as disabled when RequirementType.All (#586)

### LINKED ISSUE #586: IgnoreMissingFeatureFilters Option Should Not Throw For RequirementType.All
**Problem:**
[FeatureManagementOptions.IgnoreMissingFeatureFilters when true will throw if a feature has RequirementType.All set](https://github.com/microsoft/FeatureManagement-Dotnet/blob/b9907a0213c034560957aedf756a64ac95f15c95/src/Microsoft.FeatureManagement/FeatureManager.cs#L425).

This implementation is extremely limiting as it prevents a global level option from being specified and used for requirement type Any features if a single feature of any number of potential features uses requirement type All. 

**Proposed Solution:**
- Remove the throw condition linked above, and allow the option to be specified even if some features use RequirementType.All
- Update [this IF condition instead to ](https://github.com/microsoft/FeatureManagement-Dotnet/blob/b9907a0213c034560957aedf756a64ac95f15c95/src/Microsoft.FeatureManagement/FeatureManager.cs#L487) to be `if (!_options.IgnoreMissingFeatureFilters || featureDefinition.RequirementType == RequirementType.All) { .. }`

linglingye001:
## Why this PR?

In [FeatureManager.cs L425](https://github.com/microsoft/FeatureManagement-Dotnet/blob/b9907a0213c034560957aedf756a64ac95f15c95/src/Microsoft.FeatureManagement/FeatureManager.cs#L425), within the `IsEnabledAsync` method, there is a hard throw when `IgnoreMissingFeatureFilters` is `true` and a feature uses `RequirementType.All`:
```
if (featureDefinition.RequirementType == RequirementType.All &&
    _options.IgnoreMissingFeatureFilters)
{
    throw new FeatureManagementException(
        FeatureManagementError.Conflict,
        "The 'IgnoreMissingFeatureFilters' flag cannot be used in combination with a feature of requirement type 'All'.");
}
```

- If any single feature uses `RequirementType.All`, the entire application crashes when that feature is evaluated, even though the option was intended for other features using `RequirementType.Any`.
- It prevents users from enabling `IgnoreMissingFeatureFilters` globally in scenarios where they have a mix of `RequirementType.Any` and `RequirementType.All` features.

Fix https://github.com/microsoft/FeatureManagement-Dotnet/issues/586

## Visible Changes
Remove the throw for `RequirementType.All && IgnoreMissingFeatureFilters`, and update the missing filter handling to treat it as disabled when `RequirementType.All`


jimmyca15 (approved):

jimmyca15:
PR title should be updated to reflect the specific scenario being addressed and issue linked.

linglingye001:
> PR title should be updated to reflect the specific scenario being addressed and issue linked.

Updated

Issue #367, opened 2024-02-11
Cahing for a custom FeatureDefinitionProvider


tahazayed:
I am building a custom FeatureDefinitionProvider and I want to support caching for it, but this interface IFeatureDefinitionProviderCacheable is internal; would you please change it to be public?

tahazayed:
I am using the same way as tests/Tests.FeatureManagement/InMemoryFeatureDefinitionProvider.cs

thompson-tomo:
You shouldn't need to add that interface to your provider given it is simply to alter handling for test cases and is not needed in actual use case. Please try it without and also what i would be doing is, implementing your provider as a configuration provider that way you definitions are immediately cached for the library to use.

jimmyca15:
@tahazayed Indeed this functionality should be available to you.

@rossgrambo [this conversation](https://github.com/microsoft/FeatureManagement-Dotnet/pull/229#discussion_r1189033113) mentions a plan to open up this functionality and to file an issue. Are we tracking it?

rossgrambo:
No, I don't believe it was ever tracked. I'll treat this as the tracking task and look into it.

tahazayed:
@jimmyca15 , @rossgrambo Thank you. In my circumstances, the alternative SaaS dynamic feature management costs $20,000 per month.

rossgrambo:
Hey @tahazayed, thanks for letting us know. I want to align on our different cacheing to ensure you're getting the caching you want.

IFeatureDefinitionProviderCacheable is an interface that enables a specific cache. This cache is for the binding of IConfiguration to a specific class. For our TimeWindowFilter- this cache prevents [IConfiguration.Get<TimeWindowFilterSettings>()](https://github.com/microsoft/FeatureManagement-Dotnet/blob/main/src/Microsoft.FeatureManagement/FeatureFilters/TimeWindowFilter.cs#L36) from firing too much, as this type conversion is expensive a few ms, and can be reduced to roughly ~1/100th of the time with this cache.

[...]

PR #591, opened 2026-04-15
Update cache check

linglingye001:
### Problem
The `IFeatureDefinitionProviderCacheable` interface in `IFeatureDefinitionProviderCacheable.cs` is internal, which prevents anyone implementing a custom `IFeatureDefinitionProvider` from opting into the filter parameter binding cache — a performance optimization that avoids expensive `IConfiguration.Get<T>()` calls on every evaluation.  https://github.com/microsoft/FeatureManagement-Dotnet/issues/367
 
### How the Caching Works Today
`if (!(_featureDefinitionProvider is IFeatureDefinitionProviderCacheable) || Cache == null)`

If the provider does not implement the internal marker interface, filter parameter binding (the `binder.BindParameters(context.Parameters`) call) runs on every evaluation — no caching. The `ConfigurationWrapper` class enables cache invalidation by creating a new reference when configuration changes, so `cacheItem.Parameters != context.Parameters` detects staleness.

## Visible Changes

Update the cache check in `FeatureManager` to also recognize `ConfigurationWrapper`-based parameters, so wrappers around `ConfigurationFeatureDefinitionProvider` automatically get caching without any API changes. For fully custom providers, [PR #589](https://github.com/microsoft/FeatureManagement-Dotnet/pull/589) is being worked on to let providers supply `FeatureDefinition` objects directly (bypassing the configuration layer entirely), which would make the caching question moot.


tests/Tests.FeatureManagement/FeatureManagementTest.cs
```diff
@@ -967,10 +967,10 @@ public async Task BindsFeatureFlagSettings()
             FeatureFilterConfiguration testFilterConfiguration = new FeatureFilterConfiguration
             {
                 Name = "Test",
-                Parameters = new ConfigurationBuilder().AddInMemoryCollection(new Dictionary<string, string>()
+                Parameters = new ConfigurationWrapper(new ConfigurationBuilder().AddInMemoryCollection(new Dictionary<string, string>()
```
zhiyuanliang-ms, line 970: Why do we need to update this? The `paramater` type is `IConfiguration`.

jimmyca15 (approved):

tests/Tests.FeatureManagement/FeatureManagementTest.cs
linglingye001: `ConfigurationWrapper` serves a specific purpose for cache invalidation. The [cache ](https://github.com/microsoft/FeatureManagement-Dotnet/blob/main/src/Microsoft.FeatureManagement/FeatureManager.cs#L697) compares parameters by reference equality.  When `ConfigurationFeatureDefinitionProvider` re-reads configuration, it creates a new `ConfigurationWrapper` instance.  The test does exactly this to verify cache update. Here use  `ConfigurationWrapper` is to mirror what `ConfigurationFeatureDefinitionProvider` does internally.

PR #607, opened 2026-05-27
support variant service match mode

### LINKED ISSUE #604: FeatureServiceProvider

zhiyuanliang-ms:
## Why this PR?

A follow up PR for #606, trying to resolve #604

Today `VariantServiceProvider<T>` only selects an implementation based on the assigned variant name. This PR adds an alternative mode where the implementation is selected based on the feature flag's enabled status (true/false), useful when a feature has no variants (e.g. a simple toggle flag).

Introduce `VariantServiceMatchMode` enum which has `Variant` (default) and `Status` value.

Usage:

```C#
// Service implementations
[VariantServiceAlias(true)]
public class EnabledCalculator : ICalculator { /* ... */ }

[VariantServiceAlias(false)]
public class DisabledCalculator : ICalculator { /* ... */ }

// Registration (keyed for lazy instantiation)
services.AddKeyedSingleton<ICalculator, EnabledCalculator>(true);
services.AddKeyedSingleton<ICalculator, DisabledCalculator>(false);

// Or normal registration
// services.AddSingleton<ICalculator, EnabledCalculator>();
// services.AddSingleton<ICalculator, DisabledCalculator>();

    .WithVariantService<ICalculator>("AdvancedMath", VariantServiceMatchMode.Status);
```
Note that the `AddKeyed*` family takes object? serviceKey, not just string.
image

### DIFF
--- a/src/Microsoft.FeatureManagement/VariantServiceProvider.cs
+++ b/src/Microsoft.FeatureManagement/VariantServiceProvider.cs
@@ -1,4 +1,4 @@
-﻿// Copyright (c) Microsoft Corporation.
+// Copyright (c) Microsoft Corporation.
 // Licensed under the MIT license.
 //
 using Microsoft.Extensions.DependencyInjection;
@@ -13,17 +13,18 @@ using System.Threading.Tasks;
 namespace Microsoft.FeatureManagement
 {
     /// <summary>
-    /// Used to get different implementations of TService depending on the assigned variant from a specific variant feature flag.
+    /// Used to get different implementations of TService depending on either the assigned variant or the enabled status of a feature flag.
     /// </summary>
     internal class VariantServiceProvider<TService> : IVariantServiceProvider<TService> where TService : class
     {
         private readonly IServiceProvider _serviceProvider;
         private readonly IVariantFeatureManager _featureManager;
         private readonly string _featureName;
+        private readonly VariantServiceMatch _matchMode;
         private readonly ConcurrentDictionary<string, TService> _variantServiceCache;
 
         /// <summary>
-        /// Creates a variant service provider.
+        /// Creates a variant service provider that selects an implementation based on the assigned variant of the feature flag.
         /// </summary>
         /// <param name="featureName">The feature flag that should be used to determine which variant of the service should be used.</param>
         /// <param name="featureManager">The feature manager to get the assigned variant of the feature flag.</param>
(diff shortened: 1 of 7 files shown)


src/Microsoft.FeatureManagement/VariantServiceProvider.cs
```diff
@@ -80,16 +106,54 @@ private TService ResolveVariantService(string variantName)
                 service => IsMatchingVariantName(service.GetType(), variantName));
         }
 
-        private bool IsMatchingVariantName(Type implementationType, string variantName)
+        private TService ResolveByStatus(bool enabled)
         {
-            string implementationName = ((VariantServiceAliasAttribute)Attribute.GetCustomAttribute(implementationType, typeof(VariantServiceAliasAttribute)))?.Alias;
+            if (_serviceProvider is IKeyedServiceProvider)
+            {
+                TService keyedService = _serviceProvider.GetKeyedService<TService>(enabled);
```
Stepami, line 113: what if i want to register my feature services against other keys?

zhiyuanliang-ms: I didn't test it. But my expectation is that you can register the same service multiple times using different keys,

Stepami: I'm trying to tell that if I want to use some key other than the feature status the variant sp won't get my impl as it doesn't know and can't know it.

The point is for VariantSp to be able to resolve my service not just with boolean key but any user-provided one

zhiyuanliang-ms: Hey, @Stepami This is a valid call. The solution we are considering is that:
Remove the VariantServiceMatchMode
Introduce a new `VariantServiceProviderOptions` (the name is not decided)
```C#
Service.AddFeatureManagement()
       .WithVariantService<TService>("MyFeature", new VariantServiceProviderOptions()
       {
           EnabledKey (NTD, we can have a better name) = "VariantServiceA", // work as fallback, pnly take effect where there is no variant allocated
           DisabledKey = "VariantServiceB"
       });
```

What do you think about this?

Stepami: Yeah, i think that's right direction into flexibility. But i also liked the idea of feature status value as default key if there's none from user. Maybe we could do some static factory methods, like:
```cs
VariantServiceProviderOptions.Variant()
VariantServiceProviderOptions.Status() // same as Status(enabledKey: true, disabledKey: false)
VariantServiceProviderOptions.Status(enabledKey: "VariantA", disabledKey: "VariantB")
```

zhiyuanliang-ms: The existing options in this library: [`FeatureManagementOptions`](https://github.com/microsoft/FeatureManagement-Dotnet/blob/main/src/Microsoft.FeatureManagement/FeatureManagementOptions.cs) and [`TargetingEvaluationOptions`](https://github.com/microsoft/FeatureManagement-Dotnet/blob/main/src/Microsoft.FeatureManagement/Targeting/TargetingEvaluationOptions.cs). We don't have the pattern like `Options.XXX()`

My preference would be to keep the options class property-based, this is more common in .NET

We can have default value for `EnabledKey` and `DisabledKey`

But, the feature flag status matching will be the fallback. If variant is not null and the name can match any implementation, that implementation will be returned. Variant has higher priority than flag status during the service matching.

Stepami: > But, the feature flag status matching will be the fallback. If variant is not null and the name can match any implementation, that implementation will be returned. Variant has higher priority than flag status during the service matching.

[Yeah it's totally makes sense considering that the old schema is the fallback](https://learn.microsoft.com/en-us/azure/azure-app-configuration/feature-management-dotnet-reference#net-feature-management-schema)

image

zhiyuanliang-ms:
Integrated to #606

PR #608, opened 2026-06-10
Feature/status resolution

### LINKED ISSUE #604: FeatureServiceProvider

Stepami:
## Why this PR?

Closes #604 

## Visible Changes

- `IFeatureServiceProvider<TService>` interface to get the implementation based on the feature status, maybe using additional context
- `WithFeatureService<TService, TEnabled, TDisabled>` extension method to setup feature service provider in di
- `FeatureServiceProviderOptions` class that specifies the keys used by a feature service provider to resolve an implementation based on the feature flag status when keyed di is available.

### DIFF
--- a/src/Microsoft.FeatureManagement/FeatureManagementBuilderExtensions.cs
+++ b/src/Microsoft.FeatureManagement/FeatureManagementBuilderExtensions.cs
@@ -75,5 +75,51 @@ namespace Microsoft.FeatureManagement
 
             return builder;
         }
+
+        /// <summary>
+        /// Adds a <see cref="FeatureServiceProvider{TService,TEnabled,TDisabled}"/> to the feature management system.
+        /// </summary>
+        /// <param name="builder">The <see cref="IFeatureManagementBuilder"/> used to customize feature management functionality.</param>
+        /// <param name="featureName">The feature flag that should be used to determine which implementation of the service should be used. The <see cref="IFeatureServiceProvider{TService}"/> will return different implementations of TService according to the feature status.</param>
+        /// <param name="options">Options used to configure the feature service provider.</param>
+        /// <returns>A <see cref="IFeatureManagementBuilder"/> that can be used to customize feature management functionality.</returns>
+        /// <exception cref="ArgumentNullException">Thrown if feature name parameter is null.</exception>
+        /// <exception cref="InvalidOperationException">Thrown if a feature service of the type has already been added.</exception>
+        public static IFeatureManagementBuilder WithFeatureService<TService, TEnabled, TDisabled>(this IFeatureManagementBuilder builder, string featureName, FeatureServiceProviderOptions options = null)
+            where TService : class
+            where TEnabled : class, TService
+            where TDisabled : class, TService
+        {
+            if (string.IsNullOrEmpty(featureName))
+            {
+                throw new ArgumentNullException(nameof(featureName));
+            }
+
+            options ??= new FeatureServiceProviderOptions();
+
+            if (builder.Services.Any(descriptor => descriptor.ServiceType == typeof(IFeatureServiceProvider<TService>)))
+            {
+                throw new InvalidOperationException($"A feature service of {typeof(TService).FullName} has already been added.");
+            }
+
+            if (builder.Services.Any(descriptor => descriptor.ServiceType == typeof(IFeatureManager) && descriptor.Lifetime == ServiceLifetime.Scoped))
+            {
+                builder.Services.AddScoped<IFeatureServiceProvider<TService>>(sp => new FeatureServiceProvider<TService, TEnabled, TDisabled>(
+                    sp,
+                    sp.GetRequiredService<IFeatureManager>(),
+                    featureName,
+                    options));
+            }
+            else
+            {
+                builder.Services.AddSingleton<IFeatureServiceProvider<TService>>(sp => new FeatureServiceProvider<TService, TEnabled, TDisabled>(
+                    sp,
+                    sp.GetRequiredService<IFeatureManager>(),
+                    featureName,
+                    options));
+            }
+
+            return builder;
+        }
     }
 }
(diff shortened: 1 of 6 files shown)


Stepami:
Hello @zhiyuanliang-ms ! It's been a little while. Could you take a look at my proposal of the FeatureServiceProvider API please?

zhiyuanliang-ms:
Hey, @Stepami Sorry for the late response. I actually discussed this with my colleague. We came up with the following design:

Introduce a new overload `WithVariantService<TService, TEnabled, TDisabled>("FeatureFlagName")`

The `TEnabled` and `TDisabled` are just used to specify the fallback implementation when there is no variant (in this case, the `VariantServiceProvider` will use feature flag status to retrieve service implementation from DI).

This is very similar to your proposal. But the benefit is that we don't need to introduce a new API with a new name and people can keep using `IVaraintServiceProvider`.

What do you think?

Stepami:
> Hey, @Stepami Sorry for the late response. I actually discussed this with my colleague. We came up with the following design:
> 
> Introduce a new overload `WithVariantService<TService, TEnabled, TDisabled>("FeatureFlagName")`
> 
> The `TEnabled` and `TDisabled` are just used to specify the fallback implementation when there is no variant (in this case, the `VariantServiceProvider` will use feature flag status to retrieve service implementation from DI).
> 
> This is very similar to your proposal. But the benefit is that we don't need to introduce a new API with a new name and people can keep using `IVaraintServiceProvider`.
> 
> What do you think?

@zhiyuanliang-ms

`IVariantServiceProvider` does not support context propagation for feature filters.

jimmyca15:
@stepami

Are you referring to the lack of a way to pass the filter context in as a parameter to resolve the service? That does seem like a deficiency of IVariantServiceProvider. We can consider how to address it.

zhiyuanliang-ms:
Hey, @Stepami

> `IVariantServiceProvider` does not support context propagation for feature filters.

This is a good call. We missed this for variant evaluation as well. [ValueTask<Variant> GetVariantAsync(string feature, ITargetingContext context, CancellationToken cancellationToken = default)](https://github.com/microsoft/FeatureManagement-Dotnet/blob/main/src/Microsoft.FeatureManagement/IVariantFeatureManager.cs#L55)

This makes thing tricky. It will be clean if we can add a new `GetService` method that takes a context parameter to `IVariantServiceProvider`. However, this will be a breaking change.

My current thought is that we can introduce a new `IContextualVariantServiceProvider`:

```C#
public interface IContextualVariantServiceProvider<TService> where TService : class
{
        ValueTask<TService> GetServiceAsync<TContext>(TContext context, CancellationToken cancellationToken);
}
```

We will have the current `VariantServiceProvider` to implement it as well. And the `WithVariantService` will then also register `IContextualVariantServiceProvider` to DI.

Stepami:
@jimmyca15 yes i am! @zhiyuanliang-ms did a great job of explaining my point.

However, i still do not understand why status based resolution should be bound to variant evaluation

zhiyuanliang-ms:
> However, i still do not understand why status based resolution should be bound to variant evaluation

Hey, @Stepami
Variant feature flag is the super set of simple feature flag. For the design simplicity, we want to unify them into variant. If something supports variant, it should also support toggle feature flag.

Stepami:
@zhiyuanliang-ms thanks, i got your point.

Why is addding a new `GetService` method that takes a context parameter to `IVariantServiceProvider` a breaking change?

zhiyuanliang-ms:
> @zhiyuanliang-ms thanks, i got your point.
> 
> Why is addding a new `GetService` method that takes a context parameter to `IVariantServiceProvider` a breaking change?

Hey, @Stepami

`IVariantServiceProvider` is an public interface. If someone is implementing it, then it will be broken. But I am wondering we can avoid this by adding default implementation for the new method. I need to do some investigation.

Stepami:
hey, @zhiyuanliang-ms
> `IVariantServiceProvider` is an public interface. If someone is implementing it, then it will be broken.

That's a good call. Well, there are not many options left. If your goal is to have as little new API as possible, then `IContextualVariantServiceProvider` is the way to go.

The next question i would ask is whether this interface extends `IVariantServiceProvider` or not.

> But I am wondering we can avoid this by adding default implementation for the new method. I need to do some investigation.

IMHO the only way it works is opposite - non-generic method calls `GetServiceAsync<object>(context: null, ct)`. It's already there in the `IFeatureManager` impl.

zhiyuanliang-ms:
Hey, @Stepami

> The next question i would ask is whether this interface extends `IVariantServiceProvider` or not.

Yes, `IContextualVariantServiceProvider` should extend `IVariantServiceProvider`

> But I am wondering we can avoid this by adding default implementation for the new method

I mean: adding a new method for an existing interface with a default implementation to avoid break change.

But since we have aligned with introducing `IContextualVariantServiceProvider`, this doesn't matter anymore. ;)
> non-generic method calls GetServiceAsync(context: null, ct). It's already there in the IFeatureManager impl

I agree.

zhiyuanliang-ms:
Hey, @Stepami

I sent out PR #611 to implement the proposed design.


cc @jimmyca15 @linglingye001

PR #612, opened 2026-08-16
Fix `IsEnabledAsync` null context

zhiyuanliang-ms:
## Why this PR?

https://github.com/microsoft/FeatureManagement-Dotnet/pull/611#discussion_r3792046607


src/Microsoft.FeatureManagement/FeatureManager.cs
```diff
@@ -178,6 +183,11 @@ public async ValueTask<bool> IsEnabledAsync(string feature, CancellationToken ca
         /// <returns>True if the feature is enabled, otherwise false.</returns>
         public async ValueTask<bool> IsEnabledAsync<TContext>(string feature, TContext appContext, CancellationToken cancellationToken = default)
         {
+            if (appContext is null)
```
zhiyuanliang-ms:
```suggestion
            if (appContext == null)
```

jimmyca15 (approved):

linglingye001 (approved):

Stepami (approved):

PR #615, opened 2026-08-24
feat(variants): allow in-memory objects

Stepami:
## Why this PR?

Implementing `IFeatureDefinitionProvider` makes it inconvinient to fill variant bound values if they don't come from configuration.

Similarly to feature filters I decided to introduce an object that can be used as an alternative `ConfigurationValue`. Custom `IFeatureDefinitionProvider` implementations can populate this property directly instead of constructing an `IConfigurationSection` instance.

## Visible Changes

- `Microsoft.FeatureManagement.VariantDefinition.ConfigurationObject`
- `Microsoft.FeatureManagement.Variant.ConfigurationObject`

### DIFF
--- a/src/Microsoft.FeatureManagement/FeatureManager.cs
+++ b/src/Microsoft.FeatureManagement/FeatureManager.cs
@@ -847,10 +847,11 @@ namespace Microsoft.FeatureManagement
                 variantConfiguration = variantDefinition.ConfigurationValue;
             }
 
-            return new Variant()
+            return new Variant
             {
                 Name = variantDefinition.Name,
-                Configuration = variantConfiguration
+                Configuration = variantConfiguration,
+                ConfigurationObject = variantDefinition.ConfigurationObject,
             };
         }
     }
--- a/src/Microsoft.FeatureManagement/Variant.cs
+++ b/src/Microsoft.FeatureManagement/Variant.cs
@@ -19,5 +19,11 @@ namespace Microsoft.FeatureManagement
         /// The configuration of the variant.
         /// </summary>
         public IConfigurationSection Configuration { get; set; }
+
+        /// <summary>
+        /// The configuration of the variant.
+        /// When set, variants should prefer this over <see cref="Configuration"/>.
+        /// </summary>
+        public object ConfigurationObject { get; set; }
     }
 }
--- a/src/Microsoft.FeatureManagement/VariantDefinition.cs
+++ b/src/Microsoft.FeatureManagement/VariantDefinition.cs
@@ -22,7 +22,7 @@ namespace Microsoft.FeatureManagement
         public IConfigurationSection ConfigurationValue { get; set; }
 
         /// <summary>
-        /// A parameter object that can be used as an alternative to <see cref="ConfigurationValue"/>.
+        /// A configuration object that can be used as an alternative to <see cref="ConfigurationValue"/>.
         /// Custom <see cref="IFeatureDefinitionProvider"/> implementations can populate this property directly
         /// instead of constructing an <see cref="IConfiguration"/> instance.
         /// When set, variants should prefer this over <see cref="ConfigurationValue"/>.
--- a/src/Microsoft.FeatureManagement/VariantDefinition.cs
+++ b/src/Microsoft.FeatureManagement/VariantDefinition.cs
@@ -22,6 +22,14 @@ namespace Microsoft.FeatureManagement
         public IConfigurationSection ConfigurationValue { get; set; }
 
         /// <summary>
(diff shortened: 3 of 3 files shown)


Stepami:
Hey, @zhiyuanliang-ms! could you take a look please?

zhiyuanliang-ms:
My understanding is that this enhancement is intended specifically for the custom feature definition provider scenario. Is that correct?

Stepami:
> My understanding is that this enhancement is intended specifically for the custom feature definition provider scenario. Is that correct?

yes it is. right now i am implementing a custom provider. i need to do something like this in order to put data from an external storage into the definition:
```cs
new VariantDefinition
{
    Name = "...",
    ConfigurationValue =
        new ConfigurationBuilder()
            .AddInMemoryCollection([new("Value", variantValue.ToString())])
            .Build()
            .GetSection("Value")
}
```

Stepami:
And the idea came to me from the `FeatureFilterConfiguration` class

zhiyuanliang-ms:
Hey, @Stepami 

We discussed this scenario internally, and we agree that supporting in-memory variant values is useful for custom `IFeatureDefinitionProvider` implementations.

One concern we have is that: for customers using the built-in `ConfigurationFeatureDefinitionProvider`, which we expect to be the common case, this property would not be populated and therefore would not provide any additional value.

Before adding a new public property, we would like to think through what the story should be for those customers as well. Ideally, this should feel like a generally useful part of the variant configuration model, rather than an API surface that only applies when a custom provider is used.

For example, one question we are considering is whether the built-in provider should also be able to populate an object representation from the configured variant value, or whether there is another API shape that gives both built-in and custom provider users a consistent way to consume variant configuration.

zhiyuanliang-ms:
I have been thinking about the following options:
1. Have `ConfigurationFeatureDefinitionProvider` populate `ConfigurationObject` as well, for example by constructing a [`JsonNode`](https://learn.microsoft.com/en-us/dotnet/api/system.text.json.nodes.jsonnode?view=net-11.0-pp) from the `IConfigurationSection`.

I used to think this is the best option. But after a second look, I realize the conversion will not be lossless. For example, a numeric value could end up being reconstructed as a string.

2. Today, `Variant.Configuration` is an `IConfigurationSection`, so customers using `ConfigurationFeatureDefinitionProvider` can naturally use the existing configuration binding APIs, such as `Bind` or `Get<T>()`, to convert the variant value into a typed object.

For example
```C#
Variant variant = await featureManager.GetVariantAsync("MyFeature");
MySettings settings = variant.Configuration.Get<MySettings>();
```

I am thinking about adding a new extension method to `Variant`. Something like:
```C#
Variant variant = await featureManager.GetVariantAsync("MyFeature");
MySettings settings = variant.GetConfiguration<MySettings>();
```

Or we can even add a new extension method to `IVariantFeatureManager`
like
```C#
MySettings settings = await featureManager.GetVariantConfigurationAsync<MySettings>("MyFeature");
```

This can unify the story that people who uses custom feature definition provider with in-memory object and built-in configuration feature defition provider.

Stepami:
hey @zhiyuanliang-ms !

I agree that `ConfigurationFeatureDefinitionProvider` could populate this property.

I'll take a few days to think this through

Stepami:
hey @zhiyuanliang-ms 

I decided to do both populate the data supplied by the buil-in configuration provider and create the common API used to consume variant configuration through extension method.

### ConfigurationObject
Now `ConfigurationObject` is filled with a read-only dictionary of relative configuration paths and string values inside the `ConfigurationFeatureDefinitionProvider`. I found this solution balanced as we don't lose the data as in JSON reconstruction and preserve original configuration structure. Relative paths preserve nested settings and array indices, while the empty key represents an inline scalar value. Case-insensitive lookup follows configuration conventions, and the read-only wrapper prevents consumers from modifying the shared dictionary.

### GetConfiguration
`Variant.GetConfiguration<T>()` provides a common consumption API:
- If `ConfigurationObject` is assignable to `T`, it returns that object directly. Custom providers can supply typed settings without serialization or binding, and reference identity is preserved.
- Otherwise, it delegates to `Configuration.Get<T>()`. Built-in-provider consumers can request typed settings without handling the dictionary themselves. An incompatible object does not prevent using the existing configuration representation.
- If the variant or fallback configuration is absent, it returns `default(T)`. Callers can request nullable value types, such as `int?`, when they need to distinguish absence from zero.

`T` is unconstrained because variant values can be both complex objects and inline scalars. Pattern matching supports reference types and boxed value types.

I placed the method on `Variant` as an extension so callers can consume an already evaluated variant and retain access to its name, without expanding `IVariantFeatureManager` or changing custom implementations.

Stepami:
hey @zhiyuanliang-ms! have you seen the update describe above?

PR #617, opened 2026-08-26
Open telemetry update

yuanqu72:
This pull request adds a new OpenTelemetry demo to the repository, demonstrating how to integrate OpenTelemetry-based telemetry and metrics with ASP.NET Core applications. It also introduces a new telemetry-related project, updates the solution structure, and enhances the documentation to include the new example.

**Major additions and changes:**

**New Example Application:**
- Added a new example project, `OpenTelemetryDemo`, which demonstrates using OpenTelemetry for logging custom events and tracking metrics in an ASP.NET Core web app. The demo includes pages for checkout and error handling, and showcases integration with Azure Monitor OpenTelemetry Exporter. [[1]](diffhunk://#diff-77e000abf1ac4bacf469d1fcf286941b53670aad018a56df096414c26aff195dR1-R17) [[2]](diffhunk://#diff-dd1ab3e32e996226e9ab6900a82d48c2252dcb88f1866d3273734cc4285abdd4R1-R20) [[3]](diffhunk://#diff-9ff6b708b71a5eab9ae3e341a2d612d332eebab357df408698a5a6767cf1f2ffR1-R46) [[4]](diffhunk://#diff-f236a3a79cfeacf38e81f1935d3818949032d5e493be22146a8afae4124673a6R1-R26) [[5]](diffhunk://#diff-f6c27a2149174db625ad9bb7e50f162ab559458ceb9e40b29ad6bc39d2ee6c44R1-R30) [[6]](diffhunk://#diff-121a74f896cdf03927aa05730c51eeef8679738cb26fc7a0db053b3c52da6b58R1-R37)

**Solution and Project Structure Updates:**
- Updated `Microsoft.FeatureManagement.sln` to include the new `OpenTelemetryDemo` example, the new `Microsoft.FeatureManagement.Telemetry.OpenTelemetry` project, and its associated test project. Also added a `src` solution folder and configured solution/project build settings for x64 and x86 platforms. [[1]](diffhunk://#diff-8c9e0713d6b6a2dd9be2fa7d750465a8e6b69a9de21d46d0b73dd6400e19bb15R34-R231) [[2]](diffhunk://#diff-8c9e0713d6b6a2dd9be2fa7d750465a8e6b69a9de21d46d0b73dd6400e19bb15R246-R248)

**Documentation:**
- Updated the `README.md` to reference the new OpenTelemetry demo, making it discoverable for users looking for telemetry integration examples.

**Testing Strategy:**
- Unit tests
- Validated that the newly added `OpenTelemetryDemo` is running and sending telemetry to my dedicated App Insights. 
- Compared against the previous related pr to make sure I did not miss anything: https://github.com/microsoft/FeatureManagement-Dotnet/pull/578


examples/OpenTelemetryDemo/Pages/Checkout.cshtml.cs
```diff
@@ -0,0 +1,46 @@
...
+        public int CheckoutAmount { get; set; }
+
+        public void OnPost()
+        {
+            CheckoutAmount = Random.Shared.Next(1, 100);
+
+            //
+            // Track the checkout amount metric using the OpenTelemetry metrics API
+            // (System.Diagnostics.Metrics), exported via WithMetrics/AddMeter.
+            var checkoutAmountHistogram = _meter.CreateHistogram<long>("checkoutAmount");
```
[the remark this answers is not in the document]
yuanqu72: Updated.

examples/OpenTelemetryDemo/Pages/Index.cshtml
```diff
@@ -0,0 +1,72 @@
...
+    ViewData["Title"] = "Home page";
+}
+
+<div class="text-center mb-40">
+    <h1 class="display-4">Welcome, @Model.Username !</h1>
+</div>
+
+<div class="text-center">
+    <div style="display:inline-block;width:500px;">
+        <img class="mb-40" src="@ViewData["ImageUri"]" />
```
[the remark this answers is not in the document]
yuanqu72: Updated locally.

examples/OpenTelemetryDemo/Pages/Index.cshtml.cs
```diff
@@ -0,0 +1,76 @@
...
+            {
+                string val = Request.Form["imageScore"];
+
+                if (val != null &&
+                    int.TryParse(val, out int rating))
+                {
+                    //
+                    // Track the image rating metric using the OpenTelemetry metrics API
+                    // (System.Diagnostics.Metrics), exported via WithMetrics/AddMeter.
+                    var imageRatingHistogram = _meter.CreateHistogram<long>("ImageRating");
```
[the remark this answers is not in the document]
yuanqu72: Updated.

examples/OpenTelemetryDemo/Pages/RandomizeUser.cshtml.cs
```diff
@@ -0,0 +1,29 @@
...
+                new Claim(ClaimTypes.Name, Random.Shared.Next().ToString())
+            };
+
+            var identity = new ClaimsIdentity(claims, "CookieAuth");
+            var principal = new ClaimsPrincipal(identity);
+
+            HttpContext.SignInAsync("CookieAuth", principal);
+
+            return RedirectToPage("/Index");
+        }
```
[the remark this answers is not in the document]
yuanqu72: Updated locally.

src/Microsoft.FeatureManagement.Telemetry.OpenTelemetry/TargetingActivityProcessor.cs
```diff
@@ -0,0 +1,39 @@
...
+            // Extract the targeting id from the activity's baggage
+            string targetingId = activity.Baggage.FirstOrDefault(t => t.Key == TargetingIdKey).Value;
+
+            // Don't modify the activity if there's no available targeting id
+            if (string.IsNullOrEmpty(targetingId))
+            {
+                return;
+            }
+
+            activity.SetTag(TargetingIdKey, targetingId);
```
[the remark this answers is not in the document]
yuanqu72: Updated locally.

src/Microsoft.FeatureManagement.Telemetry.OpenTelemetry/OpenTelemetryEventPublisher.cs
```diff
@@ -0,0 +1,119 @@
...
+        /// </summary>
+        /// <param name="logger">The logger used to emit the OpenTelemetry log-based custom event.</param>
+        public OpenTelemetryEventPublisher(ILogger<OpenTelemetryEventPublisher> logger)
+        {
+            _logger = logger ?? throw new ArgumentNullException(nameof(logger));
+
+            _activityListener = new ActivityListener
+            {
+                ShouldListenTo = (activitySource) => activitySource.Name == FeatureManagementActivitySourceName,
+                Sample = (ref ActivityCreationOptions<ActivityContext> options) => ActivitySamplingResult.AllData,
```
linglingye001: There's filtering in [azure monitor log processor](https://github.com/Azure/azure-sdk-for-net/blob/Azure.Monitor.OpenTelemetry.Exporter_1.9.0/sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/LogFilteringProcessor.cs#L19). If there's no surrounding activity like console app, will the evaluation log be dropped because of the filtering?

yuanqu72: I think the evaluation log will get dropped if no other listener marks the evaluation activity as recorded. I will update this line to `ActivitySamplingResult.AllDataAndRecorded`. Good catch!

examples/OpenTelemetryDemo/LoggerExtensions.cs
```diff
@@ -0,0 +1,37 @@
...
+    /// this app's custom events. Each event's message template includes the
+    /// "microsoft.custom_event.name" placeholder, which becomes a `LogRecord` attribute that the
+    /// Azure Monitor OpenTelemetry Exporter (if configured) uses to classify the log as a custom
+    /// event in Application Insights, rather than a plain trace.
+    /// </summary>
+    public static class LoggerExtensions
+    {
+        private static readonly Action<ILogger, string, int, Exception> _vote = LoggerMessage.Define<string, int>(
+            LogLevel.Information,
+            new EventId(1, "microsoft.custom_event.name"),
```
zhiyuanliang-ms: Is `microsoft.custom_event.name` required by Azure Monitor? If there any doc about this?

zhiyuanliang-ms:
Should this PR target on preview branch or main branch?

src/Microsoft.FeatureManagement.Telemetry.OpenTelemetry/FeatureEvaluationEventPublisher.cs
```diff
@@ -0,0 +1,119 @@
...
+using System.Diagnostics;
+
+namespace Microsoft.FeatureManagement.Telemetry.OpenTelemetry
+{
+    /// <summary>
+    /// Listens to <see cref="Activity"/> events from feature management and emits them as OpenTelemetry log-based custom events.
+    /// </summary>
+    internal sealed class OpenTelemetryEventPublisher : IDisposable
+    {
+        private const string AzureMonitorCustomEventNameKey = "microsoft.custom_event.name";
```
zhiyuanliang-ms, line 14: It looks like "microsoft.custom_event.name" is the contract Azure Monitor OpenTelemetry uses to identify a log record as a [custom event](https://learn.microsoft.com/en-us/azure/azure-monitor/app/opentelemetry-add-modify?tabs=aspnetcore#send-custom-events)).

I’m surprised Azure Monitor doesn’t expose this value as a public constant for consumers. This leaves our library relying on a string-based contract, but I don’t see a better alternative to take a hard dependency from Azure Monitor OpenTelemetry package.

src/Microsoft.FeatureManagement.Telemetry.OpenTelemetry/FeatureManagementBuilderExtensions.cs
```diff
@@ -0,0 +1,87 @@
...
+        /// <remarks>
+        /// <see cref="TargetingActivityProcessor"/> and <see cref="TargetingLogProcessor"/> are automatically registered with the app's <c>TracerProvider</c> and <c>LoggerProvider</c> respectively,
+        /// so exported traces/spans and log-based telemetry (including custom events) are enriched with targeting information whenever a <c>TargetingId</c> is present in the current
+        /// <see cref="System.Diagnostics.Activity"/>'s baggage. This method moves its processor registrations to the front of the service collection, so they always run before exporters
+        /// configured elsewhere (for example via <c>ILoggingBuilder.AddOpenTelemetry(options =&gt; ...)</c> or <c>TracerProviderBuilder.AddConsoleExporter()</c>), regardless of whether this
+        /// method is called before or after those exporters are configured.
+        /// </remarks>
+        /// <param name="builder">The feature management builder.</param>
+        /// <returns>The feature management builder.</returns>
+        public static IFeatureManagementBuilder AddOpenTelemetry(this IFeatureManagementBuilder builder)
```
zhiyuanliang-ms: How about renaming it to `WithOpenTelemetry`? We've already established the `WithXXX` pattern: `WithVariantService`, `WithTargeting`

src/Microsoft.FeatureManagement.Telemetry.OpenTelemetry/FeatureManagementBuilderExtensions.cs
```diff
@@ -0,0 +1,87 @@
...
+            builder.Services.AddSingleton<TargetingActivityProcessor>();
+
+            builder.Services.AddSingleton<TargetingLogProcessor>();
+
+            // Automatically wire TargetingActivityProcessor ahead of any span exporters so every
+            // exported Activity/span is enriched with TargetingId.
+            builder.Services.ConfigureOpenTelemetryTracerProvider((serviceProvider, tracerProviderBuilder) =>
+                tracerProviderBuilder.AddProcessor(serviceProvider.GetRequiredService<TargetingActivityProcessor>()));
+
+            MoveLastDescriptorToFront(builder.Services, "OpenTelemetry.Trace.IConfigureTracerProviderBuilder");
```
zhiyuanliang-ms: I think this is kind of anti-pattern which fight against the OTEL DI registration pattern.
I don't like the idea of reordering another library’s DI registrations. It is a hack.

I think we should have OTEL integration extension method for `IOpenTelemetryBuilder`

Something like:

```C#
public static IOpenTelemetryBuilder WithFeatureManagement(this IOpenTelemetryBuilder builder)
{
    return builder
        .WithTracing(tracing => tracing
            .AddSource("Microsoft.FeatureManagement")
            .AddProcessor<TargetingActivityProcessor>())
        .WithLogging(logging => logging
            .AddProcessor<TargetingLogProcessor>());
}
```
And we can simplify the FeatureManagementBuilder extension method

We are going to have  two apis:
`IOpenTelemetryBuilder.WithFeatureManagement()` registers the activity source and processors using public OTel APIs.
IFeatureManagementBuilder.WithAzureMonitorTelemetry() only registers the FeatureEvaluation publisher and its lifecycle.

The ideal usage will be like:

```C#
    .WithTargeting()
    .WithAzureMonitor(); // <- new api

services.AddOpenTelemetry()
    .WithFeatureManagement() // new api
    .UseAzureMonitor();
```

src/Microsoft.FeatureManagement.Telemetry.OpenTelemetry/FeatureEvaluationEventPublisher.cs
zhiyuanliang-ms: This class is really responsible for publishing `FeatureEvaluation` as an Azure Monitor custom event using the Azure Monitor-specific "microsoft.custom_event.name" contract.

I am thinking about the more appropriate class name is `AzureMonitorEventPublisher` or `FeatureEvaluationEventPublisher`. And the namespace should mention Azure Monitor.

zhiyuanliang-ms: Since this is an internal class, we can use a long name `AzureMonitorFeatureEvaluationEventPublisher`

src/Microsoft.FeatureManagement.Telemetry.OpenTelemetry/FeatureManagementBuilderExtensions.cs
yuanqu72: Final decision:
* Moved the targeting and log processor registration under `IOpenTelemetryBuilder` (`OpenTelemetryBuilderExtension.cs`). This API design makes the telemetry configuration and its ordering more visible. 
* Removed DI descriptor reordering. 
* Now, the order guarantee (processors need to be before exporters to ensure targetingId is attached across the span) is done through documented registration order. See updated `Program.cs` in the sample and its README.

src/Microsoft.FeatureManagement.Telemetry.OpenTelemetry/FeatureEvaluationEventPublisher.cs
yuanqu72: Updated the file name to "FeatureEvaluationEventPublisher.cs". 

The class publishes through `ILogger` ; it does not directly export to Azure Monitor. It does add the Azure Monitor-specific `microsoft.custom_event.name`  attribute, but other OpenTelemetry backends can still receive the log and its attributes.

yuanqu72:
> Should this PR target on preview branch or main branch?

I think main branch make sense because 1) It does not break existing customers using AppInsightsTelemetry since it is an addition, 2) Python FeatureManagement already supports OpenTelemetry.

examples/OpenTelemetryDemo/LoggerExtensions.cs
zhiyuanliang-ms: How about having a const for `"microsoft.custom_event.name"`

src/Microsoft.FeatureManagement.Telemetry.OpenTelemetry/OpenTelemetryHostedService.cs
```diff
@@ -0,0 +1,47 @@
...
+//
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Hosting;
+
+namespace Microsoft.FeatureManagement.Telemetry.OpenTelemetry
+{
+    /// <summary>
+    /// A hosted service used to construct and dispose the <see cref="FeatureEvaluationEventPublisher"/>
+    /// </summary>
+    internal sealed class OpenTelemetryHostedService : IHostedService
```
zhiyuanliang-ms:
```suggestion
    internal sealed class FeatureEvaluationEventPublisherHostedService : IHostedService
```

src/Microsoft.FeatureManagement.Telemetry.OpenTelemetry/OpenTelemetryBuilderExtensions.cs
```diff
@@ -0,0 +1,71 @@
...
+
+            if (!builder.Services.Any((ServiceDescriptor d) => d.ServiceType == typeof(IHostedService) && d.ImplementationType == typeof(OpenTelemetryHostedService)))
+            {
+                builder.Services.Insert(0, ServiceDescriptor.Singleton<IHostedService, OpenTelemetryHostedService>());
+            }
+
+            return builder;
+        }
+
+        private sealed class FeatureManagementTelemetryRegistration
```
zhiyuanliang-ms: I think we may not need this.
If we really want to prevent duplicated `WithFeatureManagement` call, we can check DI container and throw.

src/Microsoft.FeatureManagement.Telemetry.OpenTelemetry/OpenTelemetryBuilderExtensions.cs
```diff
@@ -0,0 +1,71 @@
...
+        /// <summary>
+        /// Adds feature evaluation event publishing and targeting enrichment to OpenTelemetry.
+        /// </summary>
+        /// <remarks>
+        /// Automatically registers <see cref="TargetingActivityProcessor"/> and <see cref="TargetingLogProcessor"/>
+        /// to enrich spans and logs with <c>TargetingId</c> from activity baggage.
+        /// </remarks>
+        /// <param name="builder">The OpenTelemetry builder.</param>
+        /// <returns>The supplied OpenTelemetry builder.</returns>
+        public static IOpenTelemetryBuilder WithFeatureManagement(this IOpenTelemetryBuilder builder)
```
zhiyuanliang-ms: Can we have a mechanism to check whether `UseAzureMonitor` is called before calling `WithFeatureManagement`? If possible, we can throw exception in this case.

examples/OpenTelemetryDemo/Program.cs
```diff
@@ -0,0 +1,82 @@
...
+
+OpenTelemetryBuilder openTelemetryBuilder = builder.Services.AddOpenTelemetry();
+
+// Register targeting processors before exporters so TargetingId is added before telemetry is exported.
+// Keep WithFeatureManagement() before the UseAzureMonitor() exporter configuration below.
+openTelemetryBuilder.WithFeatureManagement();
+
+if (!string.IsNullOrEmpty(connectionString))
+{
+    openTelemetryBuilder.UseAzureMonitor(o => o.ConnectionString = connectionString);
```
zhiyuanliang-ms, line 39: What will customer's code look like if they use the following way to use azure monitor:

Image

https://learn.microsoft.com/en-us/azure/azure-monitor/app/opentelemetry-add-modify?tabs=aspnetcore#send-custom-events

src/Microsoft.FeatureManagement.Telemetry.OpenTelemetry/TargetingActivityProcessor.cs
```diff
@@ -0,0 +1,42 @@
...
+using System.Diagnostics;
+
+namespace Microsoft.FeatureManagement.Telemetry.OpenTelemetry
+{
+    /// <summary>
+    /// Adds targeting information to outgoing OpenTelemetry <see cref="Activity"/> spans.
+    /// </summary>
+    public class TargetingActivityProcessor : BaseProcessor<Activity>
+    {
+        private const string TargetingIdKey = "TargetingId";
```
zhiyuanliang-ms: The `TargetingIdKey` is used in both processors. Can we create a shared const for it?

examples/OpenTelemetryDemo/LoggerExtensions.cs
yuanqu72: Updated.

src/Microsoft.FeatureManagement.Telemetry.OpenTelemetry/OpenTelemetryHostedService.cs
yuanqu72: Updated locally.

src/Microsoft.FeatureManagement.Telemetry.OpenTelemetry/OpenTelemetryBuilderExtensions.cs
yuanqu72: I agree that the class is not needed. But throwing error is not necessary since duplicate calls aren’t inherently invalid. Make it no-op is still better in my opinion.  

New implementation: Removed the marker class and now use the `FeatureEvaluationEventPublisherHostedService` registration as the early guard. Repeated `WithFeatureManagement()` calls remain no-ops, preventing duplicate processor configuration without throwing.

src/Microsoft.FeatureManagement.Telemetry.OpenTelemetry/OpenTelemetryBuilderExtensions.cs
yuanqu72: Azure Monitor has an [internal registration marker](https://github.com/Azure/azure-sdk-for-net/blob/Azure.Monitor.OpenTelemetry.Exporter_1.8.3/sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/UseAzureMonitorExporterRegistration.cs) we could inspect without adding a package dependency. However, its name isn’t a public contract and could change between versions. To keep this library exporter-independent and avoid relying on Azure Monitor internals, I suggest documenting the required call order rather than throwing based on this marker.

examples/OpenTelemetryDemo/Program.cs
yuanqu72: This setup works for sending custom events to Azure Monitor, but it doesn’t enable our full feature-management integration. Our current `WithFeatureManagement()` API uses the application host to manage the feature-evaluation publisher’s lifecycle and configures targeting enrichment for logs and traces. Customers should follow our demo setup for automatic feature-evaluation publishing and targeting enrichment when targeting information is available.

With a standalone `LoggerFactory`, they can manually add `TargetingLogProcessor` before the exporter, as shown below. This adds `TargetingId` to logs when it is available in the current activity’s baggage, but does not enable automatic feature-evaluation event publishing.

```
using var loggerFactory = LoggerFactory.Create(builder =>
{
    builder.AddOpenTelemetry(options =>
    {
        options.AddProcessor(new TargetingLogProcessor());
        options.AddAzureMonitorLogExporter();
    });
});
```

src/Microsoft.FeatureManagement.Telemetry.OpenTelemetry/TargetingActivityProcessor.cs
yuanqu72: Updated.

src/Microsoft.FeatureManagement.Telemetry.OpenTelemetry/OpenTelemetryBuilderExtensions.cs
zhiyuanliang-ms: It seems like OTel doesn't expose a public way to inspect registered exporters (I think we can open a feature request issue for them), So, I agree the best effort we can do here is to document the required call order.

examples/OpenTelemetryDemo/Program.cs
zhiyuanliang-ms: Adding `TargetingLogProcessor` only addresses the targeting id erichment, but it doesn't enable `FeatureEvaluation` event. So, the LoggerFactory scenario is actually broken. I am thinking whether it is a good idea to offer `IDisposable SubscribeToFeatureEvaluationEvents(
    this ILoggerFactory loggerFactory)`

``` C#
using ILoggerFactory loggerFactory = LoggerFactory.Create(builder =>
{
    builder.AddOpenTelemetry(options =>
    {
        options.AddAzureMonitorLogExporter();
    });
});

using IDisposable subscription = loggerFactory.SubscribeToFeatureEvaluationEvents();

```

zhiyuanliang-ms: I actually expect people to write code like this:

```c#
services.AddLogging(builder =>
{
    builder.AddOpenTelemetry(options =>
        options.AddAzureMonitorLogExporter());
});

services.AddOpenTelemetry()
    .WithFeatureManagement();
```

AFAIK, standalone `LoggerFactory` is not a common use case.

zhiyuanliang-ms: The feature managment package is designed primarily for dependency injection usage. We probably shouldn’t complicate the API solely to support this edge case in the doc example.

zhiyuanliang-ms (approved):

build/install-dotnet.ps1
zhiyuanliang-ms, line 1: Is this change related to the PR?

yuanqu72: This change improves diagnostics for .NET installation failures, but it isn’t required for OpenTelemetry support. I encountered some silent failures during testing, so updated the code to surface the errors more clearly.

zhiyuanliang-ms (approved):

src/Microsoft.FeatureManagement.Telemetry.OpenTelemetry/FeatureEvaluationEventPublisherHostedService.cs
```diff
@@ -0,0 +1,47 @@
...
+        }
+
+        /// <summary>
+        /// Uses the service provider to construct a <see cref="FeatureEvaluationEventPublisher"/> which will start listening for activities.
+        /// </summary>
+        /// <param name="cancellationToken">The cancellation token.</param>
+        /// <returns>A <see cref="Task"/> representing the asynchronous operation.</returns>
+        public Task StartAsync(CancellationToken cancellationToken)
+        {
+            _serviceProvider.GetRequiredService<FeatureEvaluationEventPublisher>();
```
WeihanLi: Maybe this should be disposed when stop

jimmyca15: If it's in the service container, the service container should dispose of it when the container is disposed.

PR #623, opened 2026-09-10
Make precedence source ordered

### LINKED ISSUE #621: Microsoft-schema flag in appsettings.json silently shadows the same flag from Azure App Configuration (.NET schema)
### Summary

When `appsettings.json` declares a flag under the Microsoft schema (`feature_management:feature_flags`) and Azure App Configuration supplies a flag with the same name, the file wins unconditionally and the App Configuration definition — including its feature filters — is silently ignored.

This is not the array-merging problem from #507. It happens because the two sources land in **different schemas**, and `ConfigurationFeatureDefinitionProvider` resolves the Microsoft schema first regardless of configuration source order:

```csharp
_getFeatureDefinitionFunc = (featureName) =>
{
    return Task.FromResult(GetMicrosoftSchemaFeatureDefinition(featureName) ?? GetDotnetSchemaFeatureDefinition(featureName));
};
```

The Azure App Configuration provider emits a flag into the Microsoft schema (`feature_management:feature_flags`) only when the flag uses variants or telemetry; otherwise it emits into the classic .NET schema (`FeatureManagement:<name>`) — see Azure/AppConfiguration-DotnetProvider#640. So for a plain flag (no variants, no telemetry) App Configuration contributes **nothing** to `feature_management:feature_flags`, and a same-named entry in `appsettings.json` shadows it permanently.

### `CustomConfigurationMergingEnabled` does not cover this

`FindFeatureFlags` collects only `feature_management:feature_flags` sections, per configuration provider. The Azure App Configuration provider contributes zero sections there for plain flags, so the merged list is file-only and the `LastOrDefault` lookup has nothing from App Configuration to choose.

### Repro

`appsettings.json`:

```json
{
  "feature_management": {
    "feature_flags": [
      { "id": "MyFlag", "enabled": false }
    ]
  }
}
```

App Configuration: feature flag `MyFlag`, enabled, with a `Microsoft.Targeting` filter, no variants and no telemetry.

```csharp
var config = new ConfigurationBuilder()
    .AddJsonFile("appsettings.json")
    .AddAzureAppConfiguration(options =>
    {
        options.Connect(connectionString);
        options.UseFeatureFlags();
    })
    .Build();

var services = new ServiceCollection();
services.AddSingleton<IConfiguration>(config);
services.Configure<ConfigurationFeatureDefinitionProviderOptions>(o => o.CustomConfigurationMergingEnabled = true);
services.AddFeatureManagement().WithTargeting();
```

**Expected:** the App Configuration definition wins, being the later configuration source, and the targeting filter is evaluated.

**Actual:** `IsEnabledAsync("MyFlag")` always returns `false` and the targeting filter never runs.

Dumping what each configuration provider contributes to `feature_management:feature_flags`:

```
-- JsonConfigurationProvider:      [0] id=MyFlag enabled=False
-- AzureAppConfigurationProvider:  (no feature_flags)
```

Versions: `Microsoft.FeatureManagement.AspNetCore` 4.6.0, `Microsoft.Azure.AppConfiguration.AspNetCore` 8.6.0, .NET 10.

### Why this is easy to walk into

The failure is silent and asymmetric. Flags that exist *only* in App Configuration keep working normally, so a deployment looks healthy while every flag that also has a file default is quietly frozen at that default — including flags whose App Configuration definition carries a targeting audience, which then applies to nobody (or, in the inverse case, a targeted rollout becomes a global one because the file default is `true`).

Moving file defaults from `FeatureManagement` to `feature_management`, which the current docs encourage, is by itself enough to trigger this, and `CustomConfigurationMergingEnabled` reads like the setting that addresses precisely this scenario.

### Possible resolutions

Any of, in rough order of preference:

- Make precedence source-ordered rather than schema-ordered, so a .NET-schema definition from a later configuration source takes priority over a Microsoft-schema definition from an earlier one.
- Have `CustomConfigurationMergingEnabled` also collect `FeatureManagement:<name>` sections per provider, so both schemas participate in the same per-source merge.
- Failing either, document the interaction, state in the `CustomConfigurationMergingEnabled` docs that it does not apply to Azure App Configuration flags without variants or telemetry, and log a warning when one flag name is defined in both schemas.

linglingye001:
## Why this PR?

Fix https://github.com/microsoft/FeatureManagement-Dotnet/issues/621

`ConfigurationFeatureDefinitionProvider` always evaluates the Microsoft schema first, source order is never consulted when the same flag is split across schemas. A Microsoft-schema definition from the first provider beats a .NET-schema definition from every later provider.

## Visible Changes

  - Tracks which schema last defines each feature by configuration-source order.
  - Later sources win across .NET and Microsoft schemas when custom merging is enabled.
  - Microsoft schema wins when both schemas occur in the same source.
  - Deduplicates cross-schema entries during feature enumeration.
  - Leaves default, non-custom behavior unchanged.

### Example
When custom merging is enabled:

  Provider` 1: Microsoft schema → MyFlag = false
  Provider 2: .NET schema       → MyFlag = targeting filter
  
  Winning schema map:
  MyFlag → .NET
  
  Result: Provider 2 wins and the targeting filter is parsed.


src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
```diff
@@ -178,12 +187,54 @@ private void EnsureInit()
...
-                _dotnetFeatureDefinitionSections = GetDotnetFeatureDefinitionSections();
+                LoadFeatureDefinitionSections();
 
+                _initialized = 1;
+            }
+        }
+
+        private void LoadFeatureDefinitionSections()
+        {
+            _dotnetFeatureDefinitionSections = GetDotnetFeatureDefinitionSections();
```
zhiyuanliang-ms, line 198: `_dotnetFeatureDefinitionSections` comes from the aggregated `_configuration`, while the winning schema is determined source by source.

I would expect that when custom merging is enabled, .NET feature flag sections should also be collected source by source. The current implementation is kind of inconsistant

src/Microsoft.FeatureManagement/ConfigurationFeatureDefinitionProvider.cs
```diff
@@ -281,18 +318,41 @@ private void FindFeatureFlags(IConfiguration configuration, List<IConfigurationS
...
+
+        private void AddFeatureDefinitions(
+            IConfiguration configuration,
+            List<IConfigurationSection> microsoftFeatureDefinitionSections,
+            IDictionary<string, FeatureDefinitionSchema> featureDefinitionSchemas)
+        {
+            IConfigurationSection dotnetFeatureManagementSection = configuration
+                .GetSection(DotnetFeatureManagementFields.FeatureManagementSectionName);
+
+            foreach (IConfigurationSection featureSection in dotnetFeatureManagementSection.GetChildren())
```
zhiyuanliang-ms, line 338: Previously, `ConfigurationFeatureDefinitionProvider` kept two responsibilities separate:

1. Loading feature definition configuration sections.
2. Parsing those sections into feature definitions.

`_dotnetFeatureDefinitionSections` and `_microsoftFeatureDefinitionSections` acted as the ground truth, and the contents were interpreted only when feature definitions were requested.

In the current implementation, the loading phase also partially parse the feature definitions while traversing the configuration sources. It reads feature flag IDs to maintain `_featureDefinitionSchemas`.

zhiyuanliang-ms: I think we can use the following data structure to maintain feature definition sections grouped by configuration source:

```C#
private IEnumerable<FeatureDefinitionSectionsBySource> _featureDefinitionSources;

private class FeatureDefinitionSectionsBySource
{
    public IEnumerable<IConfigurationSection> DotnetSections { get; }

    public IEnumerable<IConfigurationSection> MicrosoftSections { get; }
}
```

When getting a feature definition, we can traverse the configuration sources from highest to lowest precedence. Within each source, we check the Microsoft schema first, then .NET schema.

linglingye001: Grouping sections by source would retain the information needed for precedence without a separate schema map. I think it is a cleaner design.

zhiyuanliang-ms (approved):

tests/Tests.FeatureManagement/FeatureManagementTest.cs
```diff
@@ -255,6 +261,438 @@ public async Task RespectsAllFeatureManagementSchemas()
             Assert.True(await featureManager.IsEnabledAsync("FeatureZ"));
         }
 
+        [Fact]
```
zhiyuanliang-ms, line 264: Could you help open another PR after this one is merged to split up ·FeatureManagementTest·
Otherwise, this file is just going to keep growing indefinitely in the future.

linglingye001: Will do.

